PR #310, opened 2015-01-26
Feature/2.2.1

fblundun:


alexanderdean:
Looks great! I will get this released this afternoon...

fblundun:
I've just noticed that the PerformanceTiming fix for Safari fails on Firefox - I will try to find something that works for both...

alexanderdean:
Thanks @fblundun - please ping in this thread when fixed...

fblundun:
OK @alexanderdean , it's fixed now!

PR #335, opened 2015-03-06
Feature/2.4.0

fblundun:


src/js/tracker.js
```diff
@@ -901,7 +962,7 @@
 			 * @param string url
 			 */
 			setReferrerUrl: function (url) {
```
alexanderdean: @fblundun can you add some doco on setReferrerUrl in the user manual? There doesn't seem to be any.

fblundun: I've added it in.

PR #380, opened 2015-06-29
Log error when pixel fails to load

fivetran:
If there is an error while loading the pixel from the server, the current behavior is that snowplow fails silently. This change logs an error.

### DIFF
--- a/src/js/out_queue.js
+++ b/src/js/out_queue.js
@@ -221,7 +221,9 @@
 					executeQueue();
 				};
 
-				image.onerror = function () {
+				image.onerror = function (error) {
+					console.log('Error loading snowplow pixel', error);
+
 					executingQueue = false;
 				};
 


fblundun:
Thanks @fivetran ! Could you sign our CLA? https://github.com/snowplow/snowplow/wiki/CLA

alexanderdean:
Scheduled for debug mode as this doesn't make sense for production

fivetran:
I signed the CLA

alexanderdean:
Confirmed!

alexanderdean:
Scheduled to 3.0.0

paulboocock:
This PR won't quite fit with plans for #381 as we won't be using console.log for the new debug mode. We will emit errors/warnings when the pixel fails to load those. Closing this in favour of the wider debug mode coming in v3.

PR #440, opened 2015-12-10
Respect doNotTrack in IE 11 and Safari 7.1.3+

grzegorzewald:
As per https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack IE 11 as well as Safari 7.1.3+ store doNotTrack setting in window rather than navigator object.

### DIFF
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -155,7 +155,7 @@
 			configCookiePath = '/',
 
 			// Do Not Track browser feature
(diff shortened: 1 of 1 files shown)


alexanderdean:
Thanks! Can you sign our CLA please: https://github.com/snowplow/snowplow/wiki/CLA

grzegorzewald:
@alexanderdean: Signed!

alexanderdean:
Confirmed! Thanks so much

PR #457, opened 2016-02-16
Release/2.6.0

jbeemster:


src/js/in_queue.js
```diff
@@ -112,8 +112,13 @@
...
 		function createNewNamespace(namespace, endpoint, argmap) {
 			argmap = argmap || {};
-			trackerDictionary[namespace] = new TrackerConstructor(functionName, namespace, version, pageViewId, mutSnowplowState, argmap);
-			trackerDictionary[namespace].setCollectorUrl(endpoint);
+
+			if (!trackerDictionary.hasOwnProperty(namespace)) {
+				trackerDictionary[namespace] = new TrackerConstructor(functionName, namespace, version, pageViewId, mutSnowplowState, argmap);
+				trackerDictionary[namespace].setCollectorUrl(endpoint);
+			} else {
+				helpers.warn('Tracker namespace ' + namespace + ' already exists.')
```
fblundun: Missing semicolon!

src/js/lib/helpers.js
```diff
@@ -290,4 +290,77 @@
...
+	object.findRootDomain = function () {
+		var cookiePrefix = '_sp_root_domain_test_';
+		var cookieName = cookiePrefix + new Date().getTime();
+
+		var split = window.location.hostname.split('.');
+		var position = split.length - 1;
+		while (position >= 0) {
+			var currentDomain = split.slice(position, split.length).join('.');
+			document.cookie = cookieName + '=;domain=' + currentDomain + ';';
+			if (document.cookie.indexOf(cookieName) > -1) {
```
fblundun: We already have the browser-cookie-lite module as a dependency. You can see it in node_modules/browser-cookie-lite/browser-cookie.js. We might as well use it for setting and getting cookies here.

src/js/tracker.js
```diff
@@ -796,6 +845,220 @@
...
+						context['device']['isIncognito'] = fingerprint.browserHasIncognitoEnabled;
+					}
+				}
+
+				return {
+					schema: 'iglu:io.augur.snowplow/identity_lite/jsonschema/1-0-0',
+					data: context
+				};
+			}
+		}
```
fblundun: I think that I have been overusing `hasOwnProperty` - it's time for us to stop using it so much in the JS Tracker. It is only really helpful when you are looping through the properties of an object ([reference](http://blog.javascripting.com/2014/09/19/when-should-javascript-developers-use-hasownproperty/)), and it makes code less readable.

For example, 

```
var consumer = augur.hasOwnProperty('consumer') ? augur.consumer : {};
```

can be rewritten to 

```js
var consumer = augur.consumer || {};
```

and

```js
if (fingerprint.hasOwnProperty('browserHasIncognitoEnabled')) {
  context['device']['isIncognito'] = fingerprint.browserHasIncognitoEnabled;
}
```

to

```js
context['device']['isIncognito'] = fingerprint.browserHasIncognitoEnabled;
```

This is probably my fault for bigging up my old hasOwnProperty-filled code to you!

jbeemster: But isn't there a chance that if the data structure changes that the key might not be available and in that happening an exception occurring when trying to get that value?

I agree it is very verbose at the moment.

fblundun: In JS trying to get a nonexistent property of an object doesn't throw an exception, it just returns `undefined`. It can make debugging hard but in this case it is useful.

Note that this doesn't work for nested properties:

```js
({}).device; // undefined
({}).device.isIncognito; // Uncaught TypeError: Cannot read property 'isIncognito' of undefined
```

To do it with nested properties, you need [brototype](https://github.com/letsgetrandy/brototype).

jbeemster: My next thought then is do we want to populate the value for the key with 'null' even if it is not available?  Wouldn't it be easier to know when we need to update the logic if we start seeing fields not being populated rather than just having them all be 'null'?

fblundun: I think that our JSON stringification library actually changes `{"a": undefined}` to `"{}"`, meaning that if `fingerprint.browserHasIncognitoEnabled === null` we will see `null` in the JSON but if the property is undefined it won't exist in the JSON at all...

jbeemster: I see!  Will double check that but should allow a lot less code bloat then.

src/js/lib/helpers.js
```diff
@@ -290,4 +291,78 @@
...
+	object.findRootDomain = function () {
+		var cookiePrefix = '_sp_root_domain_test_';
+		var cookieName = cookiePrefix + new Date().getTime();
+		var cookieValue = '_test_value_' + new Date().getTime();
+
+		var split = window.location.hostname.split('.');
+		var position = split.length - 1;
+		while (position >= 0) {
+			var currentDomain = split.slice(position, split.length).join('.');
+			cookie.cookie(cookieName, cookieValue, 0, '/', currentDomain);
```
fblundun: I think it's slightly confusing to use 0 as the ttl here - it relies on the fact that browser-cookie-lite interprets a ttl of 0 as meaning that no ttl is set rather than that the ttl should be 0 seconds. Could we change it to null?

jbeemster: According to the documentation: https://github.com/litejs/browser-cookie-lite

"If set to 0, or omitted, the cookie will expire at the end of the session (when the browser closes)."

src/js/tracker.js
```diff
@@ -796,6 +845,215 @@
...
+		 * Creates a context from the window['optimizely'].data.experiments object
+		 *
+		 * @return array Experiment contexts
+		 */
+		function getOptimizelyExperimentContexts() {
+			var experiments = window['optimizely'].data.experiments;
+			if (experiments) {
+				var contexts = [];
+
+				for (var key in experiments) {
```
fblundun: For loops like this are a good example of when we actually should use hasOwnProperty. If somebody else has executed something like [Object.prototype.watch](https://gist.github.com/eligrey/384583) on the page, the new "watch" property will be included in this loop. Presumably `Object.prototype.watch.code` evaluates to `undefined`, so we would end up with an extra context with no populated fields.

src/js/tracker.js
```diff
@@ -738,12 +745,54 @@
...
+					}
+				}
+
+				if (autoContexts.optimizelyAudiences) {
+					combinedContexts = combinedContexts.concat(getOptimizelyAudienceContexts());
+				}
+
+				if (autoContexts.optimizelyDimensions) {
+					combinedContexts = combinedContexts.concat(getOptimizelyDimensionContexts());
+				}
```
fblundun: Can we change these `concat`s to `push`es? (The reason I use concat at the top of this function is to create a new object rather than modifying an existing one, which can cause problems when there are multiple trackers on the page sharing state.)

CHANGELOG
```diff
@@ -1,3 +1,25 @@
+Version 2.6.0 (2016-02-xx)
```
alexanderdean: Finalize date to today please @jbeemster

Issue #401, opened 2015-08-07
Support dependency-based wait for tracking


alexanderdean:
In the scenario that Augur (e.g.) can retrieve a user fingerprint in 400ms, you may prefer to wait that 400ms and attach an Augur fingerprint to all events, rather than have some late-firing events have Augur data and some early-firing events have no Augur data. The exact same applies to `setUserId`.

This ticket would make it possible to delay all event tracking based on a precondition such as the Augur data being available. You can think of this as effectively a blocking wait-to-send, except of course in a JavaScript async world this wouldn't be literally implemented with a block.

We should probably consider a "dead man's handle", where after a configuration timeout, the block is removed (because e.g. there is an instrumentation bug and that user ID is **never** going to be passed into the dataLayer for this page load).

fblundun:
One possible API:

```js
// After the Snowplow tag has loaded
snowplow_fn('notify', 'condition2');
snowplow_fn('addDependency', 'condition1');
snowplow_fn('addDependency', 'condition2');
snowplow_fn('trackPageView'); // and other tracker methods
snowplow_fn('notify', 'condition1');
// Now trackPageView and so on actually happen
```

```js
// After the Snowplow tag has loaded
snowplow_fn('addDependency', 'condition3', 1000); // with a timeout
snowplow_fn('trackPageView'); // and other tracker methods
// trackPageView happens after 1000 ms have passed
```

alexanderdean:
Interesting! Can you clarify what `notify` and `addDependency` mean? Also, would 'conditionX' actually be a function that needs to evaluate to true for the dependency to pass?

fblundun:
The idea is that 'conditionX' is just a string label for a particular condition. `addDependency` means "don't start until this condition is met" and `notify` means "this condition has now been met so if we were previously waiting on it we can now proceed".

For instance you might say

```js
snowplow_fn('addDependency', 'augurLoaded');
function augurLoadedCallback() {
  snowplow_fn('notify', 'augurLoaded');
}
```

to wait for Augur to load.

alexanderdean:
Got it! I think that's really nice.

alexanderdean:
I'd probably call it "notifyDependency" so there is symmetry between the add and the notify

alexanderdean:
More `passDependency` than `notifyDependency`?

fblundun:
[...]

PR #518, opened 2016-09-26
Parrable context

chantalgo:
Patch for the Parrable context according to the schema change on #404

### DIFF
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -1104,7 +1104,7 @@
 			if (parrable) {
 				var context = { encryptedId: null, optout: null};
 				context['encryptedId'] =  parrable.browserid;
-				context['optout'] =  parrable.browserid;	
+				context['optout'] =  parrable.optout;	
 				return {
 					schema: 'iglu:com.parrable/encrypted_payload/jsonschema/1-0-0',
 					data: context
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -1102,7 +1102,7 @@
 		function getParrableContext() {
 			var parrable = window['_hawk'];
 			if (parrable) {
-				var context = { encryptedID: null, optout: null};
+				var context = { encryptedId: null, optout: null};
 				context['encryptedId'] =  parrable.browserid;
 				context['optout'] =  parrable.browserid;	
 				return {
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -1103,8 +1103,8 @@
 			var parrable = window['_hawk'];
 			if (parrable) {
 				var context = { encryptedID: null, optout: null};
-				context['encryptedID'] =  parrable.browserid;
-				context['optout'] =  parrable.browserid		
+				context['encryptedId'] =  parrable.browserid;
+				context['optout'] =  parrable.browserid;	
 				return {
 					schema: 'iglu:com.parrable/encrypted_payload/jsonschema/1-0-0',
 					data: context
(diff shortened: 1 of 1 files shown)


chantalgo:
@alexanderdean Could you help me figure out the issue here? Does it have to do with the fact that the schema change has not yet been patched?

src/js/tracker.js
```diff
@@ -1087,6 +1095,24 @@
...
 		/**
+		 * Creates a context from the window['_hawk'] object
+		 *
+		 * @return object The Parrable context
+		 */
+		function getParrableContext() {
+			var parrable = window['_hawk'];
+			if (parrable) {
+				var context = { encryptedID: null, optout: null};
+				context['encryptedID'] =  parrable.browserid;
```
alexanderdean: I think it's `encryptedId`, not `ID`

alexanderdean:
Yes @chantalgo the schema change hasn't been released yet...

chantalgo:
We usually serve the output value inside our async callback `_hawk("parrable", function(parrable) { … }` but that requires being called after we can assert `hawk.js` has been loaded. 

I added a commit that should be a fix for the output value.

It is also important to note that since we are using the variable`browserid` (readily available upon load) instead of the value obtained from the callback, the first impression value is lost and it can be null. (Using the callback would guarantee a value.)

alexanderdean:
> It is also important to note that since we are using the variablebrowserid (readily available upon load) instead of the value obtained from the callback, the first impression value is lost and it can be null. (Using the callback would guarantee a value.)

Yes, that's similar to our Augur implementation. We have a ticket to add a dependency system for inter-async-JS dependencies (#401) but that hasn't been actioned yet.

chantalgo:
@alexanderdean Do I need to make any changes for the updates?

alexanderdean:
Hi @chantalgo - no, you should be good to go!

snowplowcla:

:memo: **Please visit https://github.com/snowplow/snowplow/wiki/CLA to learn more and sign.**

Once you've signed, please reply here (e.g. `I signed it!`) and we'll verify. Thanks.

src/js/tracker.js
```diff
@@ -1087,6 +1095,25 @@
...
+		 *
+		 * @return object The Parrable context
+		 */
+		function getParrableContext() {
+			var parrable = window['_hawk'];
+			if (parrable) {
+				var context = { encryptedId: null, optout: null};
+				context['encryptedId'] =  parrable.browserid;
+				var regex = new RegExp('(?:^|;)\\s?' + "_parrable_hawk_optout".replace(/([.*+?^=!:${}()|[\]\/\\])/g, '\\$1') + '=(.*?)(?:;|$)', 'i'), match = document.cookie.match(regex);
+				parrable['optout'] = (match && decodeURIComponent(match[1])) ? match && decodeURIComponent(match[1]) : "false";
```
alexanderdean: Do you mean `context['optout'] = ...` here?

chantalgo: Yes, apologies

chantalgo:
Any issues on why it fails, please let me know so I can fix, thank you  @alexanderdean  for your time

chantalgo:
CLA has been signed for the Parrable tem

snowplowcla:
Confirmed! @chantalgo has signed the Software Grant and Corporate Contributor License Agreement. Thanks so much.

alexanderdean:
LGTM! Scheduling into 2.7.0 assuming the rest of the Parrable integration can be completed by then.

chuwy:
Closing in favor #525. Thanks.

Issue #551, opened 2017-02-13
Add support for OptimizelyX context


digdeep:
There is now **OptimizelyX** with a new JSON blob. 
```js
optimizely.get('data')
```
We will write out the necessary JS and update this thread.  

Related to: #466

alexanderdean:
Awesome - thanks @digdeep!

Issue #525, opened 2016-11-01
Add Parrable context


chuwy:
Companion ticket for #518

chuwy:
Pushing it back to 2.8.0 which will be released shortly after 2.7.0.

alexanderdean:
Change capitalization in commit message

Issue #521, opened 2016-09-30
Exclude password fields from form tracking


alexanderdean:
Migrated from https://github.com/snowplow/snowplow/issues/2902

Issue #506, opened 2016-08-09
Add support for recording state in localStorage rather than cookies


alexanderdean:
This is to help minimize overall HTTP header size.

Most servers only allow up to 4kb (Akamai default); some will bump this to say 16kb but this can still not be enough for some user's accumulated cookies on one request. Having such big HTTP headers increase the total size of all upstream requests and binds a lot of memory on server requests and therefore reduces the number of possible concurrent connections.

We have future proofed our client_session schema to record where the session is stored: https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-1#L33

So hopefully it is not a huge change to support this? The only note is that we should add a new Tracker Protocol parameter (maybe we won't flow it through Snowplow right now) to track what storage we are using for this session, so that:
1. we can fully back-populate `client_session`s contexts from TP parameters when we finally switch to using `client_session` for web
2. if a user switches the storage, we have a theoretical way of untangling changing session identifiers (because they won't survive the switch) (theoretical because if we don't represent the new param in atomic.events, it will be very hard to review)

alexanderdean:
What namespacing should we use in localStorage for this?

falschparker82:
This affects us as well, as we're using the Javascript SDK inside Ionic / Cordova. We have a heavy problem with duplicated first sessions across users on mobile, as cookies apparently don't stick:

```
Affected users:
90% of all ios users
0,3% of all web users
43,7% of all android users
```

chuwy:
Pushing back.

BenFradet:
Trying to wrap my head around this, I'm guessing the session cookie refers to this: https://github.com/snowplow/snowplow-javascript-tracker/blob/master/src/js/tracker.js#L531-L536 but what about the device?

alexanderdean:
Hi @BenFradet - I've renamed the ticket to make it slightly less confusing.

There are two cookies currently:

1. https://github.com/snowplow/snowplow-javascript-tracker/blob/master/src/js/tracker.js#L531-L536
2. https://github.com/snowplow/snowplow-javascript-tracker/blob/master/src/js/tracker.js#L531-L536

The second is for client sessionization which is sent by the tracker back to Snowplow; the first is used to help keep track of that session.

BenFradet:
Also, do we want to stop supporting storing those as cookies altogether of provide both alternatives?

alexanderdean:
[...]

PR #562, opened 2017-03-29
2.8.0

BenFradet:
- [x] readme (#561)
- [x] parable (PR: #518, issue: #525)
- [x] client session in local storage (#506)
- [x] external activity tracking (#559)
- [x] no tracking (#563)
- [x] deprecate `useCookies` (#565)
- [x] deprecate `useLocalStorage` (#566)
- [x] opt-out of state tracking (#459)
- [x] page view id bug (#500, fixed in 2.7.0)
- [x] wiki updates
- [x] optimizely X (reviewed, PR: #567, issue #551, depends on snowplow/iglu-central#495)
- [x] one setInterval (#571)
- [x] dnt password fields (PR: #570, issue #521)
- [x] guards against non-integers `heartBeatDelay` and `minimumVisitTime` (#572)
- [x] RO access to `domainSessionIndex` (#573)
- [x] RO access to `cookieName` (#574)
- [x] RO access to `pageViewId` (#575)
- [x] prep release

### DIFF
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -186,7 +186,11 @@
 			useLocalStorage = argmap.hasOwnProperty('useLocalStorage') ? argmap.useLocalStorage : true,
 
 			// Whether to use cookies
-			configUseCookies = argmap.hasOwnProperty('useCookies') ? argmap.useCookies : true,
+			configUseCookies = argmap.hasOwnProperty('useCookies') ? (
+				helpers.warn(
+					'argmap.useCookies is deprecated. Use argmap.stateStorageStrategy instead.'),
+				argmap.useCookies
+			) : true,
 
 			// Strategy defining how to store the state: cookie, localStorage or none
 			configStateStorageStrategy = argmap.hasOwnProperty('stateStorageStrategy') ?
@@ -592,7 +596,8 @@
 		 * Set the cookies (if cookies are enabled)
 		 */
 		function initializeIdsAndCookies() {
-			var sesCookieSet = configUseCookies && !!getSnowplowCookieValue('ses');
+			var sesCookieSet =
+				configStateStorageStrategy != 'none' && !!getSnowplowCookieValue('ses');
 			var idCookieComponents = loadDomainUserIdCookie();
 
 			if (idCookieComponents[1]) {
@@ -614,7 +619,7 @@
 				idCookieComponents[5] = idCookieComponents[4];
 			}
 
-			if (configUseCookies) {
+			if (configStateStorageStrategy != 'none') {
 				setSessionCookie();
 				// Update currentVisitTs
 				idCookieComponents[4] = Math.round(new Date().getTime() / 1000);
@@ -627,7 +632,7 @@
 		 * Load visitor ID cookie
 		 */
 		function loadDomainUserIdCookie() {
-			if (!configUseCookies) {
+			if (configStateStorageStrategy == 'none') {
 				return [];
 			}
 			var now = new Date(),
@@ -684,7 +689,7 @@
 				lastVisitTs = id[5],
 				sessionIdFromCookie = id[6];
 
-			if (configDoNotTrack && configUseCookies) {
+			if (configDoNotTrack && configStateStorageStrategy != 'none') {
 				if (configStateStorageStrategy == 'localStorage') {
 					helpers.attemptWriteLocalStorage(idname, '');
 					helpers.attemptWriteLocalStorage(sesName, '');
@@ -701,7 +706,7 @@
 				memorizedSessionId = sessionIdFromCookie;
 
 				// New session?
-				if (!ses && configUseCookies) {
+				if (!ses && configStateStorageStrategy != 'none') {
 					// New session (aka new visit)
 					visitCount++;
 					// Update the last visit timestamp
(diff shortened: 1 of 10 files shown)


BenFradet:
Don't really know why the build is failing.

src/js/tracker.js
```diff
@@ -1433,6 +1484,24 @@
...
+			/**
+			 * Get the cookie name as cookieNamePrefix + basename + . + domain.
+			 *
+			 * @return string Cookie name
+			 */
+			getCookieName: function(basename) {
+				return getSnowplowCookieName(basename);
+			},
+
+			/**
```
BenFradet: I don't know if should just make those public instead of having a public wrapper.

alexanderdean: @chuwy thoughts?

chuwy: @BenFradet what do mean by public wrapper? Can you elaborate on options?

BenFradet: Either I make the existing `getSnowplowCookieName` and `getPageViewId` public or write a wrapper (as in this PR).

src/js/tracker.js
```diff
@@ -184,13 +185,22 @@
...
-			configUseCookies = argmap.hasOwnProperty('useCookies') ? argmap.useCookies : true,
+			configUseCookies = argmap.hasOwnProperty('useCookies') ? (
+				helpers.warn(
+					'argmap.useCookies is deprecated. Use argmap.stateStorageStrategy instead.'),
+				argmap.useCookies
+			) : true,
+
+			// Strategy defining how to store the state: cookies, localStorage or none
+			configStateStorageStrategy = argmap.hasOwnProperty('stateStorageStrategy') ?
+				argmap.stateStorageStrategy : (!configUseCookies ? 'none': 'cookie'),
```
BenFradet: I chose to not store anything if the user took the time to set `useCookies` to false, what do you guys think?

src/js/tracker.js
```diff
@@ -532,20 +548,36 @@
...
+		 * Sets a cookie based on the storage strategy:
+		 * - if 'localStorage': attempts to write to localStorage or failover to cookies
+		 * - if 'cookie': writes to cookies
+		 * - if something else no-op
+		 */
+		function setCookie(name, value, timeout) {
+			if (configStateStorageStrategy == 'localStorage') {
+				if (!helpers.attemptWriteLocalStorage(name, value)) {
+					cookie.cookie(name, value, timeout, configCookiePath, configCookieDomain);
+				}
```
BenFradet: I don't know if the failover to cookies is warranted, input welcome.

src/js/tracker.js
alexanderdean: Let's stick with the wrapper approach. We can do a more concerted refactor for the v3.0.0

BenFradet: yup @chuwy and I agreed on this also.

BenFradet:
@chuwy @alexanderdean 

One question remains though: what should we do with [`useLocalStorage`](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/src/js/tracker.js#L74), since it's used only for storing outgoing events right now?

I'm afraid this might become confusing to have both `useLocalStorage` and `stateStorageStrategy`, what do you guys think?

alexanderdean:
Hey @BenFradet - as discussed today, I think that useLocalStorage is a peer of useCookies, as per:

http://snowplowanalytics.com/blog/2014/12/15/snowplow-javascript-tracker-2.2.0-released/

I think we should deprecated useLocalStorage as we are deprecating useCookies; I think that the new variable with the word `state` in its name covers both session tracking and event buffering well; I think that we should break out both deprecations into their own tickets so that they are more visible in the CHANGELOG et al.

src/js/tracker.js
```diff
@@ -1410,7 +1413,8 @@
 
 				// Periodic check for activity.
 				lastActivityTime = now.getTime();
-				setInterval(function heartBeat() {
+				clearInterval(pagePingInterval)
+				pagePingInterval = setInterval(function heartBeat() {
```
BenFradet: ping @chuwy @alexanderdean

chuwy: LGTM (just add a semicolon after `clearInterval`). Do you have evidences it was invoked multiple times?

BenFradet: It's one of only of two explanations I could come up with regarding #568, but no, no evidence.
Anyway, I don't think that could hurt the codebase since not keeping track of intervals is a pretty bad idea altogether.

chuwy: Agree!

alexanderdean: LGTM. I've added a suggestion about a new ticket to cover the other scenario...

src/js/tracker.js
```diff
@@ -1833,8 +1836,15 @@
...
-				configHeartBeatTimer = heartBeatDelay * 1000;
+				if (minimumVisitLength === parseInt(minimumVisitLength, 10) &&
+						heartBeatDelay === parseInt(heartBeatDelay, 10)) {
+					activityTrackingEnabled = true;
+					configMinimumVisitTime = new Date().getTime() + minimumVisitLength * 1000;
+					configHeartBeatTimer = heartBeatDelay * 1000;
+				} else {
+					helpers.warn("Activity tracking not enabled, please provide integer values " +
+						"for minimumVisitLength and heartBeatDelay.")
+				}
```
BenFradet: @alexanderdean is this what you had in mind?

alexanderdean: Yes! LGT @chuwy ?

chuwy: Yes, this looks sensible. 

@alexanderdean what do you think about using `trackError` in such warnings? Otherwise user may never know he has wrong tracker setup (not counting absent page pings in logs).

alexanderdean: Interesting idea @chuwy - please create a ticket for 3.0.0 to explore this idea...

BenFradet:
@alexanderdean good to go when green

BenFradet:
@alexanderdean ready to merge

PR #552, opened 2017-02-14
Add support for OptimizelyX context #551

miike:
RFC for adding support for an Optimizely X summary context. Currently the Optimizely summary context does not work due to changes in the Javascript API between 'Classic' and 'X'.

Note: This context doesn't currently existing in Iglu Central though it's a pretty simple schema. I think it makes sense to have this as a separate schema (rather than adding to `getOptimizelySummary`) as Optimizely consider it a separate product and some fields are either not accessible or applicable between Optimizely Classic and Optimizely X.

Would love to get your thoughts on this @chuwy @alexanderdean

Issue #553, opened 2017-02-17
Add defensive check for window.optimizely.data


alexanderdean:
* Install and activate a tracking blocker (eg Ghostery)
* Block Optimizely
* Snowplow JavaScript fails to initialize

```javascript
Uncaught TypeError: Cannot read property 'experiments' of undefined
```

Theory: Possibly this is because the initial check for `window.optimizely passes`, but `window.optimizely.data` is blocked and consequently `window.optimizely.data.state` is querying a null object.

PR #567, opened 2017-04-05
Add support for OptimizelyX summary context

### LINKED ISSUE #551: Add support for OptimizelyX context

miike:
Refactor of #552, fixes #551

### DIFF
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -930,7 +930,7 @@
 		 */
 		function getOptimizelyXData(property, snd) {
 			var data;
-			if (windowAlias.optimizely && windowAlias.optimizely.data) {
+			if (windowAlias.optimizely) {
 				data = windowAlias.optimizely.get(property);
 				if (typeof snd !== 'undefined' && data !== undefined) {
 					data = data[snd]
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -1200,7 +1200,7 @@
 		function getOptimizelyXSummaryContexts() {
 			return lodash.map(getOptimizelyXSummary(), function (experiment) {
 				return {
-					schema: 'iglu:com.optimizely.snowplow/optimizelyx_summary/jsonschema/1-0-0',
+					schema: 'iglu:com.optimizely.optimizelyx/summary/jsonschema/1-0-0',
 					data: experiment
 				};
 			});
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -769,6 +769,13 @@
 					})
 				}
 
+				if (autoContexts.optimizelyXSummary) {
+					var activeExperiments = getOptimizelyXSummaryContexts();
+					lodash.each(activeExperiments, function (e) {
+						combinedContexts.push(e);
+					})
+				}
+
 				if (autoContexts.optimizelyExperiments) {
 					var experimentContexts = getOptimizelyExperimentContexts();
 					for (var i = 0; i < experimentContexts.length; i++) {
@@ -916,6 +923,23 @@
 		}
 
 		/**
+		 * Check that *both* optimizely and optimizely.data exist
+		 *
+		 * @param property optimizely data property
+		 * @param snd optional nested property
+		 */
+		function getOptimizelyXData(property, snd) {
+			var data;
+			if (windowAlias.optimizely && windowAlias.optimizely.data) {
+				data = windowAlias.optimizely.get(property);
+				if (typeof snd !== 'undefined' && data !== undefined) {
+					data = data[snd]
+				}
+			}
+			return data
+		}
+
+		/**
 		 * Get data for Optimizely "lite" contexts - active experiments on current page
 		 * 
 		 * @returns Array content of lite optimizely lite context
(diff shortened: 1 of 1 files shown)


miike:
@BenFradet Any advice on the best way to test this?

src/js/tracker.js
```diff
@@ -916,6 +923,23 @@
...
+		function getOptimizelyXData(property, snd) {
+			var data;
+			if (windowAlias.optimizely && windowAlias.optimizely.data) {
+				data = windowAlias.optimizely.get(property);
+				if (typeof snd !== 'undefined' && data !== undefined) {
+					data = data[snd]
+				}
+			}
+			return data
+		}
```
BenFradet: This seems to be the exact same thing as [`getOptimizelyData`](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/src/js/tracker.js#L900-L916), why not just reuse it?

src/js/tracker.js
```diff
@@ -938,6 +962,31 @@
...
+
+			return lodash.map(experiment_ids, function(activeExperiment) {
+				variation = state.getVariationMap()[activeExperiment];
+				variationName = variation.name;
+				variationId = variation.id;
+				visitorId = visitor.visitorId;
+				return {
+					experimentId: parseInt(activeExperiment),
+					variationName: variationName,
+					variation: parseInt(variationId),
```
BenFradet: What do you think about having this named `variationId`?

BenFradet:
@miike regarding testing, all you'd need is an optimizely account and a js-tracker built from your branch.

However, this would need snowplow/iglu-central#495 to be merged in.

src/js/tracker.js
miike: `getOptimizelyData` uses `windowAlias.optimizely.data` to grab data rather than `optimizely.get('data')` - for OptimizelyX these two calls will return different sets of data. The former will return `visitor` and `state` whereas the later will return `audiences`, `events`, `campaigns`, `pages` and `experiments`.

src/js/tracker.js
miike: I've only named it `variation` to be consistent with the existing `optimizely_summary` schema, happy to change it to `variationId` though if it makes more sense.

src/js/tracker.js
BenFradet: ah good to know! Then do we need `&& windowAlias.optimizely.data`?

src/js/tracker.js
BenFradet: True, let's leave it as it then, didn't check that `getOptimizelySummary` returned a variation my bad.

src/js/tracker.js
miike: It still exists - I'm not 100% sure if it's worth checking - but just doesn't return the same data as `get`.

BenFradet: Since we're not using it in this particular function, my opinion would be to remove the check, maybe @chuwy can chime in?

chuwy: If we're talking about `getOptimizelyData`, `windowAlias.optimizely.data` must not be deleted. This is a defensive check fixing https://github.com/snowplow/snowplow-javascript-tracker/issues/553. That bug can result in whole frontend freeze - if user defined somewhere empty `window.optimizely` object - our tracker will try access to  `undefined` (as `optimizely.data` is `undefined`) which will throw exception freezing whole execution thread.

BenFradet: but here we can remove it as we're not using it, right?

chuwy: Yes! Sorry, I didn't notice it is present in `getOptimizelyXData`.

miike: Should I remove the check entirely or replace with a check for `windowAlias.optimizely.get('data')`?

BenFradet: I'd say we're fine removing the check since we're checking `data` afterwards, what do you think @chuwy ?

chuwy: Yes, we're fine removing it. 

Just little a heads-up: are we sure that `window.optimizely.get(nonExistingProperty)` returns `undefined` and not `null`? I'm asking because in JS these are not equivalent: `null === undefined` and `null === null`, so `data !== undefined` may be non-protective enough if `optimizely.get` may return `null` (you may start to wonder how silly JS can be).

BenFradet: No mention in the doc: https://developers.optimizely.com/x/solutions/javascript/reference/index.html#reading-data

chuwy: In some other places optimizely happily returns `null`, that's why we need to be sure. Probably the only way to know it is to test on real-world website with OptimizelyX. @miike do you know any?

miike: @chuwy I've just given it a test and `optimizely.get('test')` returned `undefined` rather than `null`.

chuwy: Thanks @miike! It's fine then.

miike: Alright, I'll give this a final test on a testbed setup with Optimizely X tomorrow morning.

BenFradet: Great, then let's just remove the check!

miike: I've removed this check.

miike:
I've tested this locally using OptimizelyX and it works as expected. Something worth noting (perhaps in the wiki?) is that in order for variation names to be sent through as anything other than `null` you must untick the 'Mask descriptive names in project code and third-party integrations' in OptimizelyX -> Settings -> Privacy

alexanderdean:
Thanks @miike, appreciate the testing. By the way we met a user here in Amsterdam who will probably contribute the same support for Adobe Target: https://github.com/snowplow/snowplow-javascript-tracker/issues/569

@BenFradet can you make sure to capture Mike's point in the documentation for the tracker?

BenFradet:
@miike Thanks for the testing, that'll definitely go into the wiki.

PR #579, opened 2017-04-15
Tracking clicks on forms

bernardosrulzon:
An interesting form tracking use case is analyzing if users are actually selecting something when they open a drop-down menu, or typing something when they click on a text field. This would allow us to see which fields are adding more friction to the conversion funnel - and hopefully lower it over time :)

Thoughts?

### DIFF
--- a/core/lib/core.ts
+++ b/core/lib/core.ts
@@ -786,6 +786,44 @@ export function trackerCore(base64: boolean, callback?: (PayloadData) => void) {
 		},
 
 		/**
+		 * Track the action of a select form field being clicked
+		 *
+		 * @param formId The parent form ID
+		 * @param elementId ID of the changed element
+		 * @param nodeName "SELECT"
+		 * @param type Type of the changed element if its type is "INPUT"
+		 * @param elementClasses List of classes of the changed element
+		 * @param value The current value of the changed element
+		 * @param context Optional. Context relating to the event.
+		 * @param tstamp Optional. TrackerTimestamp of the event
+		 * @return Payload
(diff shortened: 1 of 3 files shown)


snowplowcla:
@bernardosrulzon has signed the Individual Contributor License Agreement

kingo55:
Hi Bernardo,

Form completion tracking is very useful. I think what you're after are the
on blur events. These events fire when users move away from a field.
They're the opposite of on focus.

At that point you should be able to detect if a selection or input was made
within a field.

Is that what you're looking for?

https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onblur

bernardosrulzon:
Hey Rob! Thanks, that was exactly what I was looking for - `focus` works perfectly for this use case.

Schema is here: https://github.com/snowplow/iglu-central/pull/548

core/lib/core.ts
```diff
@@ -800,7 +800,8 @@ export function trackerCore(base64: boolean, callback?: (PayloadData) => void) {
 		 *
 		 * @todo make `nodeName` enum
 		 */
-		trackFormChange: function (
+		trackFormFocusOrChange: function (
```
BenFradet: should we separate `trackFormChange` from `trackFormFocus`?

bernardosrulzon: I thought it would make sense to keep them together as they send exactly the same data. But feel free to edit this PR as you see fit - it was just a quick implementation to help us identify which forms needed more attention

Issue #548, opened 2017-01-19
Provide read-only access to some properties generated inside tracker object


chuwy:
Recentrly we encountered few cases where users wanted to get some properties that are encapsulated inside tracker object. Particularly, users wanted to get `cookieDomain` and `pageViewId` to use in app logic. It's possible to get `cookieDomain` with lengthy script, but seems not possible at all for `pageViewId`. 

I think we should provide some read-only access to at least these properties (and may be consider another properties) as getter-functions.

alexanderdean:
Nice idea, scheduling...

ryanrozich:
Our need for this has to do with click tracking. Currently we have our own click redirect and also have been using the javascript link tracking built into the javascript tracker. We have our script deployed across a wide variety of client sites.

We wanted to see whether the javascript tracker was counting as many clicks as our redirect so as a sanity check we took click count samples over multiple days from both our redirect and also from the snowplow reported clicks. On some sites they were close, and on other sites either our redirect had a much larger count of clicks than snowplow and on others it was the reverse.

Since we want to get the most complete count of clicks, our solution is to (a) continue tracking clicks from the javascript tracker and (b) use the clojure collector as a redirect and (c) rely on our data modeling scripts to deduplicate the clicks that appear in both. 

We planned on using the web_page.id (pageViewId) and the link target URL as the key for deduplication (we only need to track unique clicks, don't need to be able to count multiple clicks on the same URL for the same page view).

The problem we ran into was that, while the link_click event from the javascript tracker will append the web_page context automatically, we will need to manually add it to the click URL so that the redirect can pick it up as a context. Since we can't get this ID as a getter-function, we are stuck. Hopefully that provides more context. 

Do you have plans in the short-term to deliver this getter-function for the web_page context id? If not, it sounds like something that we could implement ourselves and submit a pull-request. It would be helpful though if you could point us to the right part of the code to make this change.

alexanderdean:
Moving forwards to 2.8.0

BenFradet:
The shortlist of things to expose seems to be:

- `pageViewId`
- `cookieDomain`

Are there any others I'm missing / someone wants to add?

alexanderdean:
Ping @yalisassoon @bogaert

bogaert:
Nothing comes to mind.

digdeep:
[...]

PR #610, opened 2017-09-28
Add getter for DomainSessionId

falschparker82:
Add getter for DomainSessionId as proposed in Issue #548

### DIFF
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -1604,6 +1604,15 @@
 			},
 
 			/**
+			 * Get the domain session ID also known as current memorized session ID.
+			 *
+			 * @return string Domain session ID
+			 */
+			getDomainSessionId: function() {
+				return memorizedSessionId;
+			},
+			
+			/**
 			 * Get the page view ID as generated or provided by mutSnowplowState.pageViewId.
 			 *
 			 * @return string Page view ID


chuwy:
Agree with @BenFradet. Also if #548 going to be implemented I would imagine a method returning object with many properties, rather than multiple methods for each property.

snowplowcla:

📝 Please visit https://github.com/snowplow/snowplow/wiki/CLA to learn more and sign.

zwacky:
@BenFradet according to these [release notes](https://snowplowanalytics.com/blog/2014/12/15/snowplow-javascript-tracker-2.2.0-released/) the `getDomainUserInfo()` it only prints out the domain user id, not domain session id. There is no other workaround than having a new function to retreive the domainSessionId.

BenFradet:
`getDomainUserid` only gives you back the domain user id.

[`getDomainUserInfo`](https://github.com/snowplow/snowplow-javascript-tracker/blob/3241605666cd29e9538b2bb31789b585c1b2969c/src/js/tracker.js#L1643-L1650) gives you back [the whole content of the domain user id cookie](https://github.com/snowplow/snowplow-javascript-tracker/blob/3241605666cd29e9538b2bb31789b585c1b2969c/src/js/tracker.js#L650-L690) which contains more than just the domain user id and in particular [the session id](https://github.com/snowplow/snowplow-javascript-tracker/blob/3241605666cd29e9538b2bb31789b585c1b2969c/src/js/tracker.js#L684-L687).

zwacky:
Thanks for your reply, I'm still at 2.6.2.
As of now the Domain Session ID is not the same that gets sent out in the pixel in my case ([image](https://i.imgur.com/qo3YN8V.png)).

I will try with the latest snowplow tracker in a bit. But maybe I'm confusing certain fields (?!)

BenFradet:
From the image, it seems that cookies are seen as disabled (we couldn't retrieve the content of the `id` cookie) and so the session id is a regenerated random uuid.

alexanderdean:
What was the conclusion on this one? Merge or close? @BenFradet @chuwy

BenFradet:
I tend to agree with @chuwy that we need to standardize what we're exposing through a single method which will make everything clear, closing.

Nonetheless, thanks a lot for your pull request @falschparker82

PR #614, opened 2017-10-19
warn if usePost is undefined (close #613)

### LINKED ISSUE #613: undefined usePost variable won't get set correctly in object.OutQueueManager
In the function object.OutQueueManager (src/js/out_queue.js) there is the following line:

```javascript
usePost = usePost && window.XMLHttpRequest && ('withCredentials' in new XMLHttpRequest());
```

If `usePost` is set to true or false, then you get the desired outcome. Not so if for some reason `usePost` is undefined. Then you just get undefined. 

At least give a warning that it's undefined perhaps?

chrissherlock:


chrissherlock:
Sorry, I mucked this one up.

PR #615, opened 2017-10-19
warn if usePost is undefined (close #613)

### LINKED ISSUE #613: undefined usePost variable won't get set correctly in object.OutQueueManager

chrissherlock:
Sorry about before, this patch should hopefully be better.

### DIFF
--- a/src/js/out_queue.js
+++ b/src/js/out_queue.js
@@ -62,7 +62,16 @@
 			outQueue;
 
 		// Fall back to GET for browsers which don't support CORS XMLHttpRequests (e.g. IE <= 9)
-		usePost = usePost && window.XMLHttpRequest && ('withCredentials' in new XMLHttpRequest());
+		if (typeof usePost != 'undefined')
+		{
+			usePost = usePost && window.XMLHttpRequest && ('withCredentials' in new XMLHttpRequest());
+			path = usePost ? '/com.snowplowanalytics.snowplow/tp2' : '/i';
+		}
+		else
+		{
+			Console.warn('usePost undefined');
+			path = '/i';
+		}
 
 		var path = usePost ? '/com.snowplowanalytics.snowplow/tp2' : '/i';
 
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -66,23 +66,21 @@
 	 * 4. appId, ''
 	 * 5. platform, 'web'
 	 * 6. respectDoNotTrack, false
-	 * 7. userFingerprint, true
-	 * 8. userFingerprintSeed, 123412414
-	 * 9. pageUnloadTimer, 500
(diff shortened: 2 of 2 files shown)


src/js/out_queue.js
```diff
@@ -62,7 +62,16 @@
...
+		if (typeof usePost != 'undefined')
+		{
+			usePost = usePost && window.XMLHttpRequest && ('withCredentials' in new XMLHttpRequest());
+			path = usePost ? '/com.snowplowanalytics.snowplow/tp2' : '/i';
+		}
+		else
+		{
+			Console.warn('usePost undefined');
+			path = '/i';
+		}
```
BenFradet: Couldn't we have 

```js
usePost = typeof usePost != 'undefined' && usePost && window.XMLHttpRequest && ('withCredentials' in new XMLHttpRequest());
```

What do you think @chuwy ?

chrissherlock: I would like to see a warning. One of the orgs who uses this has managed to encode an absolutely massive object and because this variable is undefined they are using GET requests... and thus is getting 414 errors because the URI is _enormous_... if they picked up on the fact that they hadn't set usePost, then it would have alerted them to the issue more quickly.

chuwy: But if I understand it right - if website-owner just wants to use GET-requests with reasonable payload - we'll force all of his users see "usePost undefined" message in console, which doesn't look right for me. So, this change is not about `undefined`-check it's more about `GET`-check. But correct me if I'm wrong. 

I'm quite against all log messages from tracker - in 99,9% of cases they hit console-aware users and not website-owners. But I'm quite keen to introduce something like debug-mode for tracker, where all such cases can be safely printed in development-environment and deleted in production. 

If we want to avoid 414-like issues - all we can do is to be more explicit in tracker documentation.

chuwy: Another way to fix it would be to use POST by default and allow to use GET by settings `useGet = true`, but this is quite big change for me, taking in account that CloudFront collector doesn't support POST.

snowplowcla:
@chrissherlock has signed the Individual Contributor License Agreement. Thanks so much

alexanderdean:
Agree with @chuwy - we won't add warnings into a production tracker. Debug mode is the way to go.

paulboocock:
This has been superseeded by `eventMethod` and it is no longer possible to get into a situation where usePost is `undefined`.

PR #636, opened 2018-02-20
Remove chromeFirstPaint from PerformanceTiming context (fixes #635)

### LINKED ISSUE #635: Remove chromeFirstPaint from PerformanceTiming context
Currently we rely on `chrome.loadTimes` to populate the `chromeFirstPaint` value for the `iglu:org.w3/PerformanceTiming` schema.

`loadTimes` is deprecated[1] in Chrome 64 and will be removed in Chrome 71 (release early in December 2018). Google recommend[2] replacing this with values from the [PaintTiming](https://w3c.github.io/paint-timing/) spec/API instead though the spec for this is still in draft.

Once this is more widely implemented in browsers this could belong in it's own context rather than replacing the current first paint values.

[1] https://developers.google.com/web/updates/2017/12/chrome-loadtimes-deprecated

miike:
This PR removes the dependency on the `chrome.loadTimes` API (fixes #635)

### DIFF
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -947,7 +947,7 @@
 				'navigationStart', 'redirectStart', 'redirectEnd', 'fetchStart', 'domainLookupStart', 'domainLookupEnd', 'connectStart', 
 				'secureConnectionStart', 'connectEnd', 'requestStart', 'responseStart', 'responseEnd', 'unloadEventStart', 'unloadEventEnd',
 				'domLoading', 'domInteractive', 'domContentLoadedEventStart', 'domContentLoadedEventEnd', 'domComplete', 'loadEventStart', 
-				'loadEventEnd', 'msFirstPaint', 'chromeFirstPaint', 'requestEnd', 'proxyStart', 'proxyEnd'
+				'loadEventEnd', 'msFirstPaint', 'requestEnd', 'proxyStart', 'proxyEnd'
 			];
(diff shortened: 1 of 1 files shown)


src/js/tracker.js
```diff
@@ -964,11 +964,6 @@
 				// Old Chrome versions add an unwanted requestEnd field
 				delete performanceTiming.requestEnd;
 
-				// Add the Chrome firstPaintTime to the performance if it exists
-				if (windowAlias.chrome && windowAlias.chrome.loadTimes && typeof windowAlias.chrome.loadTimes().firstPaintTime === 'number') {
-					performanceTiming.chromeFirstPaint = Math.round(windowAlias.chrome.loadTimes().firstPaintTime * 1000);
-				}
```
BenFradet: isn't it safe to leave it for now while it's being deprecated. Maybe during the time it's deprecated the new standard will be implemented by other vendors?

alexanderdean: What's the impact of leaving it @miike? The check looks pretty defensive

miike: This is safe to leave in at the moment as the behaviour won't change until it's removed in M70 - so this PR can just sit here until before then.

Chrome has implemented a [PerformancePaintTiming API](https://developer.mozilla.org/en-US/docs/Web/API/PerformancePaintTiming) since M60 but it's marked experimental and the spec is still in progress. No word on other browsers implementing anything yet but with any luck that will happen before November.

snowplowcla:
@miike has signed the Software Grant and Corporate Contributor License Agreement

alexanderdean:
De-scheduling pending clarification on #635

Issue #478, opened 2016-04-10
Use passive event listeners for mouse wheel event


orzage:
A mouse wheel event (mousewheel and DOMMouseScroll) listener, simply by being attached, causes significant performance issues.

See https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md for details on the problem.

That page also provides a proposal for passive event listeners, which work around the problem by making the event not cancelable (using preventDefault will not do anything useful.) Since SnowPlow does not use preventDefault, this should be a perfect fit. Passive event listeners are currently scheduled for release with Chrome 51 / Opera 38: https://www.chromestatus.com/feature/5745543795965952.

Finally, to prevent this from being an issue when passive event listeners are not available, the mouse wheel event listeners should either not be attached by default (at least unless passive event listeners are available) with a mechanism to enable them or, at the very least, should provide a means to declaratively disable these event listeners from ever being attached when not passive.

alexanderdean:
Thanks for raising @orzage ! We take any performance impacts of the Snowplow JavaScript Tracker on the hosting page very seriously. Assigning to @fblundun to review...

fblundun:
Thanks for pointing this out @orzage. A configuration option to disable those listeners when passive listeners are unavailable sounds like a good idea - we probably shouldn't make it disabled by default since this would break backwards compatibility. Do you know whether this also applies to `scroll` events?

orzage:
That sounds reasonable.

This does not apply to scroll events since they are not cancelable. If you are considering switching to using scroll events in place of the mouse wheel events, then yes, that could be a viable solution and a good compromise.

RByers:
That is correct - `wheel` and `touch*` events are the only problem here (`scroll` events are already asynchronous, don't actually block scrolling, though if they hog the CPU they can still be a perf issue of course).  Thanks for pushing on this!

pdufour:
Any update on this? Can all the "activity" handlers here https://github.com/snowplow/snowplow-javascript-tracker/blob/master/src/js/tracker.js#L1179-L1191 be marked as passive? Seeing this warning in the console `Handling of 'mousewheel' input event was delayed for 107 ms due to main thread being busy. Consider marking event handler as 'passive' to make the page more responive.`

alexanderdean:
Hi @pdufour - no update on this just yet, but we hope to get started on 2.7.0 in about 2 weeks' time.

yalisassoon:
[...]

PR #641, opened 2018-03-09
Use passive event listeners for mousewheel

OliverJAsh:
Probably needs re-formatting, but the logic should be sound.

Improves performance

Fixes https://github.com/snowplow/snowplow-javascript-tracker/issues/478

### DIFF
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -1483,7 +1483,7 @@
 					window.addEventListener('test', null, opts);
 				} catch (e) {}
 				helpers.addEventListener(documentAlias, 'mousewheel', activityHandler, supportsPassiveOption ? { passive: true } : undefined);
-				helpers.addEventListener(windowAlias, 'DOMMouseScroll', activityHandler, supportsPassiveOption ? { passive: true } : undefined);
+				helpers.addEventListener(windowAlias, 'DOMMouseScroll', activityHandler);
 				helpers.addEventListener(windowAlias, 'scroll', scrollHandler); // Will updateMaxScrolls() for us
 				helpers.addEventListener(documentAlias, 'keypress', activityHandler);
 				helpers.addEventListener(documentAlias, 'keydown', activityHandler);
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -1482,8 +1482,8 @@
 					});
 					window.addEventListener('test', null, opts);
 				} catch (e) {}
-				helpers.addEventListener(documentAlias, 'mousewheel', activityHandler, supportsPassiveOption ? { passive: true } : false);
-				helpers.addEventListener(windowAlias, 'DOMMouseScroll', activityHandler, supportsPassiveOption ? { passive: true } : false);
+				helpers.addEventListener(documentAlias, 'mousewheel', activityHandler, supportsPassiveOption ? { passive: true } : undefined);
+				helpers.addEventListener(windowAlias, 'DOMMouseScroll', activityHandler, supportsPassiveOption ? { passive: true } : undefined);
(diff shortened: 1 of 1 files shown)


BenFradet:
Thanks, I'll have a look on Monday :+1:

src/js/tracker.js
```diff
@@ -1472,8 +1472,18 @@
...
+				try {
+					var opts = Object.defineProperty({}, 'passive', {
+						get: function() {
+							supportsPassiveOption = true;
+						}
+					});
+					window.addEventListener('test', null, opts);
+				} catch (e) {}
+				helpers.addEventListener(documentAlias, 'mousewheel', activityHandler, supportsPassiveOption ? { passive: true } : undefined);
+				helpers.addEventListener(windowAlias, 'DOMMouseScroll', activityHandler, supportsPassiveOption ? { passive: true } : undefined);
```
OliverJAsh: Actually I'm not sure we need `DOMMouseScroll`, as it's already async apparently:

> The DOM DOMMouseScroll event is fired asynchronously when mouse wheel

https://developer.mozilla.org/en-US/docs/Web/Events/DOMMouseScroll

snowplowcla:

📝 Please visit https://github.com/snowplow/snowplow/wiki/CLA to learn more and sign.

OliverJAsh:
I signed it!

src/js/tracker.js
BenFradet: yup I think we could remove scroll events but wdyt about adding touch events?

OliverJAsh: As far as I can see, Snowplow isn't adding blocking event listeners for any of the touch events, so we have to need to make them passive.

I'll remove `DOMMouseScroll`.

OliverJAsh: Updated!

src/js/tracker.js
```diff
@@ -1472,7 +1472,18 @@
...
+				// Detect support for passive event listeners. Copied from:
+				// https://github.com/Modernizr/Modernizr/blob/a9369b031d05b90231b572e2796afba0334b21e6/feature-detects/dom/passiveeventlisteners.js#L26
+				var supportsPassiveOption = false;
+				try {
+					var opts = Object.defineProperty({}, 'passive', {
+						get: function() {
+							supportsPassiveOption = true;
+						}
+					});
+					window.addEventListener('test', null, opts);
```
mhadam: In EdgeHTML v15, a null callback produces an error. It doesn't conform to the DOM standard spec, but still best to avoid it. @BenFradet, is it a good idea to pull in this library instead? : https://github.com/rafrex/detect-passive-events

OliverJAsh: I'm surprised this behaviour wasn't caught by Modernizr, where I copied this detection from. We should patch Modernizr to fix this 🤔.

OliverJAsh:
> Though, wondering whether including a library rather than replicating the detection here is a good idea?

I just did this because it was the easiest change for me to do in a PR. I think adding a dependency would be a bit more of a involved change, but happy to do it if someone can walk me through it!

src/js/tracker.js
BenFradet: ah good point! I'm :+1: on using the lib

snowplowcla:
Confirmed! @OliverJAsh has signed the Individual Contributor License Agreement. Thanks so much

mhadam:
I've updated this under #478: 0c9b0f2264ec40162f2178c9bfb0dcc8a3ddce1d

OliverJAsh:
Superseded by https://github.com/snowplow/snowplow-javascript-tracker/issues/478 https://github.com/snowplow/snowplow-javascript-tracker/compare/issue/478

PR #674, opened 2018-10-09
Beacon API option for sending events to collector

nagyadam2092:
This pull request introduces the ability to send events using the Beacon API. Just like POST, one can define - using a config parameter - if she would like to send events using the Beacon API instead of POST or image beacons.


snowplowcla:

PR #676, opened 2018-10-12
Eliminating timezonejs

nagyadam2092:
This PR tries to shrink down the footprint of the tracker by removing unnecessary code from it.
What basically jstimezonedetect does is lookup some timezone difference information and then match that information in a huge dictionary (JS object / HashMap), and return the value of the desired key (e.g. "60,1" key will end up in "Europe, Berlin").
We don't need to have that HashMap on the client side, we could send only the timezone difference and have a HashMap on the enricher side, and have the lookup over there.
Effect on footprint (about 6K difference minified):
```
[with timezonejs]
total 1392
-rw-r--r--  1 anagy  staff   282K Oct 12 15:20 bundle.js
-rw-r--r--  1 anagy  staff   283K Oct 12 15:20 snowplow.js
-rw-r--r--  1 anagy  staff    75K Oct 12 15:20 sp.js

[without timezonejs]
total 1344
-rw-r--r--  1 anagy  staff   269K Oct 12 18:11 bundle.js
-rw-r--r--  1 anagy  staff   270K Oct 12 18:11 snowplow.js
-rw-r--r--  1 anagy  staff    69K Oct 12 18:11 sp.js
```

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -578,6 +578,7 @@
       "requires": {
         "anymatch": "1.3.2",
         "async-each": "1.0.1",
+        "fsevents": "1.2.4",
         "glob-parent": "2.0.0",
         "inherits": "2.0.3",
         "is-binary-path": "1.0.1",
@@ -1303,7 +1304,8 @@
         "esprima": "1.2.5",
         "estraverse": "1.9.3",
         "esutils": "2.0.2",
-        "optionator": "0.5.0"
+        "optionator": "0.5.0",
+        "source-map": "0.2.0"
       },
       "dependencies": {
         "esprima": {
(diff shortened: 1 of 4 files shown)


alexanderdean:
Interesting idea, but without the corresponding updated Enrichment this is removing data richness from Snowplow, and technically it's breaking the Snowplow Tracker Protocol v2 by changing one of the values in it.

anagytherealone:
Thx @alexanderdean for you insights !!
So do you think this change will never be able to land in the repo? :(

nagyadam2092:
Thx @alexanderdean for you insights !!
So do you think this change will never be able to land in the repo? :(

alexanderdean:
Never say never!

mhadam:
Going to close this, to be reopened if we revisit.

christoph-buente:
@alexanderdean i started a RFC discussion on discourse to further elaborate on that "improvement". https://discourse.snowplowanalytics.com/t/move-timezone-detection-to-backend/2492

PR #677, opened 2018-10-14
Feature config POC

nagyadam2092:
This PR introduces an approach where one could cherry pick features which she would like to include in the final bundle of the tracker.
I implemented only the form tracking configuration for now, but there are other features which could be included in the configuration file.
The approach: [`uglifyify`](https://www.npmjs.com/package/uglifyify) has the feature "dead code elimination" where e.g. `if (false)` statements (and blocks) are going to be removed from the final bundle file. This way the `.snowplow/config.json` file could be the point where one could set e.g. form tracking (`FORM_TRACKING_ENABLED `) to false, and this way it won't be included in the script because there is a check in the `tracker.js`: 
```
if ('__FORM_TRACKING_ENABLED__' === 'true')
```
which is going to be [replaced](https://github.com/outaTiME/grunt-replace) by the value of `FORM_TRACKING_ENABLED` in the `.snowplow/config.json` which will end up in a `if (false)` statement, which will be eliminated by [`uglifyify`](https://www.npmjs.com/package/uglifyify).
Possible further features which could be included in the `config.json` file:
- geolocation detection
- Parrable detection and usage
- Augur detection and usage
- Optimizely detection and usage
- link tracking
- link decoration
- logic regarding css classes in `helpers.js`
- filtering in `helpers.js`
- Yahoo / search engine detection
- pagepings
- GA cookies
- page scroll detection
- ecommerce transaction templating
There are probably missing features which could be included in the `config.json` and it might be that some of the features are so tightly coupled they could not be moved to the config easily.

### DIFF
--- a/.snowplow/config.json
+++ b/.snowplow/config.json
@@ -1,0 +1,3 @@
+{
+  "FORM_TRACKING_ENABLED": false
+}
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -33,6 +33,7 @@
 */
 
 var semver = require('semver');
+var CONFIG = require('./.snowplow/config.json');
 
 /*global module:false*/
 module.exports = function(grunt) {
@@ -93,6 +94,9 @@ module.exports = function(grunt) {
       main: {
         files: {
           'dist/bundle.js': ['src/js/init.js']
+        },
+        options: {
+          transform: ['uglifyify']
         }
       },
       test: {
@@ -101,6 +105,23 @@ module.exports = function(grunt) {
           'tests/pages/detectors.js': ['tests/scripts/detectors.js'],
           'tests/pages/snowplow.js': ['src/js/init.js']
         }
+      }
+    },
+
+    replace: {
+      dist: {
+          options: {
+              patterns: [{
+                  match: /false/g,
+                  replacement: CONFIG.FORM_TRACKING_ENABLED
+              }
+              ]
+          },
+          files: [{
+              expand: true,
+              src: '*.js',
+              dest: "./"
+          }]
       }
     },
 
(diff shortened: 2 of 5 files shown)


snowplowcla:

max-tgam:
@colmsnowplow with releases in 2.11 the features described in this POC only become more relevant. We do not use about 30-40% of the features described above and having a tracker file that loads faster and uses less server/CDN resources is super helpful.

This and related PR have seen no action for almost a year now, what would it make to move it forward?

max-tgam:
Not sure how I unassigned someone from this as I just commented on the PR...

goodits:
Hi @max-tgam, this is firmly on the radar. We've got a pretty packed Q4 but we're growing our team over the next couple of months so will keep this in mind. It'll need a good amount of testing, which sits nicely with out planned efforts over Q4 to improve the tests suites of our most popular trackers. .

max-tgam:
@goodits That is good to hear. It would be nice to have this built into snowplow instead of us prepping Snowplow source code to remove parts we do not need (like jstimezonedetect library and some of items above)

alexanderdean:
@christoph-buente is this the PR you were talking about on Thurs?

paulboocock:
To give another update on this. 
We're beginning to move towards v3 of the JS Tracker, this will include a variety of features that are going to help with ensuring the tracker is as small as can be.

We're very likely going to move away from grunt and towards rollupjs (#752). This will allow us to achieve similar results to what you have proposed here, utilising Tree Shaking and @rollup/plugin-replace we can make it easy to elimate parts of the codebase on build. We also intend to make the code base more modular and support ES Modules, allowing for unused code to be removed when you import the library into a web app (react, angular, etc.).

To ensure we don't loose this idea for v3, I've opened a new issue #792 to track our progress on this concept but with our new bundling process that is on the horizon.

PR #678, opened 2018-10-15
Opt-out form tracking

nagyadam2092:
- finalised version of https://github.com/snowplow/snowplow-javascript-tracker/pull/677
- removed grunt-contrib-uglify ([last updated 4 years ago, 186 weekly downloads](https://www.npmjs.com/package/grunt-yui-compressor))
- extracted form tracking logic
  - using UglifyJS's dead code elimination to make cherry picking available for features
  - in this PR form tracking extraction is only implemented, but as I mentioned in https://github.com/snowplow/snowplow-javascript-tracker/pull/677 there are plenty more features to be extracted this way
- the key idea is to have a `.snowplow` folder in the root of a project and if there are defined features inside `config.json` these will be updated and used in the build process so one could eliminate some features from the final version of the tracker
  - this means if you have snowplow JS tracker as a dependency you only have to have this `.snowplow` folder under your project root
- possible consequence of the dead code elimination is that lodash build process could be eliminated as well
- minified weight loss: 2 kB

Please let me know your thoughts!


snowplowcla:

src/js/tracker.js
```diff
@@ -2666,7 +2641,34 @@
...
+							mutSnowplowState.registeredOnLoadHandlers.push(function () {
+									formTrackingManager.configureFormTracking(config);
+									formTrackingManager.addFormListeners(context);
+							});
+					}
+			};
+		}
+
+		return API;
+
```
BenFradet: The indent seems to be off.

Also does this replace #677 ? If so you can close out the other PR :+1:

paulboocock:
See #677

PR #679, opened 2018-10-16
Opt-out link tracking

nagyadam2092:
- Extracted logic for link tracking
- Could be excluded from the final minified version of the tracker
- minified weight loss: 1.65 kB


snowplowcla:

paulboocock:
See #677

PR #696, opened 2018-12-10
Add option to set a custom POST request path

jaredsilver:
uBlock Origin, one of the world's most popular adblockers, blocks Snowplow.

In particular, it blocks patterns like: `/snowplow/`, `/snowplow.js`, etc.

The path that the Snowplow JavaScript tracker hits -- `/com.snowplowanalytics.snowplow/tp2` -- is a pattern that gets matched. As a result, anyone relying on this tracker is likely losing a substantial amount of data. (On DataCamp.com, for example, we noticed that we were losing almost 10% of our front-end Snowplow data because of this.)

I propose adding the ability for users of the Snowplow JavaScript tracker to change that path. This would allow users to change it to something that will not be blocked by adblockers. Further, it would allow users to use more advanced techniques like cycling the path so that adblockers have a tougher time catching on.

Obviously, updating this path alone will not fix the problem -- in fact, it'll break most people's Snowplow implementations because the path they're setting does not actually exist on the other end. However, that's a separate problem. (Our plan is to use URL masking instead of fiddling with the server.)

Please note that this code hasn't been tested at all -- I just threw this together as quickly as I could to get a conversation going. Hopefully a maintainer can take it from here and get this into master ASAP. A lot of people are losing a lot of data in the meantime.

### DIFF
--- a/src/js/out_queue.js
+++ b/src/js/out_queue.js
@@ -53,7 +53,7 @@
 	 * @param int bufferSize How many events to batch in localStorage before sending them all.
 	 *                       Only applies when sending POST requests and when localStorage is available.
 	 * @param int maxPostBytes Maximum combined size in bytes of the event JSONs in a POST request
-	 * @param string path The path
+	 * @param string postPath The path where events are to be posted
 	 * @return object OutQueueManager instance
 	 */
 	object.OutQueueManager = function (functionName, namespace, mutSnowplowState, useLocalStorage, usePost, postPath, bufferSize, maxPostBytes) {
--- a/src/js/out_queue.js
+++ b/src/js/out_queue.js
@@ -53,9 +53,10 @@
 	 * @param int bufferSize How many events to batch in localStorage before sending them all.
 	 *                       Only applies when sending POST requests and when localStorage is available.
 	 * @param int maxPostBytes Maximum combined size in bytes of the event JSONs in a POST request
+	 * @param string path The path
 	 * @return object OutQueueManager instance
 	 */
-	object.OutQueueManager = function (functionName, namespace, mutSnowplowState, useLocalStorage, usePost, bufferSize, maxPostBytes) {
+	object.OutQueueManager = function (functionName, namespace, mutSnowplowState, useLocalStorage, usePost, postPath, bufferSize, maxPostBytes) {
 		var	queueName,
 			executingQueue = false,
 			configCollectorUrl,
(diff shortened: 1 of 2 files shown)


snowplowcla:

:memo: Please visit https://github.com/snowplow/snowplow/wiki/CLA to learn more and sign.

jaredsilver:
@snowplowcla CLA signed!

snowplowcla:
Confirmed! @jaredsilver  has signed the Individual Contributor License Agreement. Thanks so much

jaredsilver:
In the long run, yes that would be beneficial. But in the short term, Snowplow users can very quickly use a URL mask without needing to update anything on the enrichment platforms’ side. This quick fix alone would be immensely valuable for a lot of users. But in the long term, it would be better if the enrichment platforms also supported this functionality out of the box.

jaredsilver:
@BenFradet Any updates on this?

BenFradet:
ping @mhadam

mhadam:
I'm trying to avoid adding stuff last minute to the coming release since it's pretty much finalized. I can see what I can do though - we have to be careful to make sure this plays nicely with Beacon API.

BenFradet:
I think this can definitely be pushed back to another release, this was just about getting your feedback

jaredsilver:
@mhadam @BenFradet I appreciate you seeing what you can do. If it makes any difference, some of the other engineers and data scientists at our company have started exploring alternatives because of how long it's taken to achieve a reasonable level of data integrity with Snowplow. These sorts of issues seem to keep popping up.

goodits:
Hi @jaredsilver. Thank you for highlighting the urgency your side and we appreciate you'd like this merged as soon as possible. We're finding a way to work it in to the upcoming release. Hoping not to keep you waiting much longer, but we'll will give you an ETA Monday. Have a great weekend!

jaredsilver:
@goodits Great, thanks so much for the update! Please let me know if there's anything I can do to help.

mhadam:
@jaredsilver we plan to get this feature out in a small release (2.9.3) within the next few days

jaredsilver:
Thanks, @mhadam! We very much appreciate it. Please let me know if there's anything we can do to assist.

mhadam:
@jaredsilver this is released here, by the way: http://d1fc8wv8zag5ca.cloudfront.net/2.9.3/sp.js

jaredsilver:
Fantastic, thanks so much @mhadam! We really appreciate the extra effort to get this out.

PR #698, opened 2019-01-02
ES6 / Babel / Mocha / Others

mmathias01:
This is a rewrite / re-factor of the main tracker into ES6. This is not even close to final but I wanted to submit the PR now so that it if this is something you guys are considering there is no duplication of work. Here is a rough summary of what I did:

- Convert to ES6
- Convert build process to use rollup.js
- Add Babel
- Tried to fixup the JSDoc syntax wherever I could. 
- Mostly remove Lodash dependencies
- Add Mocha and rewrite functional tests

There is still a decent amount to do. The built/compiled tracker works, however there are a lot of optimizations to the code that should happen. I also have a bunch of things I used for testing that should eventually be pulled out. Some additional things I would like to focus on would be:

- More unit tests / better code coverage
- Use JSDom to simulate more DOM / Browser things for unit tests
- Update many methods / functions to use more efficient ES6 code
- Update build process / docs
- Modularize (pull out the code for optimizely etc. into plugins to reduce the size of the main bundle)
- Update library to use fetch (or fetch polyfill) / beacon API

Once this ES6 conversion is in place it makes fixing some of the outstanding bugs and creating new plugins / functionality much easier. Also having 100% solid unit tests would probably be a huge plus. Anyways this is mostly to start the discussion, just trying to help however I can. 

Cheers.


mmathias01:
this did not pull the stuff from my fork for some reason.... closing ugh.

PR #700, opened 2019-01-02
Consolidate request method API

mmathias01:
A PR to address the comment I made earlier. 

My concern in a nutshell is that its probably not a good idea to keep adding arguments to the argmap that might be confusing and/or contradictory. Having two options one for post, and one for beacon makes it unclear which one takes precedent, and makes it harder to maintain argmap backwards compatibility in the future. This PR creates an eventMethod (not 100% sold on this name) property on the argmap that is future proof, while still respecting the post setting for the time being.

### DIFF
--- a/src/js/out_queue.js
+++ b/src/js/out_queue.js
@@ -51,20 +51,28 @@
 	 * @param object mutSnowplowState Gives the pageUnloadGuard a reference to the outbound queue
 	 *                                so it can unload the page when all queues are empty
 	 * @param boolean useLocalStorage Whether to use localStorage at all
-	 * @param boolean usePost Whether to send events by POST or GET
+	 * @param string eventMethod if null will use 'beacon' otherwise can be set to 'post', 'get', or 'beacon' to force. 
 	 * @param int bufferSize How many events to batch in localStorage before sending them all.
 	 *                       Only applies when sending POST requests and when localStorage is available.
 	 * @param int maxPostBytes Maximum combined size in bytes of the event JSONs in a POST request
 	 * @return object OutQueueManager instance
 	 */
-	object.OutQueueManager = function (functionName, namespace, mutSnowplowState, useLocalStorage, useBeacon, usePost, bufferSize, maxPostBytes) {
+	object.OutQueueManager = function (functionName, namespace, mutSnowplowState, useLocalStorage, eventMethod, bufferSize, maxPostBytes) {
 		var	queueName,
 			executingQueue = false,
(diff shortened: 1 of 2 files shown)


snowplowcla:
Confirmed! @mmathias01  has signed the Individual Contributor License Agreement. Thanks so much

mhadam:
Thanks! This makes sense and everything looks good. I'll slip this into the release.

mmathias01:
Cool! Thanks!

src/js/out_queue.js
```diff
@@ -51,20 +51,28 @@
...
-		useBeacon = useBeacon && navigator && navigator.sendBeacon;
-
+		
+		//Force to lower case if its a string
+		eventMethod = eventMethod.toLowerCase ? eventMethod.toLowerCase() : eventMethod;
+		
+		//Use the Beacon API if eventMethod is set null, true, or 'beacon'. 
+		var enableBeacon = (eventMethod === null || eventMethod === true || eventMethod === "beacon" || eventMethod === "true") ? true : false;
+		// Fall back to POST or GET for browsers which don't support Beacon API
+		useBeacon = enableBeacon && navigator && navigator.sendBeacon;
```
mhadam: this needs to be declared (I fixed it before deploying, but just pointing it out)

src/js/out_queue.js
```diff
@@ -51,20 +51,28 @@
...
 			outQueue;
-
-		useBeacon = useBeacon && navigator && navigator.sendBeacon;
-
+		
+		//Force to lower case if its a string
+		eventMethod = eventMethod.toLowerCase ? eventMethod.toLowerCase() : eventMethod;
+		
+		//Use the Beacon API if eventMethod is set null, true, or 'beacon'. 
+		var enableBeacon = (eventMethod === null || eventMethod === true || eventMethod === "beacon" || eventMethod === "true") ? true : false;
```
mhadam: this ternary operator is redundant as far as I can tell

src/js/out_queue.js
```diff
@@ -51,20 +51,28 @@
...
+		// Fall back to POST or GET for browsers which don't support Beacon API
+		useBeacon = enableBeacon && navigator && navigator.sendBeacon;
+
+		//Use POST if specified, or beacon is unavailable. 
+		var usePost = (eventMethod === "post" || (enableBeacon && !useBeacon)) ? true : false;
```
mhadam: likewise, doesn't need a ternary operator

PR #709, opened 2019-02-05
Fix #708 "Beacon tracking doesn't seem to flush OutQueue"

### LINKED ISSUE #708: Beacon tracking doesn't seem to flush OutQueue
Hi again!

So it looks like because of the way the POST method is set up, the `outQueue` is "flushed" (events that were sent are removed/shift()ed from the queue)  only in the `onreadystatechange` handler.

```
for (var deleteCount = 0; deleteCount < numberToSend; deleteCount++) {
    outQueue.shift();
}
```

Because the beacon API hands off control to the browser, there's no `onreadystatechange` or similar callback - you just make the `sendBeacon()` call, and get the `true` back, which is the browser telling you it queued the call.

As a result, I've had instances where the outQueue just increases to infinite length over time, since it's saved to localStorage and then picked back up

The solution seems to be break this out into logic that's called from both places - [let me know if my solution (Pull Request #709) works for you](https://github.com/snowplow/snowplow-javascript-tracker/pull/709)

ScottyMJacobson:


snowplowcla:

:memo: Please visit https://github.com/snowplow/snowplow/wiki/CLA to learn more and sign.

ScottyMJacobson:
@snowplowcla I signed it!

ScottyMJacobson:
_(closed because i was on the wrong branch - sorry!)_

PR #710, opened 2019-02-06
Fix #708 "Beacon tracking doesn't seem to flush OutQueue"

### LINKED ISSUE #708: Beacon tracking doesn't seem to flush OutQueue

ScottyMJacobson:
(on master this time - sorry!)

Should close #708

### DIFF
--- a/src/js/out_queue.js
+++ b/src/js/out_queue.js
@@ -270,16 +270,22 @@
 				// Keep track of number of events to delete from queue
 				var numberToSend = chooseHowManyToExecute(outQueue);
 
+				// The events (`numberToSend` of them), have been sent, so we remove them from the outQueue
+				// We also call executeQueue() again, to let executeQueue() check if we should keep running through the queue
+				function onPostSuccess(numberToSend) {
+					for (var deleteCount = 0; deleteCount < numberToSend; deleteCount++) {
+						outQueue.shift();
+					}
+					if (useLocalStorage) {
+						helpers.attemptWriteLocalStorage(queueName, JSON.stringify(outQueue));
+					}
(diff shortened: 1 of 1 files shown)


snowplowcla:
@ScottyMJacobson  has signed the Software Grant and Corporate Contributor License Agreement

ScottyMJacobson:
@mhadam sorry for the merging mess (this time, rather than cherry-picking, I accidentally merged develop into master but unstaged the changes I didn't want. sourcetree looks how i wanted it to be (just the changes to out_queue) but github appears to be treating it like a merge (which it was))

Let me know if I should blow it up and submit one more pull request / go for third-time's-the-charm (or if you want to take it yourself)

mhadam:
No worries, I usually take over and cherry-pick stuff into the develop branch, so PR onto `develop` in the future. That's where I stage things for release.

PR #730, opened 2019-07-09
Fix global variables in optimizelyXSummary

jethron:
The global vars seem to be confusing the minifier in some way in newer builds, and the uglified code throws an exception trying to assign to an undefined variable, which results in the events being discarded. Making the variables local resolves the issue.

Tested & fixed with v2.10.2


jethron:
Closed/reopened to rerun CI tests (failed due to Intern issues).

Issue #744, opened 2019-09-24
Add function to allow setting Useragent


miike:
Add a method to the coreTracker to allow setting the useragent from trackers such as NodeJS.

miike:
cc @jethron

PR #745, opened 2019-09-24
Function to allow setting Useragent addresses #744

miike:
Allows the useragent to be set via Javascript allowing environments such as the Node JS tracker to set the useragent.


paulboocock:
LGTM @miike 
I'll pull this into a new Core release alongside our upcoming 2.12.0 release of the JS tracker.

PR #753, opened 2019-10-08
Get some real integration testing on the go with snowplow micro

jamessnowplow:
This work is all around the integration testing setup.

The good:

I've replaced the mock collector with a snowplow micro instance fired up by travis (and locally by hand). This meant re-wiring a bunch of the integration tests to use valid schemas, I picked some at random from the list on iglucentral.

The bad:

The way we populate the tunnel endpoint via a search and replace in the build is pretty wonky. I've proliferated the wonkyness in `micro-template.js`. There has to be a better way, something around env vars im betting.

I attempted to add a few modern browsers to the testing setup but all I could manage was edge. It looks like we have some issues with intern/selenium/sauce versions. I'm going to hazard a guess updating everything might help.

The discussion point:

I've used `lodash/fp`. It comes as default in lodash these days and has some really nice bits. I'm not tied to it however, its just a preference. So feel free to have a problem with it.

### DIFF
--- a/tests/pages/integration-template.html
+++ b/tests/pages/integration-template.html
@@ -69,9 +69,9 @@
   window.snowplow('setUserId', 'Malcolm');
   window.snowplow('trackPageView', 'My Title', [ // Auto-set page title; add page context
     {
-      schema: "iglu:com.example_company/user/jsonschema/2-0-0",
+      schema: "iglu:org.schema/WebPage/jsonschema/1-0-0",
       data: {
-        userType: 'tester'
+        keywords: ['tester']
       }
     }
   ]);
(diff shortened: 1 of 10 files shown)


jamessnowplow:
removed in favour of the main repo one

tests/pages/integration-template.html
```diff
@@ -118,29 +118,29 @@
   window.snowplow('trackTrans');
 
   var testAcceptRuleSet = {
-    accept: ['iglu:com.acme_company/*/jsonschema/*-*-*']
+    accept: ['iglu:com.snowplowanalytics.snowplow/ad_impression/jsonschema/*-*-*']
```
colmsnowplow: If we can, it would be good to preserve the rule having a wildcard mid-path for at least one of the rulesets. 

Can imagine a scenario where someone makes an innocuous-looking change which messes with on that front. This is one that would have a big impact if not caught till production since they might be attaching global contexts to a lot of different events.

Let me know if there's some challenge to doing this that I don't see - if we need to we can leave this as is and just log an issue to solve for this when we improve the tests themselves.

colmsnowplow:
#754 Replaces this one.

PR #754, opened 2019-10-08
Add Snowplow Micro tests

jamessnowplow:
This work is all around the integration testing setup.

The good:

I've replaced the mock collector with a snowplow micro instance fired up by travis (and locally by hand). This meant re-wiring a bunch of the integration tests to use valid schemas, I picked some at random from the list on iglucentral.

The bad:

The way we populate the tunnel endpoint via a search and replace in the build is pretty wonky. I've proliferated the wonkyness in micro-template.js. There has to be a better way, something around env vars im betting.

I attempted to add a few modern browsers to the testing setup but all I could manage was edge. It looks like we have some issues with intern/selenium/sauce versions. I'm going to hazard a guess updating everything might help.

The discussion point:

I've used lodash/fp. It comes as default in lodash these days and has some really nice bits. I'm not tied to it however, its just a preference. So feel free to have a problem with it.

### DIFF
--- a/tests/pages/integration-template.html
+++ b/tests/pages/integration-template.html
@@ -69,9 +69,9 @@
   window.snowplow('setUserId', 'Malcolm');
   window.snowplow('trackPageView', 'My Title', [ // Auto-set page title; add page context
     {
-      schema: "iglu:com.example_company/user/jsonschema/2-0-0",
+      schema: "iglu:org.schema/WebPage/jsonschema/1-0-0",
       data: {
-        userType: 'tester'
+        keywords: ['tester']
       }
     }
   ]);
@@ -81,9 +81,9 @@
 
   window.snowplow('trackStructEvent', 'Mixes', 'Play', 'MRC/fabric-0503-mix', '', '0.0');
   window.snowplow('trackUnstructEvent', {
-    schema: 'iglu:com.acme_company/viewed_product/jsonschema/5-0-0',
+    schema: 'iglu:com.snowplowanalytics.snowplow/ad_impression/jsonschema/1-0-0',
     data: {
-      productId: 'ASO01043'
+      bannerId: 'ASO01043'
     }
   }, [], {type: 'ttm', value: 1477401868});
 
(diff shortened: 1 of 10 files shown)


tests/pages/integration-template.html
```diff
@@ -118,29 +118,29 @@
   window.snowplow('trackTrans');
 
   var testAcceptRuleSet = {
-    accept: ['iglu:com.acme_company/*/jsonschema/*-*-*']
+    accept: ['iglu:com.snowplowanalytics.snowplow/ad_impression/jsonschema/*-*-*']
```
colmsnowplow: If we can, it would be good to preserve the rule having a wildcard mid-path for at least one of the rulesets. 

Can imagine a scenario where someone makes an innocuous-looking change which messes with on that front. This is one that would have a big impact if not caught till production since they might be attaching global contexts to a lot of different events.

Let me know if there's some challenge to doing this that I don't see - if we need to we can leave this as is and just log an issue to solve for this when we improve the tests themselves.

colmsnowplow:
👍 This is great @jamessnowplow , thanks. I added one comment that's not a blocking issue. 

Could you rename your commit to match [this issue](https://github.com/snowplow/snowplow-javascript-tracker/issues/755)? 

We follow a 1:1 map between issues and commits - you can see examples in the commit history.

On your comments:

> The way we populate the tunnel endpoint via a search and replace in the build is pretty wonky. I've proliferated the wonkyness in micro-template.js. There has to be a better way, something around env vars im betting.

Yeah would prefer something better but it's not the end of the world tbh. 

> I attempted to add a few modern browsers to the testing setup but all I could manage was edge. It looks like we have some issues with intern/selenium/sauce versions.

Yes - I can add some context to this - basically the later version of saucelabs changes how the requests are made, so as a quick workaround I had to lock in the saucelabs version to an old one. Hoping this isn't a major headache to get around.

> I've used lodash/fp. It comes as default in lodash these days and has some really nice bits. I'm not tied to it however, its just a preference. So feel free to have a problem with it.

No issue whatsoever, we're using it in some of our closed source projects too.

tests/pages/integration-template.html
jamessnowplow: Good plan. I think the mid path just got lost in my search and replace frenzy.

I've put them both back and everything seems to still be ok.

colmsnowplow:
👍 LGTM - if you could just squash the commits so we have 1:1 commit:issue we can merge to develop. 

Very happy with this, thanks James!

colmsnowplow:
(We can squash later if we like too, develop is just for staging.)

jamessnowplow:
no worries, i like a tidy history too.

PR #757, opened 2019-10-15
Package updates and test harness updating

jamessnowplow:
Quite a lot in here. We should talk it through when you have had a chance to look.

I have gone for a two tool testing setup. Try as I might none of the jest+<browser tool> seemed like they were going to work nicely in the long run so I have gone for a two tier approach.

Units:
Tested with jest via `npm run tests:unit`

Functional and Integration:
Tested using webdriver.io. It has nice sauce integration and caused the least pain setting up and integrating.

The way micro integrates has also changed a chunk. Webdriver runs a bunch of tests in parallel which is awesome but gives us problems with the single micro instance approach. By wrapping the micro init and shutdown into each integration spec using `micro.js` we can make sure tests don't cause each other problems.

Ngrok is gone. Sauce connect with a bit of help (dns localhost aliases) covers the same issue.


tests/wdio.sauce.conf.js
```diff
@@ -0,0 +1,89 @@
...
+      platform: 'Windows 8',
+      version: '10',
+    },
+    {
+      browserName: 'internet explorer',
+      platform: 'Windows 7',
+      version: '9',
+    },
+    /* note the change from platform to platformName
+     * that goes with the selenium version switch
```
colmsnowplow: Keen to understand the selenium version switch when we catch up.

tests/wdio.sauce.conf.js
```diff
@@ -0,0 +1,89 @@
...
+     */
+    {
+      browserName: 'safari',
+      browserVersion: 'latest',
+      platformName: 'macOS 10.14',
+      'sauce:options': { seleniumVersion: '3.14.0' },
+    },
+    /* this fails on a click handler in helpers.js
```
colmsnowplow: Let's also make sure to cover this too & evaluate if it's indicative of something we should look into.

colmsnowplow:
👍 LGTM. As discussed on the call we need to investigate/resolve the two failing integration tests before merge to prod, but this is good to merge to develop.

PR #771, opened 2019-10-28
Fix tracker continuing to insert events up to max Local Storage quota

paulboocock:
Introduced new initialisation parameter for the tracker. 
`maxLocalStorageQueueSize`

This controls how many events the tracker will attempt to place onto the local storage queue if local storage has been selected as the stateStorageStrategy. Defaulted to 1000, as events normally range from 1kb to 2kb, meaning the tracker shouldn't use more than 1mb to 2mb of Local Storage and as most browsers have a default of 5mb for the local storage quota this should still leave headroom for other keys to be stored in local storage. The value can be lowered if the site is sending particularly large events and is using local storage.

### DIFF
--- a/src/js/out_queue.js
+++ b/src/js/out_queue.js
@@ -57,10 +57,11 @@
 	 *                       Only applies when sending POST requests and when localStorage is available.
 	 * @param int maxPostBytes Maximum combined size in bytes of the event JSONs in a POST request
 	 * @param boolean useStm Whether to add timestamp to events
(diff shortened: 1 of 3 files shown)


paulboocock:
That is correct @colmsnowplow - the outQueue will grow in an attempt to send the event (either in this batch or the next) but if it fails then we will only write (in your example) the first 100 outQueue events back to local storage, effectively discarding the latest event.

PR #779, opened 2020-01-13
Issue/750 reset page ping timer

paulboocock:
Changed default behaviour to fix a bug, where in a Single Page Application if you track a new Page View event, the Page Ping timers do not reset back to 0 and continue pinging the collector.

Now if you send a Page View event, the Page Pings will reset back to as they would with the first Page View (maintaining both the initial page ping delayed event and then the heartbeat page ping events).

As this is a change in behaviour that has existed for a while, I have introduced a new config flag `resetActivityTrackingOnPageView` that can be used to maintain previous behaviour (in case anyone has built modelling steps around this 'bugged' logic).

### DIFF
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -1609,19 +1609,31 @@
 				helpers.addEventListener(windowAlias, 'resize', activityHandler);
 				helpers.addEventListener(windowAlias, 'focus', activityHandler);
 				helpers.addEventListener(windowAlias, 'blur', activityHandler);
-
-				// Periodic check for activity.
-				lastActivityTime = now.getTime();
 			}
 
-			if (activityTrackingConfig.activityTrackingEnabled && pagePingInterval === null && activityTrackingInstalled) {
+			// Periodic check for activity.
+			lastActivityTime = now.getTime();
+
+			//Clear page ping heartbeat on new page view
+			if (pagePingInterval !== null) {
+				clearTimeout(pagePingInterval);
+				pagePingInterval = null;
+			}
+
+			if (activityTrackingConfig.activityTrackingEnabled && activityTrackingInstalled) {
 				pagePingInterval = activityInterval({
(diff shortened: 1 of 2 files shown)


src/js/tracker.js
```diff
@@ -1554,9 +1549,11 @@
 
 			// Send ping (to log that user has stayed on page)
 			var now = new Date();
+			var installingActivityTracking = false;
```
colmsnowplow: Do we need `installingActivityTracking` as distinct from `activityTrackingInstalled`? 

We set both to false globally first, then at the same time we set them both to true if activity tracking is enabled, & we don't use them in conjunction - so I can't see where one would be true when the other is false.

colmsnowplow:
LGTM in general - just one question in the comments.

src/js/tracker.js
paulboocock: `installingActivityTracking` is defined in scope of `logPageView`, whereas `activityTrackingInstalled` is declared in for the `Tracker`. To maintain the old functionality of only starting page pings one the first page view, I need another boolean that only gets set to true on the first `logPageView()` (`installingActivityTracking` is only true when `activityTrackingInstalled` is false on first `logPageView()`, but it is false on subsequent called where `activityTrackingInstalled` is true). If I used `activityTrackingInstalled` then I would end up resetting the page ping timers on every `logPageView` which I only want to do if `resetActivityTrackingOnPageView` is true (new default).

I'll likely rip this out with the next MAJOR revision we make and reset page pings on every page view but as we want this to be a MINOR revision, I needed a way to maintain backwards compatibility in case anyone is relying on this (albeit broken imo) behaviour.

colmsnowplow: Ah, I see where my confusion came from - I didn't follow the indentation properly. 

Yeah, this all makes a whole heap of sense, agreed on all fronts. Thanks for explaining!

Issue #751, opened 2019-10-04
Beacon implementation fails in Chrome and Safari 12+ due to browser bugs


mgrahamx:
Hey guys,

The implementation of Beacon API support in this library appears to be impacted by two separate browser bugs.

https://github.com/snowplow/snowplow-javascript-tracker/blob/ed751aaf8486f72d0abc0f0f880754c3dd61aa65/src/js/out_queue.js#L306

Here you are sending a Beacon request with JSON as the type, which is not in Chrome's CORS pre-approved headers list. This causes Chrome to throw an exception when you try to send a beacon this way. Of course this means it will fallback to POST and correctly log the event.

ref: https://bugs.chromium.org/p/chromium/issues/detail?id=490015

In Safari 12 & 13, it appears the same bug is present however it will silently fail and send a blank request body. This is a much bigger problem because it doesn't fallback and no tracking data is actually sent.

mgrahamx:
Have not tested Firefox, but I suspect the underlying security vulnerability that causes this bug has either been fixed or it throws similarly to Chrome. We aren't seeing Firefox tracking issues.

phliem:
I had the exact same problem. On Chrome 77.0 and Safari 13.0.3. Firefox was working perfectly though. The events were sent properly to snowplow.

Not sure if it is the best way to fix this, but changing the `eventMethod` to no use Beacon works for me. Just in case someone else needs to fix the events on chrome too here what I did.

On initialisation in the argmap, change the `eventMethod` to `get` or `post` 
Here is an example
```
snowplow("newTracker", "WebTracker", snowplowCollectorUrl, {
    appId: "Webapp",
    platform: "web",
    eventMethod: "post"
});
```

paulboocock:
We've done some investigation into this and we can confirm it's an issue effecting browsers based on Chromium (Chrome, Opera, Edge Beta) however events are still sent using POST. 

This issue doesn't effect Firefox as it sends a preflight and then correctly sends the request.

As the Chromium based browsers throw an exception, which we are handling, we fall back to POST requests and events are then sent. There isn't much we can do until the Chromium team improve their handling of beacon requests, we're are now tracking https://bugs.chromium.org/p/chromium/issues/detail?id=724929 which should fix this issue in Chromium.

[...]

PR #781, opened 2020-01-16
Update beacon support to handle "gotchas" (close #716)

### LINKED ISSUE #716: Update beacon support to handle "gotchas"
https://robertsahlin.com/analytics-beacon-transport-mechanism-gotchas/

paulboocock:
#716 #751 Now fires the first event with POST to ensure any CORS preflight requests have been sent by the browser before we start using Beacon. This does now require sessionStorage to be available for beacon to work, however we will gracefully fallback to Post if sessionStorage isn't available, this seems like a reasonable sacrifice to improve beacon events from Safari.

Having to send the preflight via POST is an unfortunate consequence of Safari's Beacon implementation that doesn't send preflight requests, where we potentially lose events if the beacon event is sent to a url that the browser has never visited before.
This won't fix beacon events on Chromium based browsers but they successfully fallback to POST until the Chromium team fix sending Blobs via Beacon API.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
-Version 2.13.0 (2019-11-21)
+Version 2.13.0 (2019-01-16)
 ---------------------------
 Add activity tracking callback mechanism (#774)
+Update beacon support to handle "gotchas" (#716)
 
 Version 2.12.0 (2019-10-31)
 ---------------------------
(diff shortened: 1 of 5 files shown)


ScottyMJacobson:
Interesting - my only comment so far would be "do we really want to read and then write from session storage on _every_ request?" i know we already have a loooot of local variables in our  `out_queue.js` scope but i think it'd be helpful to cache it somehow

the other comment i'm going to write in the issue thread

paulboocock:
Switching the release/2.13.0 branch. Accidently closed this, see #786 for new PR.

PR #786, opened 2020-01-23
Issue/716 update beacon support

paulboocock:
#716 #751 Now fires the first event with POST to ensure any CORS preflight requests have been sent by the browser before we start using Beacon. This does now require sessionStorage to be available for beacon to work, however we will gracefully fallback to Post if sessionStorage isn't available, this seems like a reasonable sacrifice to improve beacon events from Safari.

Having to send the preflight via POST is an unfortunate consequence of Safari's Beacon implementation that doesn't send preflight requests, where we potentially lose events if the beacon event is sent to a url that the browser has never visited before.
This won't fix beacon events on Chromium based browsers but they successfully fallback to POST until the Chromium team fix sending Blobs via Beacon API.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
-Version 2.13.0 (2019-11-21)
+Version 2.13.0 (2019-01-16)
 ---------------------------
 Add activity tracking callback mechanism (#774)
+Update beacon support to handle "gotchas" (#716)
 
 Version 2.12.0 (2019-10-31)
 ---------------------------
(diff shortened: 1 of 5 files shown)


paulboocock:
@ScottyMJacobson I've taken your feedback from #781 (which accidently closed when I switched from develop to release branch). We're now only writing to sessionStorage when required (so only when Beacon is enabled) and only reading if we've not only already read the value once. As I mentioned on the other ticket, we've got some more testing to do with this but initial tests show that this is a step in the right direct for Beacon at the moment.

Issue #772, opened 2019-10-30
Fix Optimizely contexts not collecting for OptimizelyX


paulboocock:
**Affected Version:** `2.11.0`

**Description:** When using the latest OptimizelyX script, the `data` object on the optimizely script has changed so we no longer collect a number of contexts. We need to pull the data from window.optimizely.get('data') rather than window.optimizely.data as that has been deprecated.

```
window.optimizely.data
Object { note: "Obsolete, use optimizely.get('data') instead" }
```

This should allow the following contexts to start collecting again:
      optimizelySummary,
      optimizelyExperiments,
      optimizelyStates,
      optimizelyVariations,
      optimizelyVisitor,
      optimizelyAudiences,
      optimizelyDimensions

Along with the fix in #763 which fixes optimizelyXSummary we should have full tracking of optimizely contexts again.

paulboocock:
This only occurs if you use OptimizelyX and try to enable the Optimizely autocontexts. That isn't an actual use case that we see and the new OptimizelyXSummary captures much of the useful information in these additional contexts (such as experimentId, variation and visitorId).

Issue #765, opened 2019-10-24
Add activity tracking callback mechanism


jamessnowplow:
It would be great if we had some mechanism to hook in to the current page ping setup and instead of sending a ping on every tick call a user supplied callback.

This would allow people to implement their own activity ping sending strategies. For instance aggregating and sending on unload/visibilitychange, using custom schemas etc.

Issue #750, opened 2019-10-02
Reset activity on page view


asgergb:
If you enable activity tracking in a single page app, activity/heartbeats spans (virtual) page views. In short, we are interested in the time spent on individual pages and not total visit length. Therefore, we would like the user's activity to be reset when we trigger `trackPageView`.

For example, if you set the heartbeat delay to be 30 seconds, and the user spends 15 seconds on page 1 and 15 seconds on page 2, the 30 seconds will be attributed to page 2.

Sample implementation with a config variable to enable the reset: https://github.com/Jyllands-Posten/snowplow-javascript-tracker/commit/0b9e49398d9cb7c5e0052473cd31585043db7365

This also happens to solve a related problem with single page apps where the contexts in the new page view for page 1 are not included in the heartbeats on page 2 because the heartbeats inherit page 1's contexts. In general, you can use the context callback to solve this, however, when using Google Tag Manager, the dataLayer variables are frozen when the first page view is triggered (https://www.simoahava.com/analytics/fetch-latest-value-data-layer-variable/).

yalisassoon:
@asgergb - many thanks for opening this issue.

I _think_ there are a couple of issues on single page apps with our current approach:

1. When you trigger a manual `trackPageView` call to indicate that a virtual page view is fired, the activity tracking does not reset. This means that if the e.g. the page pings are set to fire every 10 seconds, but only after an initial waiting period of 30 seconds, there isn't a 30 second interval after the virtual page view before the first page ping on the new virtual page view fires. (Instead the previous schedule of page pings is maintained based on the last "real" (rather than virtual) page view.
2. The global contexts that are sent with each page view are those inherited from the last "real" page view rather than the last virtual page view, which should be the desired behaviour.

If that's right then the most likely way we'd solve this is to add one or two arguments to the trackPageView methods that would be called from an SPA on firing a virtual page view that:

1. Reset activity tracking
2. Reset the all global contexts

Does that make sense? Want to make sure there isn't an aspect of running Snowplow on a SPA that we're missing?

asgergb:
1. Yes, the schedule depends entirely on the first page view.
2. If by "global" contexts you mean the contexts send to `trackPageView`, yes, the additional page views inherit the first page view's contexts.

[...]

PR #787, opened 2020-01-23
Release/2.13.0

paulboocock:


src/js/tracker.js
```diff
@@ -1954,21 +1946,15 @@
...
 		 * Enable/disable user fingerprinting. User fingerprinting is enabled by default.
-		 * @param bool enable If false, turn off user fingerprinting
 		 */
-		apiMethods.enableUserFingerprint = function(enable) {
-			helpers.warn('enableUserFingerprintSeed is deprecated. Instead add a "userFingerprint" field to the argmap argument of newTracker.');
-			if (!enable) {
-				userFingerprint = '';
-			}
+		apiMethods.enableUserFingerprint = function() {
+			helpers.warn('User Fingerprinting is no longer supported. This function will be removed in a future release.');
```
istreeter: This string 'User Fingerprinting....' appears three times in the minified bundle.  You could minify further if the string was a constant, which you reused on lines 1834, 1950, 1957.

src/js/tracker.js
```diff
@@ -2052,9 +2038,33 @@
...
-				configMinimumVisitTime = new Date().getTime() + minimumVisitLength * 1000;
-				configHeartBeatTimer = heartBeatDelay * 1000;
+				activityTrackingConfig = {
+					activityTrackingEnabled: true,
+					configMinimumVisitLength: minimumVisitLength,
+					configHeartBeatTimer: heartBeatDelay * 1000
+				}
+			} else {
+				helpers.warn("Activity tracking not enabled, please provide integer values " +
+					"for minimumVisitLength and heartBeatDelay.")
```
istreeter: And this string appears twice.  It depends how much you care about minification, of course!

src/js/lib/helpers.js
```diff
@@ -348,6 +348,37 @@
 		}
 	};
 
+		/**
+	 * Attempt to get a value from localStorage
+	 *
+	 * @param string key
+	 * @return string The value obtained from localStorage, or
+	 *                undefined if localStorage is inaccessible
```
istreeter: This documentation says it returns undefined, but actually you return null.  Does it matter?

src/js/tracker.js
```diff
@@ -1604,25 +1562,60 @@
...
+					pagePingCallbackInterval = activityInterval({
+						...activityTrackingCallbackConfig,
+						configMinimumVisitTime: lastActivityTime + activityTrackingCallbackConfig.configMinimumVisitLength * 1000
+					});
+				}
 			}
 		}
 
+		function activityInterval({ activityTrackingEnabled, configHeartBeatTimer, configMinimumVisitTime, callback }) {
+			if (!activityTrackingEnabled) return;
```
istreeter: So here you `return` nothing, which basically sets pagePingInterval to `undefined`.  But elsewhere in the code you are checking `pagePingInterval === null` (not undefined).  I think you are safe for now, but this might be a null/undefined bug waiting to happen in future.

You could just drop the null equality check at line 1572, because I think it safe to do `clearTimeout(null)`.

paulboocock: Great spot!
In fact, the early return can go, we're never going to end up inside this function if `activityTrackingEnabled` is `false`. It's called in two places currently and both check for this being true before calling activityInterval.

src/js/lib/helpers.js
paulboocock: Tidied this up to be consistent with the localStorage methods.

src/js/tracker.js
paulboocock: Opened #791 to sort this out for all repeated strings

src/js/tracker.js
paulboocock: Opened #791 to sort this out for all repeated strings

src/js/tracker.js
```diff
@@ -1112,21 +1095,20 @@
...
 				var variation = state.getVariationMap()[activeExperiment];
-				var variationName = variation.name;
-				var variationId = variation.id;
-				var visitorId = visitor.visitorId;
+				var variationName = (variation && variation.name && variation.name.toString()) || null;
+				var variationId = variation && variation.id;
+				var visitorId = (visitor && visitor.visitorId && visitor.visitorId.toString()) || null;
 				return {
-					experimentId: parseInt(activeExperiment),
+					experimentId: parseInt(activeExperiment) || null,
```
istreeter: Is zero a legitmate value for activeExperiment?  Because here you would replace it with null.  Same for variationId.

src/js/tracker.js
```diff
@@ -1112,21 +1095,20 @@
 		function getOptimizelyXSummary() {
 			var state = getOptimizelyXData('state');
 			var experiment_ids = state.getActiveExperimentIds();
-			var experiments = getOptimizelyXData('data', 'experiments');
 			var visitor = getOptimizelyXData('visitor');
 
 			return map(experiment_ids, function(activeExperiment) {
 				var variation = state.getVariationMap()[activeExperiment];
```
istreeter: Maybe you could pull `state.getVariationMap()` up outside of the loop, so you don't have to re-call it for each activeExperiment in the array.

src/js/tracker.js
```diff
@@ -1112,21 +1095,20 @@
...
+				var variationName = (variation && variation.name && variation.name.toString()) || null;
+				var variationId = variation && variation.id;
+				var visitorId = (visitor && visitor.visitorId && visitor.visitorId.toString()) || null;
 				return {
-					experimentId: parseInt(activeExperiment),
+					experimentId: parseInt(activeExperiment) || null,
 					variationName: variationName,
-					variation: parseInt(variationId),
+					variation: parseInt(variationId) || null,
 					visitorId: visitorId
```
istreeter: I suppose you could also check the type of visitorId.  It depends how paranoid you want to be.

istreeter:
We have 1 context that supports the new optimizelyX: optimizelyXSummary

And we have 7 contexts that only work on the old deprecated optimizely classic:
*  optimizelyExperiments: true
*  optimizelyStates: true
*  optimizelyVariations: true
*  optimizelyVisitor: true
*  optimizelyAudiences: true
*  optimizelyDimensions: true
*  optimizelySummary: true

Should we look to support those 7 contexts on optimizelyX?  It would be  nice if the tracker looked at the `window.optimizely` object and worked out what version it is on.  Btw, I could not test any of the optimizely classic options.

One other small thing: you could add `bundle-postbabel.js` to the `dist/.gitignore` file

src/js/tracker.js
```diff
@@ -2052,12 +2034,34 @@
...
+				helpers.warn(activityTrackingWarning)
+			}
+		};
+
+		/**
+		 * Enables page activity tracking (replaces collector ping with callback).
+		 *
+		 * @param int minimumVisitLength Seconds to wait before sending first page ping
+		 * @param int heartBeatDelay Seconds to wait between pings
+		 * @param function callback function called with ping data
```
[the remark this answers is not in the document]
paulboocock: This is a good idea, particularly as it allows the page ping to be tied back to the page view which started the activity tracking.
I've passed this back on the object into the callback now.
i.e. 

```
            window.snowplow('enableActivityTrackingCallback', 10, 10, function (a) {
                console.log(a.context);
            }
```

As it is already an array, it can be passed in as the last parameter onto any 'track' calls that may be done in the activity tracking callback. I like the extra flexibility that this offers users.

src/js/tracker.js
```diff
@@ -1604,25 +1562,52 @@
 				helpers.addEventListener(windowAlias, 'resize', activityHandler);
 				helpers.addEventListener(windowAlias, 'focus', activityHandler);
 				helpers.addEventListener(windowAlias, 'blur', activityHandler);
+			}
 
+			if (resetActivityTrackingOnPageView || ((activityTrackingConfig.activityTrackingEnabled || activityTrackingCallbackConfig.activityTrackingEnabled) && installingActivityTracking)) {
```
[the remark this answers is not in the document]
paulboocock: Can't believe I missed this!

src/js/tracker.js
paulboocock: I don't believe so. In all my testing Experiements seem to be around 9 or 10 digit numbers.

src/js/tracker.js
paulboocock: I'm toString'ing or nulling this just above. I think that should be enough to ensure we end up with a string.

paulboocock:
> We have 1 context that supports the new optimizelyX: optimizelyXSummary
> 
> And we have 7 contexts that only work on the old deprecated optimizely classic:
> 
>     * optimizelyExperiments: true
> 
>     * optimizelyStates: true
> 
>     * optimizelyVariations: true
> 
>     * optimizelyVisitor: true
> 
>     * optimizelyAudiences: true
> 
>     * optimizelyDimensions: true
> 
>     * optimizelySummary: true
> 
> 
> Should we look to support those 7 contexts on optimizelyX? It would be nice if the tracker looked at the `window.optimizely` object and worked out what version it is on. Btw, I could not test any of the optimizely classic options.
> 
> One other small thing: you could add `bundle-postbabel.js` to the `dist/.gitignore` file

We discussed this a while back, these contexts are not particularly useful and the OptimizelyXSummary actually captures much of the useful data that is in these 7 extra contexts but does so in the single Summary context: https://github.com/snowplow/snowplow-javascript-tracker/issues/772

CHANGELOG
```diff
@@ -1,3 +1,15 @@
+Version 2.13.0 (2019-01-27)
+---------------------------
+Add activity tracking callback mechanism (#774)
+Reset activity on page view (#750)
```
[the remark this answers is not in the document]
paulboocock: Through the combination of changes in #750 and #765 we should now be reseting the page view contexts for future page pings when there is a second page view.

src/js/out_queue.js
```diff
@@ -286,6 +287,9 @@
 				xhr.onreadystatechange = function () {
 					if (xhr.readyState === 4 && xhr.status >= 200 && xhr.status < 400) {
 						clearTimeout(xhrTimeout);
+						if (useBeacon && !beaconPreflight) {
+							helpers.attemptWriteSessionStorage('sp_beaconPreflight', true);
```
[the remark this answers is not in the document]
paulboocock: Great spot. I've followed the pattern for queueName and used the functionName and namespace for this. These two variables should ensure this key exists for each collector URL that the tracker uses.

src/js/tracker.js
```diff
@@ -2052,12 +2034,34 @@
...
+		};
+
+		/**
+		 * Enables page activity tracking (replaces collector ping with callback).
+		 *
+		 * @param int minimumVisitLength Seconds to wait before sending first page ping
+		 * @param int heartBeatDelay Seconds to wait between pings
+		 * @param function callback function called with ping data
+		 */
+		apiMethods.enableActivityTrackingCallback = function (minimumVisitLength, heartBeatDelay, callback) {
```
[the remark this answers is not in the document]
paulboocock: I've removing the duplication in these two functions now, whilst maintaining the two seperate APIs. I've done this so both can be used at the same time, allowing for easier transition or if someone just fancies using both. I think the code is a little more complicated now but there is certainly less duplication.

src/js/tracker.js
```diff
@@ -1604,25 +1563,70 @@
...
-
-					// There was activity during the heart beat period;
-					// on average, this is going to overstate the visitDuration by configHeartBeatTimer/2
-					if ((lastActivityTime + configHeartBeatTimer) > now.getTime()) {
-						// Send ping if minimum visit time has elapsed
-						if (configMinimumVisitTime < now.getTime()) {
-							logPagePing(finalizeContexts(context, contextCallback)); // Grab the min/max globals
-						}
+
+				for (var key in activityTrackingConfig.configurations) {
```
istreeter: You could use `let` instead of `var` to keep it scoped to the block.

src/js/tracker.js
```diff
@@ -1604,25 +1563,70 @@
...
+		 * @param heartBeatDelay string The length between checks to see if we should send a page ping
+		 * @param callback function A callback function to execute
+		 */
+		function configureActivityTracking(minimumVisitLength, heartBeatDelay, callback) {
+			if (minimumVisitLength === parseInt(minimumVisitLength, 10) &&
+				heartBeatDelay === parseInt(heartBeatDelay, 10)) {
+				return {
+					activityTrackingEnabled: true,
+					configMinimumVisitLength: minimumVisitLength,
+					configHeartBeatTimer: heartBeatDelay * 1000,
```
istreeter: I haven't tested this, but it looks like you are using the string version of heartBeatDelay instead of the numeric version you got by `parseInt`

Also, I think `activityTrackingEnabled` never gets used.  You only ever use the `enabled` property on the parent object.

src/js/tracker.js
```diff
@@ -348,7 +334,14 @@
...
 
 			// Whether first trackPageView was fired and pageViewId should not be changed anymore until reload
-			pageViewSent = false;
+			pageViewSent = false,
+
+			// Activity tracking config for callback and pacge ping variants
+			activityTrackingConfig = {
+				enabled: false,
+				installed: false, // Guard against installing the activity tracker more than once per Tracker instance
+				configurations: {}
```
istreeter: I like this 👍

But.... I think you never use the object key for anything.  Therefore, why not store it as a plain array instead of a key-val dictionary?

This has a nice side effect that the client can call `enableActivityTrackingCallback` as many times as they want with different callback functions.

You might prefer to drop `enabled` and instead do a quick test for if the array is empty.

paulboocock: I did this first (using an array) but enableActivityCallback has always worked in such a way that you can call it multiple times and it adjusts the previous version. I feared adding multiple versions would either lead to bugs, cause some sites to send duplicated page pings or make the code more complicated by only allowing 1 page ping version but multiple callback versions. 
My thinking ultimately ended up as allowing a single version of each was safer and more consistent.

paulboocock: Oh, forgot to talk about .configurations. 
I made this seperate so when I iterate over the keys, I don't have to check if I'm looking at a callback configuration or one of the activityTrackingConfig booleans. Maybe there is another way of tidying this up but without multiple objects I'm not sure what would be neater.

src/js/tracker.js
paulboocock: Agreed. However, the whole codebase is for(var in ...). Leaving as is for consistency but we should look at updating these, particularly for v3.

src/js/tracker.js
paulboocock: My comment in nonsense on this function. heartBeatDelay should always be an `int` (see apiMethods that call this for clarification). I've updated the comment.

PR #796, opened 2020-02-25
Fix SameSite cookie warning for storage Cookies (close #795)

### LINKED ISSUE #795: Fix SameSite cookie warning for storage Cookies
The Storage cookies don't currently set the SameSite attribute as it doesn't make any difference for how the JS Tracker reads/writes cookies. Chrome however, will produce a developer console warning due to this. We should set the SameSite attribute to remove the warning in Chrome.

paulboocock:
This feature now sets the two cookies which are set by the JavaScript tracker to `SameSite=None; Secure` by default. This has no impact on the behaviour of the cookie from the point of view of the JavaScript tracker, it simply reads and writes this cookie as a storage target to be added to the JSON payload - it does not need to be sent to the collector as a header.

It does however, remove the warning in Chrome when these cookies are included in requests to collectors (due to the cookie being set on the same domain as the collector).

Also, the default needs to be `SameSite=None; Secure` rather than `SameSite=Lax` so that the tracker continues to work in iframes with the default settings. The only change to the defaults here, are that cookies will only work on HTTPS now. If that is a problem for users then they can set `cookieSecure: false` on tracker initialisation. 

Old behaviour (with Chrome warning and not working in third party iframes) can be achieved with `cookieSameSite: null` and `cookieSecure: false`.

Users can also choose to go with `cookieSameSite` parameter of `None` (default), `Lax` or `Strict` if they wish, I'll ensure to document this and mention it in the upgrading section of the blog post.

### DIFF
--- a/src/js/lib/detectors.js
+++ b/src/js/lib/detectors.js
@@ -38,7 +38,7 @@
 		isFunction = require('lodash/isFunction'),
 		isUndefined = require('lodash/isUndefined'),
 		tz = require('jstimezonedetect').jstz.determine(),
-		cookie = require('browser-cookie-lite'),
+		helpers = require('./helpers'),
 
 		object = typeof exports !== 'undefined' ? exports : this, // For eventual node.js environment support
 		
@@ -99,8 +99,8 @@
 		var cookieName = testCookieName || 'testcookie';
 
 		if (isUndefined(navigatorAlias.cookieEnabled)) {
-			cookie.cookie(cookieName, '1');
-			return cookie.cookie(cookieName) === '1' ? '1' : '0';
+			helpers.cookie(cookieName, '1');
+			return helpers.cookie(cookieName) === '1' ? '1' : '0';
 		}
 
 		return navigatorAlias.cookieEnabled ? '1' : '0';
--- a/src/js/tracker.js
+++ b/src/js/tracker.js
@@ -39,7 +39,6 @@
 		map = require('lodash/map'),
 		helpers = require('./lib/helpers'),
 		proxies = require('./lib/proxies'),
-		cookie = require('browser-cookie-lite'),
 		detectors = require('./lib/detectors'),
 		sha1 = require('sha1'),
 		links = require('./links'),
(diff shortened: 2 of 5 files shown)


src/js/lib/detectors.js
```diff
@@ -99,8 +99,8 @@
 		var cookieName = testCookieName || 'testcookie';
 
 		if (isUndefined(navigatorAlias.cookieEnabled)) {
-			cookie.cookie(cookieName, '1');
-			return cookie.cookie(cookieName) === '1' ? '1' : '0';
+			helpers.cookie(cookieName, '1');
```
istreeter: Are we in danger of producing a warning here, because we do not set samesite on the test cookie?

(I have not yet got as far as testing this for myself - but I will do)

src/js/tracker.js
```diff
@@ -202,6 +201,12 @@
 			// Default is user agent defined.
 			configCookiePath = '/',
 
+			// First-party cookie samesite attribute
+			configCookieSameSite = argmap.hasOwnProperty('cookieSameSite') ? argmap.cookieSameSite : 'None',
+
+			// First-party cookie secure attribute
+			configCookieSecure = argmap.hasOwnProperty('cookieSecure') ? argmap.cookieSecure : true,
```
istreeter: Instead of creating a new config setting here, could you test if `window.location.protocol == 'http'`?

src/js/lib/detectors.js
istreeter: I have now tested this for myself.  The test cookie is only set on very old (ancient!) browsers, for which `navigator.cookieEnabled` is undefined.  So there will be no issue with samesite warning messages on those old browsers.  So there is no issue here to worry about.

src/js/tracker.js
paulboocock: I could but a user might wish to set `Lax` and not Secure cookies even on HTTPS sites, so that the same cookie is accessible on HTTP too. 
I do quite like the idea of using `window.location.protocol == 'https'` as the default option instead of just `true`. My worry is with Chromes update SameSite=None and not secure are an incompatible set of default values (although SameSite=None cookies don't work on http at all with the latest changes as they have to be Secure). So with that in mind, these cookies are just not going to work well on HTTP sites anymore so I've come to the conclusion that extra code to manage http sites just doesn't add much value. If users want to use the tracker on http then they need to toggle this to `false` and `Lax`.

istreeter: Your answer is very reasonable.  My suggestion was trying to protect http users from having to make a config change.  But actually it makes sense that http users should explicitly control each parameter.

PR #802, opened 2020-03-11
Upgrade typescript to 3.7.3

HawaiianSpork:
Also removed what appears to be unneeded dependencies.


paulboocock:
Hi @HawaiianSpork thanks for your PR.
We're currently planning our next major release which will involve bringing everything up to date, including bringing TypeScript into the main browser tracker codebase and switching to rollupjs for builds. We'll likely not look at merging this until we're ready to move on v3 and the rest of our major upgrades.

PR #831, opened 2020-08-23
Upgrade UUID to ^3.3.3

leipert:
This matches the version used in the snowplow-javascript-tracker. This
will help utilizing just _one_ version of uuid and should reduce build
size of the resulting javascript.

[0]: https://github.com/snowplow/snowplow-javascript-tracker/blob/8b846ba/package.json

### DIFF
--- a/core/npm-shrinkwrap.json
+++ b/core/npm-shrinkwrap.json
@@ -13,15 +13,13 @@
     "@types/node": {
       "version": "9.6.55",
       "resolved": "https://registry.npmjs.org/@types/node/-/node-9.6.55.tgz",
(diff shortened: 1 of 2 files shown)


paulboocock:
Hi @leipert 
Thanks for this, a great spot! I'll merge this into the next Core 0.9.0 release however it won't make it immediately into a JS Tracker release but it won't be too far behind.

Our current plan is to upgrade "core" (with this change and new build tooling - rollup), upgrade the "NodeJS Tracker" with the improved Core as it hasn't been bumped in a while, then some build tooling changes (rollup again) and this core bump in a new js 2.16.0 release.

paulboocock (approved):

leipert:
Thanks @paulboocock. No worries, looked into building the tracker ourselves and just saw this.

Maybe it is possible to upgrade even further, but I think there is different browser support in uuid 7 and up.

PR #834, opened 2020-08-26
release/core/0.9.0

paulboocock:


core/src/contexts.ts
```diff
@@ -101,91 +112,91 @@ export function validateVendorParts(parts: Array<string>): boolean {
...
+export function isEventJson(input: unknown): input is PayloadDictionary {
+    const payload = input as PayloadDictionary;
+    if (isNonEmptyJson(payload) && ('e' in payload))
+        return (typeof(payload.e) === 'string');
     return false;
 }
 
-export function isRuleSet(input: any) : boolean {
+export function isRuleSet(input: unknown): input is Record<string, unknown> {
+    const ruleSet = input as Record<string, unknown>;
```
istreeter, line 166: This looks like one of the few places where the code is changed, rather than just superficial changes.

Previously there was a `if (isPlainObject(input))` whereas here you cast it to a Record.  Please just confirm you are satisfied this is always a plain object?

core/src/payload.ts
```diff
@@ -57,12 +61,12 @@ export function isNonEmptyJson(property): boolean {
 /**
  * Is property a JSON?
  */
-export function isJson(property: Object): boolean {
-	return (typeof property !== 'undefined' && property !== null &&
-	(property.constructor === {}.constructor || property.constructor === [].constructor));
+export function isJson(property: unknown): boolean {
+	const record = property as Record<string, unknown>;
```
istreeter, line 65: Should this be `property as Record<string, unknown> | undefined | null`?  Based on the checks you still on line 66.

And is there a reason to reassign `record = property` instead of changing the signature of the function to accept a `Record<string, unknown>`?

Both genuine questions, not complaints!

core/src/contexts.ts
paulboocock: Ah yeah, I think I removed this in the reshuffle. I don't think we can be sure this is always a plan object as the input comes from the user of the tracker (typically). I'll add this back in.

core/src/payload.ts
paulboocock: I can't change the signature to `Record<string, unknown>` as this function should be allowed to be passed anything and checks if it is valid json (and there is a weird test that passes it a function). 
As the tests are in TypeScript, making it `Record<string, unknown>` means the test doesn't compile so I think to be true to its real usecase then it should stay as `unknown`. 
It's a bit of an odd function for a typescript library in reality, pretty much exists to sanitise nonsense from any JavaScript consumers.
As for `property as Record<string, unknown> | undefined | null`, yeah I think that's a more explicit approach for sure 👍

paulboocock:
For some extra confidence and test runs:
Also a build of the current JS Tracker with 0.9.0-beta.2 https://travis-ci.org/github/snowplow/snowplow-javascript-tracker/builds/721619173
And a build of the new NodeJS Tracker release with 0.9.0-beta.2 https://travis-ci.org/github/snowplow/snowplow-nodejs-tracker/builds/721628664

istreeter (approved):
Lots of good work here, making it more strictly typed without changing the functionality.

The introduction of classes is setting the likely future of this code base to be more object-oriented.

paulboocock:
Based on your comments and our earlier discussion I've switched back to the earlier style and removed the classes. 
This change removes any breaking changes, is more functional and ends up 3kb smaller when minified in the JS Tracker.

PR #843, opened 2020-09-23
Release 2.16.0

paulboocock:
This is a challenging PR to look at the diff due to the removal of the IIFEs around every file, apologies!


paulboocock:
Switched this to be the release PR as the other feature planned for 2.16.0 has been pushed back to v3.

README.md
```diff
@@ -42,15 +41,17 @@ host$ docker build -t core .
 host$ docker run -v "$(pwd)":"/code" core npm run build
 ```
 
+Build the core library using `npm run build` and run unit tests with `npm run test`.
```
colmsnowplow, line 44: nit - should it say `cd ./core` first?

package.json
```diff
@@ -1,42 +1,51 @@
...
+    "test:unit": "jest tests/unit/*.spec.js",
+    "test:e2e:sauce": "gulp testEndToEnd && wdio tests/wdio.sauce.conf.js",
+    "test:e2e:local": "gulp testEndToEnd && wdio tests/wdio.local.conf.js"
+  },
   "dependencies": {
     "jstimezonedetect": "1.0.5",
+    "lodash": "^4.17.20",
     "sha1": "git://github.com/pvorb/node-sha1.git#910081c83f3661507d9d89e66e3f38d8b59d5559",
-    "snowplow-tracker-core": "^0.8.0",
+    "snowplow-tracker-core": "file:./core",
```
colmsnowplow, line 15: For the prod build & deploy, do we still point to the core directory now rather than the old way of deploying core first then installing it from npm?

colmsnowplow (approved):
LGTM! New build & test process via npm is lovely. :)

README.md
paulboocock: Not for the docker image, as the instructions have you mounting to `/core` folder. This might be misleading for those not using the docker image though, I'll add a little clarification.

package.json
paulboocock: I've gone back and forth on this a lot. The issue with pointing at a released `core` is that it's really hard to run the e2e tests on saucelabs with any updates to core. It generally makes the whole development process a little fiddly.

I went with this option as ultimately it seems safer too - if someone did make changes to core and didn't push a core release (which would be bad, but possible) then the tracker would still have the latest core code. We should always ensure we do core releases when we update it but this technique means if we ever made a mistake and forgot then we don't bork the JS Tracker.

PR #860, opened 2020-11-23
Cleanup deprecated methods (close #557)

### LINKED ISSUE #557: Cleanup deprecated methods
There's quite a few methods which were deprecated long time ago, such as [this](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/src/js/in_queue.js#L88) or all setter methods of tracker. We can wait until 3.0.0, but they print deprecation warning, so I don't think anyone really use them and it's safe to remove.

paulboocock:
I've ripped out the "old" stuff here. There is still considerable change planned for v3 beyond this but thought this one was worth a PR to check I haven't removed anything that seems particularly risky?


core/src/core.ts
```diff
@@ -184,24 +184,7 @@ export interface Core {
...
-  trackUnstructEvent: (
-    properties: Record<string, unknown>,
-    context?: Array<SelfDescribingJson>,
-    tstamp?: Timestamp,
-    afterTrack?: (Payload: PayloadDictionary) => void
-  ) => PayloadData;
-
-  /**
-   * Log an self-describing (previously unstruct) event
+   * Log an self-describing event
```
colmsnowplow, line 187: Nit: `a self-descibing`

examples/web/async-large.html
```diff
@@ -85,9 +80,9 @@
         };p[i].q=p[i].q||[];n=l.createElement(o);g=l.getElementsByTagName(o)[0];n.async=1;
         n.src=w;g.parentNode.insertBefore(n,g)}}(window,document,"script","../sp.js","snowplow_2"));
 
-        window.snowplow_2('newTracker', 'cf', '<<MY_COLLECTOR_URL>>', { // Initialise a tracker
+        window.snowplow_2('newTracker', 'sp', '<<MY_COLLECTOR_URL>>', { // Initialise a tracker
```
colmsnowplow, line 83: I think tracker namespace needs to be different. `sp2` perhaps? Maybe also change the collector var to illustrate going to two separate places.

src/js/lib/proxies.js
```diff
@@ -34,33 +34,6 @@
 
 import { fromQuerystring, getHostName } from './helpers';
 
-/*
```
colmsnowplow, line 37: Do you know what this is/why it was created in the first place?

Not that I see any reason to object to removing it, just hard to think about without knowing what it is. :)

src/js/tracker.js
```diff
@@ -1780,67 +1745,6 @@ export function Tracker(functionName, namespace, version, mutSnowplowState, argm
     );
   }
 
-  /**
```
colmsnowplow, line 1783: Similar thing here - any context on what this code is for in the first place?

colmsnowplow (approved):
There are two things I commented on which I just don't really understand what the code was for in the first place. Assuming they're ok, looks good!

src/js/lib/proxies.js
paulboocock: This was how Yahoo used to show cached pages, they just use Bing now though.

src/js/tracker.js
paulboocock: I had to do a bit of Googling on this one to see what it was all about. 

Once upon a time, there was a visility state called `prerendered` that a page could be in, and it wouldn't be visible to a user. We used to allow prerendered pages to be ignored or included which is what this code used to do based on `configCountPreRendered` and we'd only trigger the page view (or whatever event) once the page transitioned into another visibility state (see the event listenter in this section).

However, this has been deprecated for a while and has poor cross browser support. From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState):
> prerender
> The page content is being prerendered and is not visible to the user (considered hidden for purposes of document.hidden). The document may start in this state, but will never transition to it from another value. Note: This was removed from the standard. Check compatibility table for details.

However, you pointing this out has made me realise I missed the config variable!

istreeter (approved):
I've looked through it, but I don't have anything helpful to add.  It all looks in order.

PR #871, opened 2020-12-01
Improve test coverage

paulboocock:
This PR covers test coverage improvements, mainly in the form of new and updates integration tests. 

Also took the opportunity to bump dev dependencies with the increased confidence the new tests give, and upgraded webdriverio to v6 (which had a few breaking changes so calling it out here, but nothing major e.g. `browser.waitUntil`).

I've also removed the E2E test run when deploying a release, as it is a different workflow it ends up running in parallel to the main build and collides with saucelabs (as we only have 1 tunnel). We'd never merge to master and release without a successful run on a PR so I see no risk in this - left unit tests in for a sanity check as they run quickly.

Will be merged into the `feature/anonymous_tracking_header` branch first and then into `release/2.17.0`.


tests/integration/autoTracking.spec.js
```diff
@@ -0,0 +1,770 @@
...
+    // time for activity to register and request to arrive
+    browser.pause(5000)
+    browser.call(() =>
+      fetchResults(docker.url).then(result => {
+        log = result
+      })
+    )
+  })
+
+  it('should send a link click event', () => {
```
colmsnowplow, line 112: Perhaps we should also check that the `page_url` field is as expected on the link click event?

tests/integration/autoTracking.spec.js
```diff
@@ -0,0 +1,770 @@
...
+                formId: 'excludedForm',
+                formClasses: [ 'excluded-form' ]
+              }
+            } 
+          }
+        }
+      })
+    ).toBe(false)
+  })
+})
```
colmsnowplow, line 770: Love this. Something else that strikes me is that anywhere we have multiple events on a page, maybe we should enable the web page context and check that it behaves correctly. (eg just a count of distinct values matching the amount of page loads we expect).

Might save us effort in thinking of a way to check this directly if we ever need to.

tests/pages/integration.html
```diff
@@ -32,15 +32,24 @@
 
   document.write(collector_endpoint)
 
+  var eventMethod = parseQuery().eventMethod;
+
   window.snowplow('newTracker', 'cf', collector_endpoint, {
```
colmsnowplow, line 37: nit - cf is short for cloudfront. Makes no odds really but maybe a different name is preferable?

colmsnowplow (approved):
LGTM!

tests/pages/integration.html
paulboocock: I've removed all the `cf` references in v3.0.0 when I deprecated all the "old" features. I'll ensure these new tests have any `cf` references removed in v3 too, I left them in this release for consistency with the other tests 🤷

PR #884, opened 2021-01-08
Don't overwrite custom user defined event id

mcepok:
In some cases, we already set an event id provided via `payloadPairs` and it's important for us to keep that event id. However, currently in the core tracker the event id is overwritten no matter what. One very simple solution to this issue is to switch the order of adding the event id and the `payloadPairs`. In the normal way of using the tracker without providing an own event id, nothing changes.

### DIFF
--- a/core/src/core.ts
+++ b/core/src/core.ts
@@ -782,8 +782,8 @@ export function trackerCore(base64: boolean, callback?: (PayloadData: PayloadDat
     tstamp?: Timestamp,
     afterTrack?: (Payload: PayloadDictionary) => void
   ): PayloadData => {
-    sb.addDict(payloadPairs);
     sb.add('eid', v4());
+    sb.addDict(payloadPairs);
     const timestamp = getTimestamp(tstamp);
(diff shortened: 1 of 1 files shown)


paulboocock:
Thanks for the PR @mcepok 

We're well on the way to v3 now which sees a significant reshuffle of the codebase. I'm happy to include this change in v3. 

If you want to build your own version of `sp.js` whilst you wait for v3, then you should be able to run: `npm run build:all` and it'll build core and the tracker then output a new `sp.js` to `/dist`.

mcepok:
We actually want this behavior for the snowplow-nodejs-tracker which, as far as I understood, is based since version 0.4.0 on the core-tracker of this repository. If the next release of the core-tracker is coupled with the javascript-tracker release, that's also fine.

Thanks a lot!

paulboocock (approved):

paulboocock:
I've cherry-picked this into the release/3.0.0 branch and opened corresponding issue #911

PR #888, opened 2021-02-03
Release/2.17.1

paulboocock:
Removed the use of cookies which don't have any samesite attributes. Also removed the physical cookie check in `hasCookies` for the detectors as `navigator.cookieEnabled` has been supported since IE4 so this code never fires.

Also removed the anonymous headers on beacon requests as they have no effect. Server Side Anonymous tracking doesn't actually work when using Beacon, will add a note to the docs to make this clear.


colmsnowplow (approved):
LGTM!

Removing some small pieces of additional code as we go, nice. :)

Issue #719, opened 2019-04-16
Improve event flushing options and remove pageUnloadTimer


yalisassoon:
Current value used in combination with automatic link click tracking can slow the loading of subsequent pages on a website, as the timer is enabled after each link click.

paulboocock:
With improvements to Beacon and the local storage queue, it seems reasonable to reduce this. 
Perhaps we should have a set of defaults depending on tracker config:
Using GET and Cookies - stay at 500ms
Using POST and Local Storage - drop to 100ms(?)
Using Beacon - drop to 0ms(?)

Needs considerable testing for event loss.

alperakgun:
>  Using Beacon - drop to 0ms(?)

@paulboocock Thinking about the question mark at the end, what risks could there be using Beacon with a `0ms`  `pageUnloadTimer`?

PS: We have a Post/Localstorage based implementation, where we want to keep the `link_click` events, and have no delay before unloading a page.

paulboocock:
Hey @alperakgun I totally missed this notification, apologies!

In an ideal world, using 0ms with beacon would be the perfect scenario, and it should technically work, but beacon behaves inconsistently across browsers, which is the main cause for me to add the `(?)`. I think you'll be fine but you're trusting the browsers to send events so you might find some browsers don't behave as you would expect (particularly older Safari (10-11ish) and Chrome < v81 (falls back to POST)). I think this is a balancing act, are you happy to lose some events for improved website performance? If so, give it a try and monitor your results over a couple of weeks - I'd strongly advise looking at results grouped by browser to see if you notice any significant changes in events from certain browsers.

alperakgun:
Thanks @paulboocock - it makes sense.

paulboocock:
I've taken a multiple pronged strategy here which allows for the removal of the pageUnloadTimer.

I'm now flushing the buffers on `page vivisilibty -> hidden` events. This means we flush more frequently but works on mobile and desktop. I'm also leaving in the `beforeUnload` flush but with no delay timer (since it doesn't work in modern browsers anyway). To keep the "old browser" functionality, when flushing I'm sending the XHR with the `sync` flag as `true`. Modern browsers ignore this but old browsers likely respect it so should still see beforeUnload flushing work.

[...]

PR #894, opened 2021-02-11
#fix #890 by disabling buffering after it has been flushed on `beforeunload`

### LINKED ISSUE #890: "Late to the Party" bug during `beforeunload` browser teardown
Hello all! Good to be back in the repo - really exciting to see how actively you're maintaining it, and hope I can contribute positively again.

**Describe the bug**
In response to a window `beforeunload` event, the snowplow singleton [runs all the `bufferFlushers`](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/src/js/snowplow.js#L118). A `OutQueueManager`'s `bufferFlusher` [sends out everything it has in its buffer so far](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/src/js/out_queue.js#L129). However, depending on the order of a user's `beforeunload` handler, or if they're using an event that comes after `beforeunload` (it's not clear to me where `pagehide` happens), they might then track their relevant event (in Eko's use case, it's called `dropoff`) _after_ this flush has happened. Basically, they're too late to the party, and just get queued, and will _maybe_ get sent the next time around the next time the user navigates to a page and it gets loaded from the localStorage queue.

**To Reproduce**
*(**edit:** Set buffer > 1, and)* attach a `beforeunload` handler that runs after snowplow tracker's. `snowplow('trackStructEvent', 'dropoff')` within that handler. It will get queued rather than sent out; you'll see it in the queue when you refresh the page. Sorry I don't have my reproduction page for this anymore. Happy to make one if it helps.

**Expected behavior**
I would hope that snowplow sort of goes into "send them out!" mode and immediately attempts to send out any more events it receives to a tracker after flushing its buffer. This sort of breaks an abstraction, but if it, for instance, [sets the bufferSize to 1 once it has flushed the buffers](https://github.com/ScottyMJacobson/snowplow-javascript-tracker/commit/b566f16ec877d4a8dfedfe7f302cdb91eb74552e), sort of disabling the buffers after the fact, that will do the trick. Would be happy to take this same commit and make a pull request on top of the current codebase 🙂.

 - OS: Mac OS X
 - Browser: Chrome
 - Version 88

ScottyMJacobson:
https://github.com/snowplow/snowplow-javascript-tracker/issues/890

Disable buffering once the tracker has cleared the buffer during `beforeUnload` by forcing bufferSize to be 1 so any subsequent events get sent out immediately

### DIFF
--- a/src/js/out_queue.js
+++ b/src/js/out_queue.js
@@ -128,6 +128,9 @@ export function OutQueueManager(
       if (!executingQueue) {
         executeQueue();
       }
+
+      // We've flushed the buffers, but if bufferSize > 1; we'll just buffer any following events...
+      bufferSize = 1; // ...So force bufferSize to 1 to force any subsequent events to send out immediately and not buffer
     });
   }
 


snowplowcla:

ScottyMJacobson:
I signed it! Assuming it hasn't changed materially since 2019

paulboocock:
Signed company CLA as Interlude, which is now Eko.

paulboocock (approved):

paulboocock:
I approved this as the change is quite innocent (and elegantly simple). 

However, on further thought and testing I'm not sure about merging it in. Mainly because I'm not sure it works like you'd hope for the use case decribed in #890 - I also don't think it's a better change than simply calling `snowplow('flushBuffer')` in your `pagehide` handler (and I'm not even sure it's going to work anyway, read on) as it does have a side effect of changing the bufferSize for future events if someone cancels the beforeUnload.

When I test this in both Firefox and Chrome, neither browser will fire the event that is sent in a `pagehide` handler before navigating to the new page. This is regardless of my bufferSize, whether it be 1 or 5, and whether I'm using `beacon` or `post`. 
You'd probably expect it to work with `beacon` but `beacon` doesn't work well in beforeunload or unload so I'd be surprised if it works with `pagehide` given that is after `beforeunload`. 
From the MDN docs on `sendBeacon()`:

> The navigator.sendBeacon() method asynchronously sends a small amount of data over HTTP to a web server. It’s intended to be used in combination with the visibilitychange event (but not with the unload and beforeunload events).

Browsers typically don't allow page navigation to be delayed now - in fact the page delay in the tracker doesn't work on the majority of new browsers in my testing and is a candidate for being dropped in v3 (although I was still gonna leave the buffer flushers in as a last chance).

I'm open to more evidence that this does work though, but I simply can't see events landing in my pipeline when triggering them in `pagehide`. They only appear when I revisit the page and get sent along with the next event.

ScottyMJacobson:
Hi @paulboocock - thanks so much for the detailed testing and response

I usually wouldn't submit a PR without doing more testing myself; this one was based on the original testing I did, and leaning on the fact that this code is live in our production tracker (which is built from this fork) sending our `pagehide` / `dropoff` events successfully - let me do more diligence and get back to you! 

_(Worth noting: just did a really cursory look at the data, and 80% of our sessions that fire our `loader.ready` event report a `dropoff` event sent in this way. Not perfect, and I'm sure there's still some inconsistency across devices, but enough to make me think that our fork / handler combo is doing _something_ right. I'm going to do some more digging to try to see why we're getting different results here - the funny thing is, when I went to repro this yesterday, I couldn't even get _any_ handler to run during `pagehide` in Chrome, which I still don't have an explanation for._

Also - 

>it does have a side effect of changing the bufferSize for future events if someone cancels the beforeUnload

This is a great point, and something I hadn't thought of because I haven't had experience cancelling during `beforeunload`. I remember seeing in the docs that this is possible (although not recommended) so I definitely concede that critique.

-----

I think next steps for us are to force a `flushBuffer` call instead as you recommended - we will, I think, just hard code it for now to come after our `dropoff` event. It couples the event to the behavior, but we're already handling `dropoff` as a special case in iOS so I'm not too mad at it. The most important thing is results! 🙂

-----

I have one more theory, and this has to do with any differences I want to try to identify between our production beacon event lifecycle and the version we're testing (latest `snowplow-javascript-tracker`). I worry that the preflight OPTIONS request, which chrome now ~~correctly*~~ sends before sending a `application/json` POST request, actually "runs out the clock" on our opportunity to get messages out as the page is dying. 

*I put an asterisk next to "correctly" because I'm noticing that Chrome doesn't seem to respect the `Access-Control-Max-Age:` field - our collector seems to return `10` by default, but then sendBeacon POST requests sent immediately after (or at least within 10 seconds) still send a preflight. If it were cached correctly, you'd expect to just see the first OPTIONS, the first POST, and then only POST requests after the first one, at least for the next 10 seconds. Wondering if you've experienced this as well.

----

Thanks for helping keep me abreast of best practice. I stumbled across [this discussion](https://github.com/mdn/sprints/issues/3722) and [this study](https://volument.com/blog/sendbeacon-is-broken), which came out late 2019 or last year, and the rabbithole of does-sendBeacon-do-what-it-promises-to-do (hand it off to the browser and then not have to worry about it getting out) seems to go deeper and deeper. Any luck / exploration on y'alls end with `fetch` with `keepalive: true`?

paulboocock:
Yeah, I've been down the sendBeacon rabbit hole a few times now looking to try and improve the reliability of it. One thing that might catch you out with your testing it that currently, the tracker will send the first request no matter what using POST (even if Beacon is set) as Safari didn't handle CORS unless it'd cached the headers first. This maybe should only trigger on Safari but I'm trying not to add any `if (safari)` calls to the tracker. Everytime I have to think about the Beacon API, it makes me a little sad how inconsistently it has been implemented 😭 
I have looked at `keepalive` but the browser support is very poor at the moment (basically only Chromium based browsers as far as I can tell), so I haven't really tried to find the time to see what benefits it might add. In reality you'd hope that sendBeacon would be just as good as `keepalive`.

Interesting theory about the OPTIONS causing enough delay that the POST doesn't fire now. I also have a strong suspisicion that the browsers have all changed so that you can no longer block page navigation by looping in `beforeunload`, so our existing unload timer idea doesn't work either. Coupled with a dodgy beacon implementation across browsers, it seems very tricky to work out where to fire events late in a pages lifecycle. The only reliable place I've seen it done is in `visibilitychanged` using Beacon.

ScottyMJacobson:
>Everytime I have to think about the Beacon API, it makes me a little sad how inconsistently it has been implemented 😭

Sums up my experience exactly... This whole experience is taking me back, way back, to early 2019

Sorry for not following up here until now - how's the refactor going?

We're about to go live with a fork that includes this change; we were ultimately able to see dropoff events arriving at our collector via beacon when testing this patch manually locally, and in our automated tests in the cloud. (Full disclosure: this has all been done on my colleague's end; I haven't gotten a chance to set up a proper testing env yet, but I will in order to test 3.0 (assuming it's still relevant then))

On that note - what does your manual testing env look like? Is it something that's part of the repo? I want to make sure I can reproduce your cases where you don't see it working for the future.

We will probably still go with that `flushBuffer` solution for our long term solution, though. Keeping the fork up to date (or not up to date) has been a headache and now that beacon is supported without the `plain/text` hack it's best to stay on the official version.

paulboocock:
Thanks for the update. Once I've got the v3 beta out, I'm going to have a look at this issue and a couple of others to see if there is anything we can do whilst we're in the feedback stage for the wider beta.

As for testing, there's a few things we do. There's the selenium tests that are part of the repo, they are pretty complete since v2.17. I also have a couple of pages I use locally that is are smattering of buttons, forms, links, etc, not part of this repo at the moment but maybe it should be. I'll add it to the todo list. 
We also typically run release candidates or betas on snowplowanalytics.com for a little while if there are significant changes, always good to test on ourselves first :-)

paulboocock:
Based on the feedback here and in #890, I've made some changes in v3 to help with your use cases and to improve beacon and unload flushing. I've documented them in  #719 as it has also allowed me to remove the pageUnloadTimer.

With that in mind I'm going to close this however I'm still keen tochat if you have further feedback.

PR #970, opened 2021-05-14
Release/3.1.0

paulboocock:
Adds functionality which improves the options around anonymous tracking to add more flexibility when toggling between anonymous and non-anonymous tracking.

PR #971, opened 2021-05-14
Release/2.18.0

paulboocock:
Same as #970 but backported into v2. Also includes the fix present in 3.0.3.

I don't intend to merge this to master, only to tag the `Prepare for 2.18.0 commit` with 2.18.0 which should be enough to create the 2.18.0 release.


adatzer (approved):
LGTM!

PR #973, opened 2021-06-01
check for sandbox SecurityErrors so snowplow clients can run in sandboxed environments

michahell:
My use-case is that the client I work for is using a low-code rapid app development solution called [Retool](https://retool.com/).
I have been building support apps with this to get around long-lasting tech debt and legacy issues, and, in all honesty:
it's pretty good! (even though it's not a long-term solution to any of those problems)
We want to track app usage and other things, however, Retool doesn't offer a tracking solution to our liking.
So we started implementing Snowplow using the javascript tracker.
This is possible, however, since Retool-built apps operate within a 'sandbox' mode (I have yet to figure out how this works - I expect CSP headers), session- and local Storage checks fail, due to javascript not being allowed to access anything on the `window` scope.
This causes the snowplow client to not want to even start, with the following error (visible from within Retool in the screenshot I will post in a reply)

`startSnowplough: Failed to read the 'localStorage' property from 'Window': The document is sandboxed and lacks the 'allow-same-origin' flag.`

I have made a very crude PR to what I think might be a solution to this problem, but it's pretty rough and probably there are better ways to do this.

### DIFF
--- a/libraries/browser-tracker-core/src/detectors.ts
+++ b/libraries/browser-tracker-core/src/detectors.ts
@@ -35,7 +35,7 @@
  */
 export function hasSessionStorage() {
   try {
-    return !!window.sessionStorage;
+    return window && !!window.sessionStorage;
   } catch (e) {
     return true; // SecurityError when referencing it means it exists
   }
@@ -48,7 +48,7 @@ export function hasSessionStorage() {
  */
 export function hasLocalStorage() {
   try {
-    return !!window.localStorage;
+    return window && !!window.localStorage;
   } catch (e) {
     return true; // SecurityError when referencing it means it exists
   }
--- a/libraries/browser-tracker-core/src/detectors.ts
+++ b/libraries/browser-tracker-core/src/detectors.ts
@@ -28,6 +28,19 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+const isSandboxedError = (error: Error): boolean => {
+  if (error.name === 'SecurityError') {
+    if (!error.message) {
+      return false;
+    }
+    const matches = error.message.match(/sandboxed/);
+    if (matches && matches.length === 1) {
+      return false;
+    }
+  }
+  return false;
+};
+
 /*
  * Checks whether sessionStorage is available, in a way that
  * does not throw a SecurityError in Firefox if "always ask"
@@ -37,7 +50,10 @@ export function hasSessionStorage() {
   try {
     return window && !!window.sessionStorage;
   } catch (e) {
-    return true; // SecurityError when referencing it means it exists
+    // SecurityError when referencing it means it exists
+    // SecurityError can also be a sandbox error:
+    // if this is the case, session storage is not available.
+    return !isSandboxedError(e);
   }
 }
 
@@ -50,7 +66,10 @@ export function hasLocalStorage() {
   try {
     return window && !!window.localStorage;
   } catch (e) {
-    return true; // SecurityError when referencing it means it exists
+    // SecurityError when referencing it means it exists
+    // SecurityError can also be a sandbox error:
+    // if this is the case, session storage is not available.
+    return !isSandboxedError(e);
   }
 }
 


snowplowcla:

michahell:
I signed it! using a mid air hand gesture

michahell:
by default 2021-06-01 at 18 35 35

paulboocock:
Interesting! Thanks for the PR, we're a little busy to look at this in much depth over the next couple of weeks but I'll find time as soon as possible to dig into this.

There's quite a lot of functionality in the tracker that depends on the `window` object beyond localStorage however, such as various event handlers, are these fine in the sandbox?

Additionally, if it's just localStorage, could you try and set `stateStorageStrategy: 'none'` when initialising the tracker? We shouldn't check for `window.localStorage` then.

```javascript
window.snowplow('newTracker', 'sp', '<<MY_COLLECTOR>>', {
  appId: 'my-app',
  stateStorageStrategy: 'none'
});
```

Side note: As a Brit, I very much enjoyed the renaming to plough 😉

michahell:
@paulboocock I was already thinking of possible other functionality depending on `window.` after I made the PR 😬 
I am not sure what is and what is not supported - still trying to figure out what this security sandbox actually entails, and how they enforce this.
I need Retool support to get back to me about that. Will update here when I know more

Yes! As a non-native speaker, whenever I think of writing 'plough' I apparently default to the British written form and not the American: somehow the latter doesn't sit right with me 🤔 

I have tried both: `stateStorageStrategy: 'none'` and `stateStorageStrategy: 'cookie'` and somehow those still trigger the error messages. I did not yet reverse-engineer what actually happens when these are set, but I would also assume they would result in not doing these checks. let me retest this, maybe this was cache or some other issue at work.

paulboocock:
With regards to `stateStorageStrategy: 'none'`, this line means we check for localStorage before checking if we are allowed to use local storage, we should check `useLocalStorage` before calling `localStorageAccessible()`. Seems pointless to check it if we're not permitted by configuration to use it. I've opened https://github.com/snowplow/snowplow-javascript-tracker/issues/974 to track that.
https://github.com/snowplow/snowplow-javascript-tracker/blob/268f6cef26edb87b51fda39126fd7ffbc41dff5b/libraries/browser-tracker-core/src/tracker/out_queue.ts#L109

The logic of the `try { } catch { }` within `hasLocalStorage()` is also rather interesting in this part of the code base. We assume when it errors that this means the browser supports localStorage, as this is more for feature detection rather than availability of it. We wrap this in a `localStorageAccessible()` which then tries to write to it and if that fails then we don't use it.

What I find interesting about your error, I'd expect `localStorageAccessible()` to return `false` and I'd expect things to still work since this part of the code base is full of try-catch statements. The tracker should continue just without any local storage functionality but it seems like the script is killed or something(?) for trying to access localStorage.

michahell:
Yes, exactly this is what is think is happening. I don't know if this is Retool specific (i.e. Retool has built their own security proxy system around this and then purposefully fails when it detects usage) which I would find hard to believe, or this is default (chrome) browser behaviour, which is a lot easier to assume (and in that case interesting for snowplough to try to cater for ;) )
Hmm... that said actually, I have not checked what happens when testing in other browsers, let me do this as well.
For me this is also new, just like the sandbox environment: usually whenever I'd had a need for using session or localStorage I would also check for existence and **then** if getting and setting an item works, just like is done in the tracker.

michahell:
Okay, so Retool is using a sandboxed Iframe: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox. I've asked which restrictions are in place, I suspect at least `allow-same-origin`, which probably is the culprit 😬

paulboocock:
Ah ok. I've knocked together a quick build and I'm wondering if this changes/fixes the error you get. Could you test this out for me? You'll need to use `stateStorageStrategy: 'none'`.

```javascript
window.snowplow('newTracker', 'sp', '<<MY_COLLECTOR>>', {
  appId: 'my-app',
  stateStorageStrategy: 'none'
});
```

[sp.zip](https://github.com/snowplow/snowplow-javascript-tracker/files/6583103/sp.zip)

michahell:
Okay thanks @paulboocock, much appreciated.
Let me figure out a way to host this somewhere in our domain (or not) and get back to you!

paulboocock:
@michahell If you want to host it for testing, you could drop it into a new public Github repository and enable Github pages on that repository.

michahell:
Retool support came back to me with this:
===
Hi again!
We looked into this further last night, it it looks like we use **all of the sandbox restrictions by default** (due to the empty string, which would be the options to remove).
// enumeration of all security restrictions
===
I chose for hosting the file on our client's bitbucket static hosting, will be up soon so I can test.

michahell:
I have modified the tracker init script as follows:
```Javascript
;(function(p,l,o,w,i,n,g){if(!p[i]){p.GlobalSnowplowNamespace=p.GlobalSnowplowNamespace||[]; p.GlobalSnowplowNamespace.push(i);p[i]=function(){(p[i].q=p[i].q||[]).push(arguments) };p[i].q=p[i].q||[];n=l.createElement(o);g=l.getElementsByTagName(o)[0];n.async=1; n.src=w;g.parentNode.insertBefore(n,g)}}(window,document,"script","https://carepaydev.bitbucket.io/sp.js","snowplow"));
```
but when I try this in Retool, it still gives me the same warning @paulboocock

paulboocock:
Can you send me the `newTracker` configuration object too? I'll see if I can reproduce to too then.

michahell:
ah yes ofc, sorry 😬 

This is the barebone tracker config I have right now (script triggered by a button: every Js has to be triggered by user action in Retool)

```JavaScript
        const namespace = 'tracker_retool_apps';
        const collectorUrl = 'payer.ke-test.carepay.dev/api/snowplow';
        const title = 'retool app snowplough test';
        
        window.snowplow('newTracker', namespace, collectorUrl, {
          appId: 'retool_app_snowplough_test',
          anonymousTracking: { withServerAnonymisation: true },
          stateStorageStrategy: 'none',
          eventMethod: 'post',
          contexts: {
            webPage: true // default, can be omitted
          }
        });
        
        window.snowplow('enableActivityTracking', {
          minimumVisitLength: 30,
          heartbeatDelay: 10
        });
        
        window.snowplow('trackPageView', { title });
```

And this is what I have on the other button in Retool:

```JavaScript
        console.log('window.snowplow: ', window.snowplow);
        
        window.snowplow('trackStructEvent', {
          category: 'Tests',
          action: 'retool button',
          label: 'buttonSendEvent',
          property: 'iglu-is-awesome',
          value: true
        });
```

michahell:
@paulboocock if it is any help, I think I can export the Retool app that I built into some Retool format, and you could import it as a working app, and then go from there?

michahell:
@paulboocock Hi! any updates on this? we would really like to track our Retool usage :(

paulboocock:
Hey @michahell sorry for the radio silence! It's on my list and heading towards the top, it's been a busy couple of weeks with our company "away" week and then loads to catch up on. I'll aim to have a look by the end of this week at the latest!

michahell:
@paulboocock no worries! just happy with your ping here :)
I think you will be able to replicate it without Retool indeed, the fully sandboxed Iframe should be the thing restricting things and nothing else, AFAIK!
Thanks!

paulboocock:
Checking in @michahell - I'm back working on the JS Tracker now so this will be something I look into soon.

michahell:
@paulboocock Welcome back, hope you had a great company away week! getting back to this would be grand! Unfortunately, I won't have much more time (about 3 weeks) remaining while working for my current client :( But at least the issue is clear, and you know that this is probably interesting for snowplow to also want to fix ;) so more sandboxed apps can also do telemetry.

paulboocock:
Quick update, I've got the tracker working in a sandboxed iframe. I had to rip out a reasonable chunk of code in the out queue processing but nothing insurmountable. I'll publish a beta for you to try once it's ready (later today).

michahell:
@paulboocock Ah this is great news! A colleague of mine will be taking the task of implementing Snowplow in Retool over, but I've kept our end up to date on progress here 🎉

paulboocock:
Hi @michahell 
I'm going to close this in favour of #983 and #981 (issue for discussion)

I've just published a beta which I hope will resolve this: https://cdn.jsdelivr.net/npm/@snowplow/javascript-tracker@3.1.1-beta.1/dist/sp.js

michahell:
@paulboocock thanks Paul! do you want me to report back after testing here, or in #981 ?

michahell:
@paulboocock i'll reply in #981 :)

PR #993, opened 2021-08-14
Release/3.1.2

paulboocock:
Small PR to allow for `window.snowplow('setUserId', undefined);` to match what you could do with v2 of the tracker., this allows for `null` and `undefined` to unset a user id. I also included the opt out cookie name in this, as you may wish to unset that too in the same way.


adatzer (approved):
LGTM!

PR #1001, opened 2021-09-09
Initial commit

greg-el:


plugins/browser-plugin-media-tracking/src/eventGroups.ts
```diff
@@ -0,0 +1,20 @@
+import { SnowplowMediaEvent } from './snowplowEvents';
+import { EventGroup } from './types';
+import { DocumentEvent, MediaEvent, TextTrackEvent, VideoEvent } from './wgEvents';
+
+export const AllEvents: any[] = Object.values(MediaEvent);
+for (let e of Object.values(SnowplowMediaEvent)) {
+  AllEvents.push(e);
+}
+
```
paulboocock, line 9: I think you can spread these together. I haven't tried this...

```
export const AllEvents: any[] = { ...Object.values(MediaEvent), ...Object.values(SnowplowMediaEvent) };
```

However... I don't think `Object.values` works in IE 😬 Might need to find a new strategy for this. Oh and no `any`s 😛

plugins/browser-plugin-media-tracking/src/helperFunctions.ts
```diff
@@ -0,0 +1,71 @@
...
+      language: textTrackList[i].language,
+      kind: textTrackList[i].kind,
+      mode: textTrackList[i].mode,
+    });
+  }
+  return out;
+}
+
+export function isTypeTextTrackEvent(e: string): boolean {
+  let fields: string[] = Object.values(TextTrackEvent);
```
paulboocock, line 33: Just highlighting more `Object.values` here.

plugins/browser-plugin-media-tracking/src/helperFunctions.ts
```diff
@@ -0,0 +1,71 @@
...
+  return fields.includes(e);
+}
+
+export function isTypeDocumentEvent(e: string): boolean {
+  let fields: string[] = Object.values(DocumentEvent);
```
paulboocock, line 38: `Object.values`

plugins/browser-plugin-media-tracking/src/consts.ts
```diff
@@ -0,0 +1,14 @@
+export const ready_state_consts: Record<number, string> = {
+  0: 'HAVE_NOTHING',
+  1: 'HAVE_METADATA',
+  2: 'HAVE_CURRENT_DATA',
+  3: 'HAVE_FUTURE_DATA',
+  4: 'HAVE_ENOUGH_DATA',
+};
+
+export const network_state_consts: Record<number, string> = {
```
paulboocock, line 9: I think I'd rename these two constants to be `READY_STATE` and `NETWORK_STATE`

plugins/browser-plugin-media-tracking/src/index.ts
```diff
@@ -0,0 +1,348 @@
...
+
+export function MediaTrackingPlugin(): BrowserPlugin {
+  return {
+    activateBrowserPlugin: (tracker: BrowserTracker) => {
+      _trackers[tracker.id] = tracker;
+    },
+  };
+}
+
+export function trackMedia(mediaId: string, config: MediaTrackingConfig = {}): void {
```
paulboocock, line 74: This was `TrackMedia` in README, but is `trackMedia` here. Although my earlier comment still stands, I think this can have a better name.

paulboocock: nit: Personal preference I think from a consistency point of view.

```suggestion
export function trackMedia(mediaId: string, config: MediaTrackingConfig = {}) {
```

plugins/browser-plugin-media-tracking/src/helperFunctions.ts
```diff
@@ -0,0 +1,71 @@
...
+  return fields.includes(e);
+}
+
+function getDuplicatesInArray(arr: number[]) {
+  return arr.filter((el, i) => arr.indexOf(el) !== i);
+}
+
+export function checkPercentBoundryArrayIsValid(percentBoundries: number[]): void {
```
paulboocock, line 46: You could proably use an `is` here. Sort of like this: https://github.com/snowplow/snowplow-javascript-tracker/blob/c2e3ccb94f9ff5fad7ea295d36bffa0c5cd2f819/libraries/browser-tracker-core/src/tracker/out_queue.ts

plugins/browser-plugin-media-tracking/src/helperFunctions.ts
```diff
@@ -0,0 +1,71 @@
...
+  let errorValues: number[] = [];
+  percentBoundries.forEach((p) => {
+    if (p < 0 || 100 < p) {
+      errorValues.push(p);
+    }
+  });
+  if (errorValues.length) {
+    throw new Error(`Percent bounds must be 1 - 100 inclusive. Adjust or remove the following: [${errorValues}]`);
```
paulboocock, line 54: We don't usually throw errors, we handle invalid state and then log an error. This ensure we don't cause issues for applications which host the tracker but never leaking an error (hopefully).

I think turning this into an `is` and then checking this where it is called and logging an error if invalid is the "classic" option in the tracker.

plugins/browser-plugin-media-tracking/src/index.ts
paulboocock: This function is quite hard to read. the mix of functions, variables and actual logic makes it pretty confusing. 

I'd suggest pulling the functions out (nested functions are never that nice). I think that'd probably solve most issues. Then group the variables and then the logic so they live together in this function.

plugins/browser-plugin-media-tracking/src/wgEvents.ts
```diff
@@ -0,0 +1,45 @@
+// https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement#events
```
paulboocock, line 1: nit: `wgEvents.ts` isn't that obvious what `wg` is. Maybe `mediaEvents.ts` is an ok name?

plugins/browser-plugin-media-tracking/src/wgProperties.ts
```diff
@@ -0,0 +1,44 @@
+// https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement
```
paulboocock, line 1: nit: Same here on the `wg` comment

paulboocock (commented):
Initial review complete and it's coming together! A few comments that I think need tidying up but it's definitely getting there.

PR #1003, opened 2021-09-17
Fix linkDecorationHandler targeting (closes #1002)

### LINKED ISSUE #1002: Issue in crossDomain link decoration
## What 
We observed that the 3.0.3 snowplow js tracker is decorating the wrong elements.

## How to reconstruct
In a situation like this
```
    <a href="abc">
        <img src.../>
    </a>
```
the sp.js tracker would add an event listener to the a-Tag, if abc matches the crossDomainLinker criterion. Now if the image is clicked. The event listener receives and event `e` and decorates `a.target.href`, but it should decorate `a.currentTarget.href`. Thus, the links default action will follow an undecorated link. The responsible code is here: 

https://github.com/snowplow/snowplow-javascript-tracker/blob/c79a74e5b397fc4c58230a04864e7cb23ae62ed8/libraries/browser-tracker-core/src/tracker/index.ts#L327

In earlier versions of the sp.js tracker the event listener ignored the incoming event, i.e. like `() => { decoration; }`, and then decorate the href of `this`, which actually equals `e.currentTarget`.

Newer versions of the sp.js tracker are also impacted.

philipherrmann:


snowplowcla:

philipherrmann:
I signed it!

snowplowcla:
Confirmed! @philipherrmann has signed the Contributor License Agreement. Thanks so much.

PR #1019, opened 2021-11-26
Add support for custom headers

paulboocock:
Add support for custom headers in the tracker configuration

You can now do this:

```javascript
window.snowplow('newTracker', 'sp', collector_endpoint, {
  appId: 'sp',
  customHeaders: {
    'Content-Language': 'de-DE, en-CA',
  },
});
```

Headers must either be CORS safe (https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header) or returned in the `Access-Control-Allow-Headers` response header (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers).

Note: The collector has a static `Access-Control-Allow-Headers` response header, so this is only useful when a Proxy is set up in front of the collector. Beneficial if someone is using a collector with some Auth for instance (See discourse post linked in Issue).

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/issue-1010_custom_headers_2021-11-26-16-12.json
+++ b/common/changes/@snowplow/browser-tracker-core/issue-1010_custom_headers_2021-11-26-16-12.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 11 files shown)


greg-el (approved):
LGTM!

PR #1021, opened 2021-12-16
 Allow alternative Access-Control-Allow-Credentials values (close #808)

### LINKED ISSUE #808: Allow alternative Access-Control-Allow-Credentials values
We had a security advisor company perform audit and they pointed out current potential vulnerability of snowplow collector code (for us risk is low but depending on case by case implementations could be high). They suggested that `Access-Control-Allow-Origin should` always be set to `*` but `Access-Control-Allow-Credentials` should always be set to `false` as that will ensure that attacker have absolutely no chance of stealing cookies.
This would require adjustment to `CollectorService.scala` to modify [this block](https://github.com/snowplow/snowplow/blob/3b8d9cc839e4af0b97c68477fb1c9f484de233e2/2-collectors/scala-stream-collector/core/src/main/scala/com.snowplowanalytics.snowplow.collectors.scalastream/CollectorService.scala#L416) as well as adding `Access-Control-Allow-Credentials` to header values.

MitchPierias:
In some implementations of Snowplow, the collection server may specify `access-control-allow-origin` header to the wildcard `'*'`. This is commonly used in staging or development environments where requests are intentionally made cross-domain. In this case, the XHRHttpRequest is rejected when `access-control-allow-origin='*'` and `withCredentials=true` with the following exception;

> Access to XMLHttpRequest at 'https://collection.example.com/i?QUERY_PARAMS' from origin 'http://localhost:3000' has been blocked by CORS policy: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'. The credentials mode of requests initiated by the XMLHttpRequest is controlled by the withCredentials attribute.

The browser terminates the request because it is not safe to include credentials when `access-control-allow-origin` accepts any origin. In theory, other sites could send cross-origin requests with credentials from the client's website, raising a genuine security concern. However, in staging or development environments, we might intentionally make cross-origin requests to a collection server that does not require credentials. Setting the `access-control-allow-credentials` to `false` will prevent cookies, authorization headers, and TLS certificates from being included in the request, and allow the scenario above to complete successfully.

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/index.ts
+++ b/libraries/browser-tracker-core/src/tracker/index.ts
@@ -261,7 +261,8 @@ export function Tracker(
         trackerConfiguration.useStm ?? true,
         trackerConfiguration.maxLocalStorageQueueSize ?? 1000,
         trackerConfiguration.connectionTimeout ?? 5000,
-        configAnonymousServerTracking
+        configAnonymousServerTracking,
+        !!trackerConfiguration.allowCredentials ?? true
       ),
       // Whether pageViewId should be regenerated after each trackPageView. Affect web_page context
       preservePageViewId = false,
--- a/libraries/browser-tracker-core/src/tracker/out_queue.ts
+++ b/libraries/browser-tracker-core/src/tracker/out_queue.ts
@@ -57,6 +57,7 @@ export interface OutQueue {
  * @param maxLocalStorageQueueSize - Maximum number of queued events we will attempt to store in local storage
  * @param connectionTimeout - Defines how long to wait before aborting the request
  * @param anonymousTracking - Defines whether to set the SP-Anonymous header for anonymous tracking on GET and POST
+ * @param allowCredentials - Sets the value of the `access-control-allow-credentials` in the request headers
  * @returns object OutQueueManager instance
  */
 export function OutQueueManager(
@@ -70,7 +71,8 @@ export function OutQueueManager(
   useStm: boolean,
   maxLocalStorageQueueSize: number,
   connectionTimeout: number,
-  anonymousTracking: boolean
+  anonymousTracking: boolean,
+  allowCredentials: boolean
 ): OutQueue {
   type PostEvent = {
     evt: Record<string, unknown>;
@@ -401,7 +403,7 @@ export function OutQueueManager(
     } else {
       xhr.open('GET', url, !sync);
     }
-    xhr.withCredentials = true;
+    xhr.withCredentials = allowCredentials;
     if (anonymousTracking) {
       xhr.setRequestHeader('SP-Anonymous', '*');
     }
--- a/libraries/browser-tracker-core/src/tracker/types.ts
+++ b/libraries/browser-tracker-core/src/tracker/types.ts
@@ -93,6 +93,12 @@ export type TrackerConfiguration = {
    */
   cookieLifetime?: number;
   /**
+   * Sets the `Access-Control-Allow-Credentials` header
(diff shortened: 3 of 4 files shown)


snowplowcla:

paulboocock:
I'm open to the idea of making this configurable, I'm curious for what reasons you'd like to change this flag though? There's already a lot of configuration options in the tracker, whilst powerful it does up the complexity somewhat, although with good default values that is partially negated.

I know sometimes it comes back in pen tests as a warning, but in terms of what the tracker does when it comes to "third party" events having it default to `true` is the most compatible option and I'm not sure why you'd want it be `false`. As far as I know, it makes no difference at all when its a same-site request.

With the PR though, this is currently a breaking change as the default you've used it `false`. We'd need that to be `true` to accept this PR (and we'd need you to sign the CLA too).

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -261,7 +261,8 @@ export function Tracker(
         trackerConfiguration.useStm ?? true,
         trackerConfiguration.maxLocalStorageQueueSize ?? 1000,
         trackerConfiguration.connectionTimeout ?? 5000,
-        configAnonymousServerTracking
+        configAnonymousServerTracking,
+        !!trackerConfiguration.allowCredentials ?? true
```
MitchPierias, line 265: Ensures even a string value for `allowCredentials` is converted to a `boolean`, in the case of a non-typescript implementation.

MitchPierias:
I signed it!

snowplowcla:
Confirmed! @MitchPierias has signed the Contributor License Agreement. Thanks so much.

MitchPierias:
> I'm open to the idea of making this configurable, I'm curious for what reasons you'd like to change this flag though? There's already a lot of configuration options in the tracker, whilst powerful it does up the complexity somewhat, although with good default values that is partially negated.
> 
> I know sometimes it comes back in pen tests as a warning, but in terms of what the tracker does when it comes to "third party" events having it default to `true` is the most compatible option and I'm not sure why you'd want it be `false`. As far as I know, it makes no difference at all when its a same-site request.

Hi Paul, thank you for taking a look at this pull request. I'd like the option to disable credentials for use in development and staging environments when the collection server does not require credentials and cross-origin is intentional. In development, for example, the `access-control-allow-origin` of the collection server is often set to the wildcard `'*'` to allow requests from `localhost`. In this instance, the browser terminates the request because `withCredentials` is always set to true, even when we don't want to provide credentials, and any origin can make requests.

> With the PR though, this is currently a breaking change as the default you've used it `false`. We'd need that to be `true` to accept this PR (and we'd need you to sign the CLA too).

I've now updated the default value and related comments to `false` to prevent a breaking change.

paulboocock:
Hey @MitchPierias 

My colleague @greg-el is going to review this, but I don't forsee any issues. I've cherry-picked your commit into a new branch based off the release/3.2.0 branch and fixed the conflicts, as it's in conflict with the next release now. No big issue, just two new configuration options being added to out_queue.

For most users who want this, I anticipate a change to the collector will be needed as described in #808 but we will deal with that later.

MitchPierias:
Hey @paulboocock, how long until we can expect the 3.2.0 release?

paulboocock:
> Hey @paulboocock and @greg-el, how long until we can expect the 3.2.0 release?

It'll be out in January. It's quite a big release but it's all currently in test so shouldn't be too long now.

PR #1022, opened 2021-12-21
Allow alternative Access-Control-Allow-Credentials values (close #808)

### LINKED ISSUE #808: Allow alternative Access-Control-Allow-Credentials values

paulboocock:
Cherry picked and resolved conflicts of #1021 for @greg-el to review before merging into 3.2.0 release.

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/issue-808_2021-12-21-20-04.json
+++ b/common/changes/@snowplow/browser-tracker-core/issue-808_2021-12-21-20-04.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
+      "comment": "Allow alternative Access-Control-Allow-Credentials values (close #808)",
+      "type": "none"
+    }
+  ],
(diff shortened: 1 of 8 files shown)


MitchPierias (approved):

MitchPierias:
Thanks for including the `withCredentials` changes @paulboocock

greg-el (approved):
LGTM!

PR #1041, opened 2022-01-17
Bump Copyright to 2022 (close #1040)

### LINKED ISSUE #1040: Bump Copyright to 2022
We should bump the copyright to 2022, in particular in `banner.js` which is written as a head to distributed files.

We should also ensure the headers at the top of every file read 2022 too.

In the Media and Youtube Tracking plugin, they have their own `banner.js`. I think they can use the common one in the root of the project `import { banner } from '../../banner';` in their `rollup.config.js` and use `banner(true)` for the alternative copyright.

paulboocock:
- #1040


greg-el (approved):
LGTM, can't find any stray 2021's anywhere.

PR #1048, opened 2022-01-26
 Fix 'paused' property never being true (close #1047)

### LINKED ISSUE #1047: Fix 'paused' property never being true
**Describe the bug**
`paused` is never true in `com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0`

greg-el:


snowplowcla:

plugins/browser-plugin-youtube-tracking/src/buildYouTubeEvent.ts
```diff
@@ -74,7 +74,7 @@ function getMediaPlayerEntities(
 ): MediaEntities {
   const playerStates: Record<string, boolean> = {
     ended: false,
-    paused: false,
+    pause: false,
```
paulboocock, line 77: Would it read better to make the change to the `YTStateEvent` enum in `constants.ts` so it can remain as `paused` here? That aligns better with the Youtube API docs:
Screenshot 2022-01-27 at 11 28 50

paulboocock (commented):
Just one comment here, bit of a nit pick but might read better.

plugins/browser-plugin-youtube-tracking/src/buildYouTubeEvent.ts
```diff
@@ -73,22 +73,22 @@ function getMediaPlayerEntities(
   eventData?: EventData
 ): MediaEntities {
   const playerStates: Record<string, boolean> = {
-    ended: false,
-    paused: false,
+    [YT.PlayerState.ENDED]: false,
+    [YT.PlayerState.PAUSED]: false,
```
paulboocock, line 77: This feels a little weird to me as using numbers as keys in an object feels a bit strange (and now wrongly typed). Its an enum (and a number), so should be `Record<number, boolean>`. However , I'm not particularly picky as we used to map the number to a the string in the previous commit which I don't see is any real difference. Maybe just fix the type.

This function is now also quite different to `getYouTubeEntities` above, however it has the same behaviour. It'd be nice to make them consistent.

plugins/browser-plugin-youtube-tracking/src/buildYouTubeEvent.ts
```diff
@@ -22,26 +21,26 @@ export function buildYouTubeEvent(player: YT.Player, eventName: string, conf: Tr
 function getYouTubeEntities(player: YT.Player, urlParameters: UrlParameters, eventData?: EventData): MediaEntities {
   const spherical: YT.SphericalProperties = player.getSphericalProperties();
   const playerStates: Record<string, boolean> = {
```
paulboocock, line 23: `number`

paulboocock (approved):

PR #1054, opened 2022-01-27
 Add ready to Media Plugin DefaultEvents (close #1053)

### LINKED ISSUE #1053: Add `ready` to Media Plugin DefaultEvents
Add a custom `ready` event to the media tracking plugin, for use in Snowplow modelling. This will fire when the video element has been found. This will be used to work out the 'impressions' of the video (a count of every time content is served to a user, regardless of if it is played). Whilst this isn't a perfect way of going about it, as it doesn't account for the video being in the viewport or not, it's the best way with events solely the media tracking plugin.

note: Related to #1051, however `loadedmetadata` proved to be unreliable for the purpose.

greg-el:
- #1053

### DIFF
--- a/plugins/browser-plugin-media-tracking/src/eventGroups.ts
+++ b/plugins/browser-plugin-media-tracking/src/eventGroups.ts
@@ -12,6 +12,7 @@ export const AllEvents: EventGroup = [
 ];
 
 export const DefaultEvents: EventGroup = [
+  SnowplowEvent.READY,
   MediaEvent.PAUSE,
   MediaEvent.PLAY,
   MediaEvent.SEEKED,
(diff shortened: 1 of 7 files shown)


snowplowcla:

paulboocock (commented):
Looks good, just need those rush change files and a test run.

plugins/browser-plugin-media-tracking/src/eventGroups.ts
```diff
@@ -12,6 +12,7 @@ export const AllEvents: EventGroup = [
 ];
 
 export const DefaultEvents: EventGroup = [
+  SnowplowEvent.READY,
```
paulboocock, line 15: I also wonder if these new defaults are enough to warrant a v3.3 or if we think they are bug fixes as we should have been capturing a ready even all along? It doesn't quite feel significant enough for a v3.3 though but it's tempting as its a behaviour change and easier to recommend from a recommended tracker version for data modelling point of view.

paulboocock (approved):

Issue #1061, opened 2022-02-22
Stop mutating context array with additional plugin contexts


paulboocock:
**Describe the bug**
The Tracker Core currently mutates the context argument which can be passed on the `track` method. All the plugin contexts are added to this same array which can have undersirable consequences, such as adding the same context multiple times if the original `context` property is reused by the plugin.

**To Reproduce**
It's possible to reproduce this issue with the HTML5 Video Tracking. 

1. Set up two trackers (i.e. call `newTracker` twice).
2. Track events with a video element
3. Notice the second tracker will have the `webPage` context attached twice

**Expected behavior**
The second tracker should only have the `webPage` context attached once

PR #1062, opened 2022-02-22
Release/3.3.1

paulboocock:
- #1061


matus-tomlein (approved):
LGTM! The function name `addPluginContexts` is slightly confusing to me as I first expected it to be mutating the `PluginContexts` instance, probably would prefer `returnWithPluginContexts`, but that's not important or in scope for this change.

Issue #362, opened 2015-04-09
Investigate having a maximum size for GET requests


fblundun:
Otherwise a single huge event might never leave localStorage because it is too big to send successfully.

alexanderdean:
This will block further events from sending.

alexanderdean:
I'm not sure why I wanted to hold back on this one... It feels like maxGetBytes would be a sensible option to add... /cc @yalisassoon

javiermar:
Is there an update on this front? It's 2019 and we're seeing this problem in our environment.

paulboocock:
Investigating this seems a worthwhile thing to do in the last release of v2. The risk of getting in a "stuck" state with local storage and GET together is an unfortunate issue. 
General advice is to only use POST with localStorage queues (tracker defaults) but the tracker is quite happy to be configured with GET and localStorage queues.

matus-tomlein:
Addressed in #449

PR #1064, opened 2022-03-08
Add configurable max GET request size in bytes attribute (close #449)

### LINKED ISSUE #449: Add configurable max GET request size in bytes attribute
Unfortunately it isn't possible to tell from JavaScript whether the GET response had status code 414 ([source](http://stackoverflow.com/questions/8108636/how-to-get-http-status-code-of-img-tags)). We could instead have a configurable maximum GET size to match the configurable maximum POST size.

Or maybe we should just tell people who are trying to send large events to use POST instead.

matus-tomlein:
This PR addresses issues #449 and #362.

It adds the `maxGetBytes` configuration parameter that sets the maximum size for GET requests. The size in bytes is checked for the full GET request URL. In case requests are over this size limit, an error is written to `console.log` and the event is ignored (not added to queue).

The default behaviour remains that the request size for GET is unlimited.

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/out_queue.ts
+++ b/libraries/browser-tracker-core/src/tracker/out_queue.ts
@@ -53,6 +53,7 @@ export interface OutQueue {
  * @param postPath - The path where events are to be posted
  * @param bufferSize - How many events to batch in localStorage before sending them all
  * @param maxPostBytes - Maximum combined size in bytes of the event JSONs in a POST request
+ * @param maxGetBytes - Maximum size in bytes of the complete event URL string in a GET request
  * @param useStm - Whether to add timestamp to events
  * @param maxLocalStorageQueueSize - Maximum number of queued events we will attempt to store in local storage
  * @param connectionTimeout - Defines how long to wait before aborting the request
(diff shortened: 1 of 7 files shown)


snowplowcla:

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -69,6 +70,7 @@ export function OutQueueManager(
   postPath: string,
   bufferSize: number,
   maxPostBytes: number,
+  maxGetBytes: number | boolean,
```
paulboocock, line 73: `numer | boolean` is an interesting choice. I think if you passed in `true` then this would break, or at least have unexpected behaviour. You'd end up doing `if (bytes >= true) {` on Line 240.

You could do `number | false` instead, or just `number` and have the behaviour of `0` (or less I guess) not applying a max bytes?

paulboocock (commented):
Other than my one comment, this looks good.

libraries/browser-tracker-core/src/tracker/out_queue.ts
matus-tomlein: Good point, I didn't consider the `true` case. Changed it to only accept numbers where 0 or less is considered as unlimited.

PR #1069, opened 2022-03-23
Enable form tracking on forms embedded within iframes where possible (close #1066)

### LINKED ISSUE #1066: Enable form tracking on forms embedded within iframes where possible
For iframes that are not retrieved from a different domain than the parent page, it is possible to access their elements and observe them. This makes it possible to enable form tracking on embedded forms such as from HubSpot.

matus-tomlein:
Issue #1066 

This PR adds the ability to set a collection of form elements to be tracked in the `options.forms` argument in `enableFormTracking`. This enables a user to find form elements inside an iframe (if not blocked by the browser) and pass them to the tracker when enabling form tracking on the parent page. This was not possible before as form elements were searched for only on the parent page not considering any embedded iframes.

The `options.forms` attribute in `enableFormTracking` can now serve multiple use cases as it also accepts the `FilterCriterion` filter for finding form elements on the page as well as a collection of specific form elements to track.

## Documentation: Tracking forms embedded inside iframes

The options for tracking forms inside of iframes are limited – browsers block access to contents of iframes that are from different domains than the parent page. We are not able to provide a solution to track events using trackers initialized on the parent page in such cases. However, since version 3.4, it is possible to track events from forms embedded in iframes loaded from the same domain as the parent page or iframes created using JavaScript on the parent page (e.g., HubSpot forms).

In case you are able to access form elements inside an iframe, you can pass them in the `options.forms` argument when calling `enableFormTracking` on the parent page. This will enable form tracking for the specific form elements. The feature may also be used for forms not embedded in iframes, but it's most useful in this particular case.

The following example shows how to identify the form elements inside an iframe and pass them to the `enableFormTracking` function:

```js
let iframe = document.getElementById('form_iframe'); // find the element for the iframe
let forms = iframe.contentWindow.document.getElementsByTagName('form'); // find form elements inside the iframe
enableFormTracking({
    options: {
        forms: forms // pass the embedded forms when enabling form tracking
    },
});
```

## Other resources

The documentation is also in a PR on data-value-resources [here](https://github.com/snowplow-incubator/data-value-resources/pull/66/commits/e09e887d5289614d933bea705db1097f367370a6).

[Here](https://github.com/matus-tomlein/snowplow-javascript-tracker-playground/blob/main/react-app/src/routes/iframeForm.jsx#L18-L23) is an sample example in an app that adds a form to an iframe using JavaScript and tracks it.

### DIFF
--- a/plugins/browser-plugin-form-tracking/src/helpers.ts
+++ b/plugins/browser-plugin-form-tracking/src/helpers.ts
@@ -70,7 +70,7 @@ const defaultFormTrackingEvents = [
 ];
 
 export interface FormTrackingOptions {
-  forms?: FilterCriterion<HTMLElement>;
+  forms?: FilterCriterion<HTMLElement> | HTMLCollectionOf<HTMLElement>;
   fields?: FilterCriterion<TrackedHTMLElement> & { transform: transformFn };
   events?: FormTrackingEvents;
 }
@@ -113,7 +113,8 @@ export function addFormListeners(tracker: BrowserTracker, configuration: FormTra
     trackingMarker = tracker.id + 'form',
     config = getConfigurationForOptions(options);
 
-  Array.prototype.slice.call(document.getElementsByTagName('form')).forEach(function (form: HTMLFormElement) {
+  var forms = config.forms ?? document.getElementsByTagName('form');
+  Array.prototype.slice.call(forms).forEach(function (form: HTMLFormElement) {
     if (config.formFilter(form)) {
       Array.prototype.slice.call(innerElementTags).forEach(function (tagname: keyof TrackedHTMLElementTagNameMap) {
         Array.prototype.slice
@@ -160,14 +161,30 @@ export function addFormListeners(tracker: BrowserTracker, configuration: FormTra
  */
 function getConfigurationForOptions(options?: FormTrackingOptions) {
   if (options) {
+    var formFilter = (_: HTMLElement) => true;
+    var forms: HTMLCollectionOf<HTMLFormElement> | null = null;
+    if (
+      !options.forms ||
+      (options.forms as any).allowlist ||
+      (options.forms as any).denylist ||
+      (options.forms as any).filter
+    ) {
+      // options.forms is null or a filter
+      formFilter = getFilterByClass(options.forms as FilterCriterion<HTMLElement> | undefined);
+    } else if ((options.forms as any).length !== undefined) {
+      // options.forms is a collection of HTML form elements
+      forms = options.forms as HTMLCollectionOf<HTMLFormElement>;
+    }
     return {
-      formFilter: getFilterByClass(options.forms),
+      forms: forms,
+      formFilter: formFilter,
       fieldFilter: getFilterByName<TrackedHTMLElement>(options.fields),
       fieldTransform: getTransform(options.fields),
       eventFilter: (event: FormTrackingEvent) => (options.events ?? defaultFormTrackingEvents).indexOf(event) > -1,
     };
   } else {
     return {
+      forms: null,
       formFilter: () => true,
       fieldFilter: () => true,
       fieldTransform: defaultTransformFn,
--- a/plugins/browser-plugin-form-tracking/src/helpers.ts
+++ b/plugins/browser-plugin-form-tracking/src/helpers.ts
@@ -70,7 +70,7 @@ const defaultFormTrackingEvents = [
 ];
 
 export interface FormTrackingOptions {
-  forms?: FilterCriterion<HTMLElement> | HTMLCollectionOf<HTMLElement> | NodeListOf<HTMLFormElement>;
+  forms?: FilterCriterion<HTMLElement> | HTMLCollectionOf<HTMLFormElement> | NodeListOf<HTMLFormElement>;
   fields?: FilterCriterion<TrackedHTMLElement> & { transform: transformFn };
   events?: FormTrackingEvents;
 }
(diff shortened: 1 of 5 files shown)


snowplowcla:

plugins/browser-plugin-form-tracking/src/helpers.ts
```diff
@@ -160,14 +161,30 @@ export function addFormListeners(tracker: BrowserTracker, configuration: FormTra
  */
 function getConfigurationForOptions(options?: FormTrackingOptions) {
   if (options) {
+    var formFilter = (_: HTMLElement) => true;
+    var forms: HTMLCollectionOf<HTMLFormElement> | null = null;
+    if (
+      !options.forms ||
+      (options.forms as any).allowlist ||
+      (options.forms as any).denylist ||
+      (options.forms as any).filter
```
matus-tomlein, line 170: I don't love this solution to identify a `FilterCriterion` by duck typing, using the properties it might have (might lead to bugs in the future in case we add a property to the `FilterCriterion` type but forget to check for them here). But I can't think of a better solution, let me know if you know of one.

greg-el: The only thing I can think of is doing something like [this](https://www.npmjs.com/package/ts-transformer-keys), but adding in a package just for this seems a bit overkill (doubly so since it involves messing with the build). Perhaps add in a comment on the `FormTrackingOptions` interface and here as a reminder to update both? I think it's going to be less than ideal either way.

greg-el: Perhaps these `any`'s could be changed to `FilterCriterion<HTMLElement>` to be a bit more explicit?

greg-el (changes requested):
Looks good overall, a couple of comments, mostly pondering about `FilterCriterion`. The only concrete change is to remove `youtube.test.ts` (assuming it's managed to slip its way into this PR accidentally.)

trackers/javascript-tracker/test/integration/youtube.test.ts
```diff
@@ -148,7 +148,7 @@ describe('YouTube Tracker', () => {
 
     await browser.waitUntil(
       async () => {
-        return browser.call(async () => {
+        return await browser.call(async () => {
```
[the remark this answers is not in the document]
matus-tomlein: Yeah, it shouldn't.. It was a change I should have made in previous PR. Cleaned it up – rebased it into the previous PR.

plugins/browser-plugin-form-tracking/src/helpers.ts
matus-tomlein: Thanks Greg, I added the type cast to `FilterCriterion<HTMLElement>`

plugins/browser-plugin-form-tracking/src/helpers.ts
```diff
@@ -70,7 +70,7 @@ const defaultFormTrackingEvents = [
 ];
 
 export interface FormTrackingOptions {
-  forms?: FilterCriterion<HTMLElement>;
+  forms?: FilterCriterion<HTMLElement> | HTMLCollectionOf<HTMLElement>;
```
paulboocock, line 73: I think this could actually be:

```suggestion
  forms?: FilterCriterion<HTMLElement> | HTMLCollectionOf<HTMLFormElement> | NodeListOf<HTMLFormElement>;
```

You could get a node list with:

```js
let nodeList = document.querySelectorAll('.form')
```

where `.form` is a CSS class to find the forms.

plugins/browser-plugin-form-tracking/src/helpers.ts
```diff
@@ -160,14 +161,30 @@ export function addFormListeners(tracker: BrowserTracker, configuration: FormTra
  */
 function getConfigurationForOptions(options?: FormTrackingOptions) {
   if (options) {
+    var formFilter = (_: HTMLElement) => true;
+    var forms: HTMLCollectionOf<HTMLFormElement> | null = null;
```
paulboocock, line 165: We should use `let`.

paulboocock (commented):
This looks good, a nice feature!

There's a commit in this PR that shouldn't be there, looks like the max get bytes commit has made its way into this branch. Likely needs rebasing.

I think you could potentially get rid of the casting with something like this:

```
function isCollectionOfHTMLFormElements(
  forms?: FilterCriterion<HTMLElement> | HTMLCollectionOf<HTMLFormElement> | NodeListOf<HTMLFormElement>
): forms is HTMLCollectionOf<HTMLFormElement> | NodeListOf<HTMLFormElement> {
  return forms != null && Array.prototype.slice.call(forms).length > 0;
}

/*
 * Configures form tracking: which forms and fields will be tracked, and the context to attach
 */
function getConfigurationForOptions(options?: FormTrackingOptions) {
  if (options) {
    let formFilter = (_: HTMLElement) => true;
    let forms: HTMLCollectionOf<HTMLFormElement> | NodeListOf<HTMLFormElement> | undefined;
    if (isCollectionOfHTMLFormElements(options.forms)) {
      // options.forms is a collection of HTML form elements
      forms = options.forms;
    } else if (!options.forms || options.forms.allowlist || options.forms.denylist || options.forms.filter) {
      // options.forms is a filter
      formFilter = getFilterByClass(options.forms);
    }
    return {
      forms: forms,
      formFilter: formFilter,
      fieldFilter: getFilterByName<TrackedHTMLElement>(options.fields),
      fieldTransform: getTransform(options.fields),
      eventFilter: (event: FormTrackingEvent) => (options.events ?? defaultFormTrackingEvents).indexOf(event) > -1,
    };
  } else {
    return {
      formFilter: () => true,
      fieldFilter: () => true,
      fieldTransform: defaultTransformFn,
      eventFilter: () => true,
    };
  }
}
```

The above is untested, and perhaps could be tidied up or made more robust but hopefully gets the idea across.

matus-tomlein:
Great solution @paulboocock! Also really cool that you can define in TypeScript that a function checks if an argument is of a type, didn't know about that. It also makes further type casting unnecessary.

plugins/browser-plugin-form-tracking/src/helpers.ts
matus-tomlein: Makes sense, added the type.

PR #1071, opened 2022-03-25
Update ContextGenerator function return type to also accept SelfDescribingJson list and undefined (close #1070)

### LINKED ISSUE #1070: Update ContextGenerator function return type to also accept SelfDescribingJson list and undefined
**Describe the bug**
ContextGenerator functions can be passed to `addGlobalContext`. The type definition of `ContextGenerator` is as follows:

```ts
type ContextGenerator = (args?: ContextEvent) => SelfDescribingJson;
```

This constraint doesn't appear to reflect what's possible with Context Generators:

* The example in the [browser tracker documentation](https://docs.snowplowanalytics.com/docs/collecting-data/collecting-from-own-applications/javascript-trackers/browser-tracker/browser-tracker-v3-reference/tracking-events/#global-context) shows a context generator conditionally returning SelfDescribingJson
* It also appears to be possible to use a ContextGenerator to return a list of SelfDescribingJson

My suggestion is to update the type definition to be this:
```ts
type ContextGenerator = (args?: ContextEvent) => SelfDescribingJson | SelfDescribingJson[] | undefined;
```

**To Reproduce**

Using the example from then documentation should redline in an IDE using TypeScript.

```ts
import { addGlobalContexts } from '@snowplow/browser-tracker';
const contextGenerator = (args) => {
  if (args.eventType == 'pv') {
    return {
      schema: 'iglu:com.acme.marketing/some_event/jsonschema/1-0-0',
      data: { test: 1 },
    };
  }
};
addGlobalContexts([contextGenerator]);
```

**Expected behavior**
`addGlobalContexts` should accept ContextGenerator functions that return SelfDescribingJson, a SelfDescribingJson list, or `undefined`.

matus-tomlein:
Issue #1070

This PR fixes the return type from `ContextGenerator` functions to also include lists of `SelfDescribingJson` and undefined.

### DIFF
--- a/libraries/tracker-core/src/contexts.ts
+++ b/libraries/tracker-core/src/contexts.ts
@@ -54,7 +54,7 @@ export interface ContextEvent {
  * to allow an additional context to be dynamically attached to the event
  * @param args - - Object which contains the event information to help decide what should be included in the returned Context
  */
-export type ContextGenerator = (args?: ContextEvent) => SelfDescribingJson;
+export type ContextGenerator = (args?: ContextEvent) => SelfDescribingJson | SelfDescribingJson[] | undefined;
 
 /**
  * A context filter is a user-supplied callback that is evaluated for each event


paulboocock (approved):
LGTM

PR #1074, opened 2022-04-05
Fix integration tests

matus-tomlein:
Something changed in the last few days on Webdriver or on Saucelabs that made integration tests almost impossible to complete successfuly on CI. They became really flaky and unpredictable. I spent some time debugging them and trying out different changes and ended up with some that seem to have helped:

* I replaced the `browser.waitUntil` function with a custom one because the WDIO one was really unpredictable. When I use the WDIO one, the tests don't pass for me locally and are more flaky on Saucelabs. The custom implementation is a temporary fix, we can try at some later point if the original works again.
* I changed all `browser.call` callbacks that both fetched the Micro results and returned some other value to do only one of those things. Then I added additional `browser.call` to fetch the results. I don't have anything to support this but I think it helped a bit and it seems cleaner to me.
* I updated the version of webdriver to 7.19 as they had a fix from a couple of days ago to a problem that I was seeing too – "maximum call stack size exceeded"
* Updated the configuration for webdriver:
   * Added a delay for 1 minute between retries (I often saw that retries for YouTube tests ran back to back and failed while they could pass if tried later)
   * Increased the number of retries from 2 to 3 (due to test flakiness)
   * Set to defer retries to the end of queue (not sure if this helps much, maybe the 1 minute delay between retries was enough)

I won't merge this PR directly into the release branch but will integrate the changes into the "Bump dependencies" commit already there – that is doing similar changes and I don't want to have two commits in the release with the same stuff.


snowplowcla:

trackers/javascript-tracker/test/integration/youtube.test.ts
```diff
@@ -137,24 +135,25 @@ describe('YouTube Tracker', () => {
...
-    await browser.waitUntil(() => $('#youtube').isExisting(), {
+    await waitUntil(browser, () => $('#youtube').isExisting(), {
       timeout: 5000,
       timeoutMsg: 'expected youtube after 5s',
     });
 
     const player = $('#youtube');
     await player.click(); // emits 'playbackqualitychange' and 'play';
-    await player.keys(Array(3).fill('ArrowRight')); // Skips to the point just before 'percentprogress' fires
+    await player.keys(Array(2).fill('ArrowRight')); // Skips to the point just before 'percentprogress' fires
```
greg-el, line 145: It's interesting that this needed to change, though I'm sure I did something similar in the past to get tests working after they seemingly started failing out of nowhere.

greg-el (approved):
LGTM, nice that you could lower some interval/timeout times for the tests too! (anything to stave off the inevitable 30-minute run time)

trackers/javascript-tracker/test/wdio.default.conf.ts
```diff
@@ -5,7 +5,10 @@ export const config = {
   waitforTimeout: 30000,
   connectionRetryTimeout: 120000,
   connectionRetryCount: 3,
-  specFileRetries: 1,
+  specFileRetries: 2,
+  specFileRetriesDelay: 60, // Delay in seconds between the spec file retry attempts
+  specFileRetriesDeferred: false, // Defer retries to the end of the queue
+  maxInstances: 5, // Maximum number of total parallel running workers
```
paulboocock, line 11: Nice, I'd hope these helps a little bit with sporadic failures.

paulboocock (approved):
LGTM

matus-tomlein:
Thanks Greg and Paul for the reviews! Integrated into the release branch.

PR #1078, opened 2022-06-02
Add client session context entity (close #1077)

### LINKED ISSUE #1077: Add client session context entity
Add the client session context entity version 1-0-2 with the following properties:

* `userId` – An identifier for the user of the session (same as `domain_userid`)
* `sessionId` – An identifier for the session (same as `domain_sessionid`)
* `sessionIndex` – The index of the current session for this user (same as `domain_sessionidx`)
* `eventIndex` – Index of the current event in the session (new property)
* `previousSessionId` – The previous session identifier for this user (new property)
* `storageMechanism` – The mechanism that the session information has been stored on the device (new property)
* `firstEventId` – Identifier of the first event for this session (new property)
* `firstEventTimestamp` – Date-time timestamp of when the first event in the session was tracked (new property)

matus-tomlein:
This PR addresses issue #1077 and adds the `client_session` ([schema version 1.0.2](http://iglucentral.com/schemas/com.snowplowanalytics.snowplow/client_session/jsonschema/1-0-2)) context entity to the Web trackers. This entity contains session information.

The context entity repeats some of the session information stored in canonical event properties (`domain_userid`, `domain_sessionid`, `domain_sessionidx`), but also adds new information. It adds a reference to the previous session (`previousSessionId`) and first event in the current session (`firstEventId`, `firstEventTimestamp`). It also adds an index of the event in the session useful for ordering events as they were tracked (`eventIndex`).

## Cookie format

The new session information is stored in the ID cookie (`sp_id`). It is appended to the end of the cookie. There are 4 new properties stored in the cookie, which now has the format:

`{domainUserId}.{createdTime}.{visitCount}.{nowTime}.{lastVisitTime}.{sessionId}.{previousSessionId}.{firstEventId}.{firstEventTsInMs}.{eventIndex}`

The new properties are:

- `previousSessionId` – ID of the previous session
- `firstEventId` – ID of the first event in the current session
- `firstEventTsInMs` Device created timestamp of the first event in the current session
- `eventIndex` – Index of the last event in the session

## Configuration

The session context entity is disabled by default. It can be enabled using the `contexts.session` property:

```js
window.snowplow('newTracker', 'sp1', 'https://...', {
    ...,
    contexts: {
        session: true
    }
});
```

Anonymous tracking has to be disabled for the session context entity to be added to events.

## Changes in the browser-tracker-core

I did some refactoring of the code handling the ID cookies. I extracted the functions from `tracker/index.ts` to `tracker/id_cookie.ts` and added unit tests for them. Also I added types for the parsed ID cookie array, which is now a TypeScript tuple. New sessions were started in a few different places so I unified the logic into a single function (`startNewIdCookieSession`).

The session information is updated and added to events in the `beforeTrack()` of `BrowserDataPlugin`. I modified the function to operate on the ID cookie tuple directly.

If the `contexts.session` configuration is enabled, `beforeTrack()` adds the context entity to the `PayloadBuilder` (see below). All the new information is tracked in the cookie (previous session ID, first event, event index) even if the configuration is disabled, but it is not used.

## Changes in PayloadBuilder

I needed a way to add the context entity to the `PayloadBuilder` object in `beforeTrack`. In order to avoid parsing the `cx` property and serializing the context again in the `beforeTrack`, I added a new helper function to the `PayloadBuilder` – `addContextEntity(entity: SelfDescribingJson)`.

This function stores the passed object in a list and adds it to the payload dictionary in the `build()` method along with other JSON objects. It tries to combine the new context entities with existing entities added through `addJson()` or already stored in the payload dictionary. This adds more complexity to it, but is safer as we don't know how the plugins might assign new context entities. It also enables `build()` to be called multiple times.

## Documentation

PR with changes in the documentation is open [in the data-value-resources repo.](https://github.com/snowplow-incubator/data-value-resources/pull/109)

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/id_cookie.ts
+++ b/libraries/browser-tracker-core/src/tracker/id_cookie.ts
@@ -1,0 +1,356 @@
+/*
+ * Copyright (c) 2022 Snowplow Analytics Ltd, 2010 Anthon Pang
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright notice,
+ *    this list of conditions and the following disclaimer in the documentation
+ *    and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the copyright holder nor the names of its
+ *    contributors may be used to endorse or promote products derived from
+ *    this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+ * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { PayloadBuilder } from '@snowplow/tracker-core';
+import { v4 as uuid } from 'uuid';
+
+/**
+ * Indices of cookie values
+ */
+const cookieDisabledIndex = 0,
+  domainUserIdIndex = 1,
+  createTsIndex = 2,
+  visitCountIndex = 3,
+  nowTsIndex = 4,
+  lastVisitTsIndex = 5,
+  sessionIdIndex = 6,
+  previousSessionIdIndex = 7,
+  firstEventIdIndex = 8,
+  firstEventTsInMsIndex = 9,
+  eventIndexIndex = 10;
+
+export type ParsedIdCookie = [
+  string, // cookieDisabled
+  string, // domainUserId
+  number, // cookieCreateTs
+  number, // visitCount
+  number, // nowTs
+  string | number, // lastVisitTs
+  string, // sessionId
+  string, // previousSessionId
+  string, // firstEventId
+  string | number, // firstEventTs
+  number // eventIndex
+];
+
+/**
+ * Schema for client client session context entity
+ */
+export interface ClientSession extends Record<string, unknown> {
+  /**
+   * An identifier for the user of the session (same as domain_userid)
+   */
+  userId: string;
+
+  /**
+   * An identifier for the session (same as domain_sessionid)
+   */
+  sessionId: string;
+
+  /**
+   * The index of the current session for this user (same as domain_sessionidx)
+   */
+  sessionIndex: number;
+
+  /**
+   * Index of the current event in the session
+   */
+  eventIndex: number;
+
+  /**
+   * The previous session identifier for this user
+   */
+  previousSessionId: string | null;
+
+  /**
+   * The mechanism that the session information has been stored on the device
+   */
+  storageMechanism: string;
+
+  /**
+   * Identifier of the first event for this session
+   */
+  firstEventId: string | null;
+
+  /**
+   * Date-time timestamp of when the first event in the session was tracked
+   */
+  firstEventTimestamp: string | null;
+}
+
+export function emptyIdCookie() {
+  const idCookie: ParsedIdCookie = ['1', '', 0, 0, 0, '', '', '', '', '', 0];
+  return idCookie;
+}
+
+/**
+ * Parses the cookie values from its string representation.
+ *
+ * @param id Cookie value as string
+ * @param domainUserId Domain user ID to be used in case of empty cookie string
+ * @returns Parsed ID cookie tuple
+ */
+export function parseIdCookie(
+  id: string,
+  domainUserId: string,
+  memorizedSessionId: string,
+  memorizedVisitCount: number
+) {
+  let now = new Date(),
+    nowTs = Math.round(now.getTime() / 1000),
+    tmpContainer;
+
+  if (id) {
+    tmpContainer = id.split('.');
+    // cookies enabled
+    tmpContainer.unshift('0');
+  } else {
+    tmpContainer = [
+      // cookies disabled
+      '1',
+      // Domain user ID
+      domainUserId,
+      // Creation timestamp - seconds since Unix epoch
+      nowTs,
+      // visitCount - 0 = no previous visit
+      memorizedVisitCount,
+      // Current visit timestamp
+      nowTs,
+      // Last visit timestamp - blank meaning no previous visit
+      '',
+      // Session ID
+      memorizedSessionId,
+    ];
+  }
+
+  if (!tmpContainer[sessionIdIndex] || tmpContainer[sessionIdIndex] === 'undefined') {
+    // session id
+    tmpContainer[sessionIdIndex] = uuid();
+  }
+  if (!tmpContainer[previousSessionIdIndex] || tmpContainer[previousSessionIdIndex] === 'undefined') {
+    // previous session id
+    tmpContainer[previousSessionIdIndex] = '';
+  }
+  if (!tmpContainer[firstEventIdIndex] || tmpContainer[firstEventIdIndex] === 'undefined') {
+    // firstEventId - blank meaning no previous event
+    tmpContainer[firstEventIdIndex] = '';
+  }
+  if (!tmpContainer[firstEventTsInMsIndex] || tmpContainer[firstEventTsInMsIndex] === 'undefined') {
+    // firstEventTs - blank meaning no previous event
+    tmpContainer[firstEventTsInMsIndex] = '';
+  }
+  if (!tmpContainer[eventIndexIndex] || tmpContainer[eventIndexIndex] === 'undefined') {
+    // eventIndex – 0 = no previous event
+    tmpContainer[eventIndexIndex] = 0;
+  }
+
+  const parsed: ParsedIdCookie = [
+    tmpContainer[0] as string,
+    tmpContainer[1] as string,
+    parseInt(tmpContainer[2] as string),
+    parseInt(tmpContainer[3] as string),
+    parseInt(tmpContainer[4] as string),
+    tmpContainer[5] ? parseInt(tmpContainer[5] as string) : (tmpContainer[5] as string),
+    tmpContainer[6] as string,
+    tmpContainer[7] as string,
+    tmpContainer[8] as string,
+    tmpContainer[9] ? parseInt(tmpContainer[9] as string) : (tmpContainer[9] as string),
+    parseInt(tmpContainer[10] as string),
+  ];
+  return parsed;
+}
+
+/**
+ * Initializes the domain user ID if not already present in the cookie. Sets an empty string if anonymous tracking.
+ *
+ * @param idCookie Parsed cookie
+ * @param configAnonymousTracking Whether anonymous tracking is enabled
+ * @returns Domain user ID
+ */
+export function initializeDomainUserId(idCookie: ParsedIdCookie, configAnonymousTracking: boolean) {
+  let domainUserId;
+  if (idCookie[domainUserIdIndex]) {
+    domainUserId = idCookie[domainUserIdIndex];
+  } else if (!configAnonymousTracking) {
+    domainUserId = uuid();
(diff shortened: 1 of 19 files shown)


snowplowcla:

libraries/browser-tracker-core/src/tracker/id_cookie.ts
```diff
@@ -0,0 +1,356 @@
...
+    tmpContainer[1] as string,
+    parseInt(tmpContainer[2] as string),
+    parseInt(tmpContainer[3] as string),
+    parseInt(tmpContainer[4] as string),
+    tmpContainer[5] ? parseInt(tmpContainer[5] as string) : (tmpContainer[5] as string),
+    tmpContainer[6] as string,
+    tmpContainer[7] as string,
+    tmpContainer[8] as string,
+    tmpContainer[9] ? parseInt(tmpContainer[9] as string) : (tmpContainer[9] as string),
+    parseInt(tmpContainer[10] as string),
```
paulboocock, line 185: We should check to ensure we get a number here. Malformed cookies could lead to unexpected results.

libraries/browser-tracker-core/src/tracker/id_cookie.ts
```diff
@@ -0,0 +1,356 @@
...
+ */
+export function startNewIdCookieSession(idCookie: ParsedIdCookie, memorizedVisitCount: number = 1) {
+  // If cookies are enabled, base visit count and session ID on the cookies
+  if (cookiesEnabledInIdCookie(idCookie)) {
+    // Store the previous session ID
+    idCookie[previousSessionIdIndex] = idCookie[sessionIdIndex];
+    // Set lastVisitTs to currentVisitTs
+    idCookie[lastVisitTsIndex] = idCookie[nowTsIndex];
+    // Increment the session ID
+    (idCookie[visitCountIndex] as number)++;
```
paulboocock, line 228: Isn't this already a number?

```suggestion
    idCookie[visitCountIndex]++;
```

libraries/browser-tracker-core/src/tracker/id_cookie.ts
```diff
@@ -0,0 +1,356 @@
...
+  }
+
+  // Create a new sessionId
+  const sessionId = uuid();
+  idCookie[sessionIdIndex] = sessionId;
+
+  // Reset event index and first event references
+  idCookie[eventIndexIndex] = 0;
+  idCookie[firstEventIdIndex] = '';
+  idCookie[firstEventTsInMsIndex] = '';
```
paulboocock, line 240: Could/should these be `undefined` rather than an empty string? This is probably a bit nitpicky, but I'm not a huge fan of `idCookie[firstEventTsInMsIndex]` being a number or a string. I'm more comfortable with it being a number or undefined though.

libraries/tracker-core/src/payload.ts
```diff
@@ -28,7 +28,8 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import { base64urlencode } from './base64';
+import { base64urlencode, base64urldecode } from './base64';
+import type { SelfDescribingJson } from './core';
```
paulboocock, line 32: I think this is introduced in TypeScript 3.8? It seems unlikely to cause any issues, but might be worth ensuring this doesn't break anything for users running earlier versions of typescript. I don't think it will, as its just an inport for this file, but I'm always nervous about this, it's way too easy to do (and I've definitely done it before in an earlier release by making some API function have some fancy TS syntax)

paulboocock (commented):
This is looking good. I haven't quite finished yet but I wanted to get my initial run through over to you. Nothing major here though.

I want to spend a little more time picking through the changes to `payloadJsonProcessor` and to run it through some quick manual tests, just to see if I can catch it out at all. I might struggle to wrap this up tomorrow, but will certainly wrap it up on Wednesday morning.

libraries/tracker-core/src/payload.ts
```diff
@@ -151,16 +168,61 @@ export function payloadBuilder(): PayloadBuilder {
...
+      let payload = payloadBuilder.getPayload();
+      if (encodeBase64 ? payload.cx : payload.co) {
+        return JSON.parse(encodeBase64 ? base64urldecode(payload.cx as string) : (payload.co as string));
+      }
+      return undefined;
+    };
+
+    for (const json of jsonForProcessing) {
+      if (json.keyIfEncoded == 'cx') {
+        context = getContextFromPayload();
```
matus-tomlein, line 196: Just making a note for myself that this is a bug – overwriting the `context` variable in a for loop. Need to add a test for calling `addContextEntity` multiple times after each other.

paulboocock (approved):
I'm approving this, pending the outstanding suggestions. It certainly solves the problem well, just a few nits knocking about. Happy to give it a quick once over once the changes have been made too.

libraries/browser-tracker-core/src/tracker/id_cookie.ts
```diff
@@ -0,0 +1,356 @@
...
+    parseInt(tmpContainer[2] as string),
+    parseInt(tmpContainer[3] as string),
+    parseInt(tmpContainer[4] as string),
+    tmpContainer[5] ? parseInt(tmpContainer[5] as string) : (tmpContainer[5] as string),
+    tmpContainer[6] as string,
+    tmpContainer[7] as string,
+    tmpContainer[8] as string,
+    tmpContainer[9] ? parseInt(tmpContainer[9] as string) : (tmpContainer[9] as string),
+    parseInt(tmpContainer[10] as string),
+  ];
```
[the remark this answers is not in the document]
matus-tomlein: I thought that it's not necessary here because we just go from 0 to 10 in order, but on the other hand, it doesn't hurt to use the index constants. Changed it!

libraries/browser-tracker-core/src/tracker/id_cookie.ts
```diff
@@ -0,0 +1,356 @@
...
+    tmpContainer[eventIndexIndex] = 0;
+  }
+
+  const parsed: ParsedIdCookie = [
+    tmpContainer[0] as string,
+    tmpContainer[1] as string,
+    parseInt(tmpContainer[2] as string),
+    parseInt(tmpContainer[3] as string),
+    parseInt(tmpContainer[4] as string),
+    tmpContainer[5] ? parseInt(tmpContainer[5] as string) : (tmpContainer[5] as string),
```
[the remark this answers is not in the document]
matus-tomlein: The problem here is that the two timestamps are optional – there can either be an integer value with the time or it can be blank (empty string when serialized to cookie). I didn't want to default to nowTs or 0 because that would change blank behaviour. I cleaned this up a bit and changed the empty string to undefined, hope that's better.

libraries/browser-tracker-core/src/tracker/id_cookie.ts
matus-tomlein: Good point, fixed!

libraries/browser-tracker-core/src/tracker/id_cookie.ts
matus-tomlein: Changed it all to undefined. I was just worried that it would cause trouble when serializing back to the cookie string, but found out that the `Array.join()` function handles undefined values well – turns them into empty string.

libraries/browser-tracker-core/src/tracker/id_cookie.ts
```diff
@@ -0,0 +1,356 @@
...
+ *
+ * @param idCookie Parsed cookie
+ * @param payloadBuilder Event payload builder
+ */
+export function updateFirstEventInIdCookie(idCookie: ParsedIdCookie, payloadBuilder: PayloadBuilder) {
+  // Update first event references if new session or not present
+  if (idCookie[eventIndexIndex] == 0) {
+    const payload = payloadBuilder.build();
+    idCookie[firstEventIdIndex] = payload['eid'] as string;
+    const ts = (payload['dtm'] || payload['ttm']) as string;
```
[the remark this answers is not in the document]
matus-tomlein: Actually, when the user sets the `ttm` true timestamp, device timestamp is not set (if I'm reading it right): https://github.com/snowplow/snowplow-javascript-tracker/blob/4dec01b9a314ca11d88372afa4b2dc67a431ee0d/libraries/tracker-core/src/core.ts#L113-L115

Not sure which one to prefer though, but in the end it's the same result as only one timestamp is present.

libraries/browser-tracker-core/src/tracker/id_cookie.ts
```diff
@@ -0,0 +1,356 @@
...
+
+/**
+ * Updates the first event references according to the event payload if first event in session.
+ *
+ * @param idCookie Parsed cookie
+ * @param payloadBuilder Event payload builder
+ */
+export function updateFirstEventInIdCookie(idCookie: ParsedIdCookie, payloadBuilder: PayloadBuilder) {
+  // Update first event references if new session or not present
+  if (idCookie[eventIndexIndex] == 0) {
```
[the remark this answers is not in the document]
matus-tomlein: Fixed!

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -801,30 +758,30 @@ export function Tracker(
...
+          updateNowTsInIdCookie(idCookie);
+          updateFirstEventInIdCookie(idCookie, payloadBuilder);
+          incrementEventIndexInIdCookie(idCookie);
+
           payloadBuilder.add('vp', detectViewport());
           payloadBuilder.add('ds', detectDocumentSize());
           payloadBuilder.add('vid', anonymizeSessionOr(memorizedVisitCount));
           payloadBuilder.add('sid', anonymizeSessionOr(memorizedSessionId));
-          payloadBuilder.add('duid', anonymizeOr(_domainUserId)); // Set to our local variable
+          payloadBuilder.add('duid', anonymizeOr(domainUserIdFromIdCookie(idCookie))); // Set to our local variable
```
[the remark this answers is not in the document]
matus-tomlein: Fixed the comment!

libraries/tracker-core/src/payload.ts
matus-tomlein: It was introduced in TS 3.8, but I wasn't able to find out by googling how it is handled in previous versions. But I can't import from the core file without the `import type` directive as that would be a circular dependency (core also imports this file). So don't see how else to solve it (other than refactoring the files and extracting `SelfDescribingJson` to another file.

libraries/tracker-core/src/payload.ts
```diff
@@ -151,16 +168,61 @@ export function payloadBuilder(): PayloadBuilder {
  * @returns The request builder, with add and build methods
  */
 export function payloadJsonProcessor(encodeBase64: boolean): JsonProcessor {
-  return (payloadBuilder: PayloadBuilder, jsonForProcessing: EventJson) => {
-    for (const json of jsonForProcessing) {
-      const str = JSON.stringify(json.json);
+  return (
```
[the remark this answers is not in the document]
matus-tomlein: I'm also worried about the added complexity especially since the `addJson` function can be used by external plugins and who knows how they use it...

I tried to clean up the code a bit and also remove the `any` usage where possible. Please take another look when you get a chance.

matus-tomlein:
Thanks for the review @paulboocock, really useful! If you have a chance to take another look at it, I'd appreciate it.

libraries/tracker-core/src/payload.ts
paulboocock: This should be fine, I'm just nervous we'll break the TS definitions as part of the build output for earlier TS versions. It seems unlikely this import will make it into the final output though.

I think the quick check is to build the tracker package, load it via file reference into a react app with TS 3.7, import and initialise the tracker, and see if it compiles and sends an event.

libraries/browser-tracker-core/src/tracker/id_cookie.ts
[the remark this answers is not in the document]
paulboocock: Everydays a learning day. This should be fine then :)

libraries/browser-tracker-core/src/tracker/id_cookie.ts
```diff
@@ -0,0 +1,361 @@
...
+  if (!tmpContainer[eventIndexIndex] || tmpContainer[eventIndexIndex] === 'undefined') {
+    // eventIndex – 0 = no previous event
+    tmpContainer[eventIndexIndex] = 0;
+  }
+
+  const parseIntOr = (value: any, defaultValue: any) => {
+    let parsed = parseInt(value as string);
+    return isNaN(parsed) ? defaultValue : parsed;
+  };
+  const parseIntOrUndefined = (value: any) => (value ? parseIntOr(value, undefined) : undefined);
```
paulboocock, line 178: Nice 👌

libraries/browser-tracker-core/src/tracker/id_cookie.ts
```diff
@@ -0,0 +1,361 @@
...
+  if (!tmpContainer[firstEventTsInMsIndex] || tmpContainer[firstEventTsInMsIndex] === 'undefined') {
+    // firstEventTs - blank meaning no previous event
+    tmpContainer[firstEventTsInMsIndex] = '';
+  }
+  if (!tmpContainer[eventIndexIndex] || tmpContainer[eventIndexIndex] === 'undefined') {
+    // eventIndex – 0 = no previous event
+    tmpContainer[eventIndexIndex] = 0;
+  }
+
+  const parseIntOr = (value: any, defaultValue: any) => {
```
paulboocock, line 174: nit: I'd usually use `unknown` here rather than `any`.

paulboocock (approved):
LGTM

libraries/browser-tracker-core/src/tracker/id_cookie.ts
matus-tomlein: It actually breaks the type checking in this case, because if I change `defaultValue` to unknown, TypeScript won't automatically infer its type based on what is passed and the `parseIntOr` will always have the return type `unknown`. Instead if I leave it as any, the implicit return type of the function is inferred from the arguments I pass to it. At least that is my understanding based on the type check errors that I am getting in VS Code.

Reading [this Stackoverflow](https://stackoverflow.com/a/51439876/3334072) maybe a better explanation is that you can't assign unknown to other types (e.g., number). So I would always have to type cast the result of `parseIntOr` on the lines below where I use it.

PR #1080, opened 2022-06-03
Add a customizable set of failure HTTP status codes for which collector requests should not be retried (close #1079)

### LINKED ISSUE #1079: Add a customizable set of failure HTTP status codes for which collector requests should not be retried
Instead of retrying collector requests for all non-2xx status codes, disable the retry functionality for the following HTTP status codes by default:

Code | Description | Rationale
-- | -- | --
400 | Bad Request | Server rejected to process the request due to an issue with it, trying again won't help.
401 | Unauthorized (RFC 7235) | Request is lacking correct authorization or refusing to process some IP address – intentional denial of request. Similar to 403 but specific to authorization issues.
403 | Forbidden | Server is refusing to take action, request should not be repeated.
410 | Gone | Indicates an intentional denial of the request as the resource was removed, request should not be retried and potentially the tracker should stop emitting events.
422 | Unprocessable Entity (WebDAV; RFC 4918) | Server was unable to process events in the request, should not be repeated without modification of the request.

All other non-2xx status codes should be retried on.

The list of HTTP status codes for which to retry or not-retry should be user customizable (e.g., the user should be able to force retries on 403 status code).

matus-tomlein:
This PR addresses issue #1079.

It adds a set of 5 HTTP status codes (400, 401, 403, 410, and 422) for which sending events should not be retried and the events should be dropped. This is in line with the strategy for handling HTTP status codes that we [outlined here](https://github.com/snowplow-incubator/data-value-resources/blob/main/30%20Tracker%20Architecture/emitter_architecture/handling_http_status_codes.md).

The behaviour for 3xx status codes has also changed – previously they were considered as successful and not retried. This doesn't fit with how we decided to handle them, so I changed them to be retried.

It also makes the retry behaviour configurable. It adds two configuration options: `retryStatusCodes` and `dontRetryStatusCodes` which are lists of non-successful status codes (>= 300) that override the default behaviour.

## Documentation: Setting custom retry HTTP status codes

The tracker provides a retry functionality that sends the same events repeatedly in case GET or POST requests to the Collector fail. This may happen due to connection issues or a non-successful HTTP status code in Collector response.

Prior to version 3.5 of the tracker, requests receiving all 4xx and 5xx HTTP status codes in Collector response were retried. Since version 3.5, the behaviour changed and became customizable:

By default, the tracker retries on all 3xx, 4xx, and 5xx status codes except for 400, 401, 403, 410, and 422. The set of status codes for which events should be retried or not is customizable. You can make use of the `retryStatusCodes` and `dontRetryStatusCodes` lists to specify them. Retry behaviour can only be configured for non-successful status codes (i.e., >= 300).

```js
retryStatusCodes: [403], // override default behavior and retry on 403
dontRetryStatusCodes: [418] // force retry on 418
```

Please note that not retrying sending events to the Collector means that the events will be dropped when they fail to be sent. Take caution when choosing the `dontRetryStatusCodes`.

PR for documentation is in the [data-value-resources repo here](https://github.com/snowplow-incubator/data-value-resources/pull/110).

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/out_queue.ts
+++ b/libraries/browser-tracker-core/src/tracker/out_queue.ts
@@ -60,6 +60,8 @@ export interface OutQueue {
  * @param anonymousTracking - Defines whether to set the SP-Anonymous header for anonymous tracking on GET and POST
  * @param customHeaders - Allows custom headers to be defined and passed on XMLHttpRequest requests
  * @param withCredentials - Sets the value of the withCredentials flag on XMLHttpRequest (GET and POST) requests
+ * @param retryStatusCodes – Failure HTTP response status codes from Collector for which sending events should be retried
+ * @param dontRetryStatusCodes – Failure HTTP response status codes from Collector for which sending events should not be retried
  * @returns object OutQueueManager instance
  */
 export function OutQueueManager(
@@ -76,7 +78,9 @@ export function OutQueueManager(
   connectionTimeout: number,
   anonymousTracking: boolean,
   customHeaders: Record<string, string>,
-  withCredentials: boolean
+  withCredentials: boolean,
+  retryStatusCodes: number[],
+  dontRetryStatusCodes: number[]
 ): OutQueue {
   type PostEvent = {
     evt: Record<string, unknown>;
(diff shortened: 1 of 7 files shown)


snowplowcla:

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -333,25 +337,34 @@ export function OutQueueManager(
...
           clearTimeout(xhrTimeout);
-          onPostSuccess(numberToSend);
-        } else if (xhr.readyState === 4 && xhr.status >= 400) {
-          clearTimeout(xhrTimeout);
-          executingQueue = false;
+          if (xhr.status < 300) {
+            onPostSuccess(numberToSend);
+          } else {
+            if (!shouldRetryForStatusCode(xhr.status)) {
+              LOG.error(`Collector responded with status code ${xhr.status}, sending events won't be retried.`);
```
paulboocock, line 363: I'm always trying to shorten strings where possible, always chasing that filesize goal, maybe this could just be `Status ${xhr.status}, will not retry`. All logs are prepended with `Snowplow:` so that should be obvious.

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -422,6 +435,22 @@ export function OutQueueManager(
...
+      return false;
+    }
+
+    // retry if status code among custom user-supplied retry codes
+    if (retryStatusCodes.includes(statusCode)) {
+      return true;
+    }
+
+    // retry if status code *not* among custom user-supplied don't retry codes and default don't retry codes
+    const dontRetryStatusCodesDefault = [400, 401, 403, 410, 422];
```
paulboocock, line 450: We're redefining this constant on every call to this function. Could we hoist it out of the function? 

Maybe even do something fancy near the config params defaults bit in `libraries/browser-tracker-core/src/tracker/index.ts`?

paulboocock (approved):
Looks good! Couple of nits.

paulboocock (approved):

PR #1088, opened 2022-08-24
Add tab identifier on web_page context. Fix #815

### LINKED ISSUE #815: Add a tab identifying context
Currently it is possible for a user to have two tabs open at the same time which are both generating events over a single page view. Whilst they and subsequent events would have different `pageview_id`s, it would be great to easily split these two page view events via a `tab_id` to understand further user behaviour.

igneel64:
## Add tab identifier in the  web_page context

As discussed in #815 it would be useful to add a tab identifier to further the analysis capabilities.

We will enrich the webpage schema with a tabId attribute which aims to identify the tab the user was in when he navigated the site. (_PR incoming for the iglu repository_)

The PR also adds the `tracker.getTabId()` method on the public API.

### Technical implementation:

The tab identifier can be implemented using the [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) API that provides access to a ‘session’ type persistence on the current tab. 

By using the API we should understand the benefits/limitations it provides:
- Persistence is valid for the same [origin](https://developer.mozilla.org/en-US/docs/Glossary/Origin) by default. Meaning different origins will have different tab identifiers even if they were navigated from the same tab.
- If the tab is closed, the identifier will be lost.
- If the user continues his navigation on different origins  and comes back again on the origin the identifier was initialised, on the same tab, the tab identifier will be the same.
- If the user refreshes the page, the tab identifier will be the same.

### Compatibility:
The compatibility is proper for our trackers based on [caniuse](https://caniuse.com/mdn-api_window_sessionstorage).

### Multi-tracker behaviour:
The  tabId attribute will be the same for the different Snowplow trackers that might be initialised on the page.

### Anonymous tracking:
This value should not be able to be used to identify anonymous users by stitching the tab identifier with a later non-anonymous usage of the tracker. On anonymous and stateless behaviour, a random uuid is returned every time.
E.g. I browse the site with anonymous tracking enabled. After some time I enable tracking to function properly. If I am on the same tab, the analyst can stitch the tab identifier and correlate the anonymous with the recognised user.

### Storage behaviour:
This value depends on user allowing storage in the [storage strategy](https://docs.snowplowanalytics.com/docs/collecting-data/collecting-from-own-applications/javascript-trackers/javascript-tracker/javascript-tracker-v3/tracker-setup/initialization-options/#Configuring_the_storage_strategy).
Additionally it can conceptually be classified in the same strategy as “local[Storage](https://developer.mozilla.org/en-US/docs/Web/API/Storage)“ since both these structures are implementing the same Web Storage API and mechanisms.

### Documentation
Will discuss what needs to change on the [webPage](https://docs.snowplowanalytics.com/docs/collecting-data/collecting-from-own-applications/javascript-trackers/browser-tracker/browser-tracker-v3-reference/tracker-setup/initialization-options/) context section.

#### Notes
Additional minor fixes: 
- A few typos.
- Split the storage helpers to their own file.
- Add the correct return type for `tracker.getPageViewId()`


snowplowcla:

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -726,6 +746,7 @@ export function Tracker(
               schema: 'iglu:com.snowplowanalytics.snowplow/web_page/jsonschema/1-0-0',
```
matus-tomlein, line 746: I think the schema version should be `1-1-0` here.

trackers/browser-tracker/test/tab_id_storage.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    tracker.disableAnonymousTracking();
+    tracker.trackPageView();
+    const storedTabId = window.sessionStorage.getItem('_sp_tab_id');
+    const tabId = tracker.getTabId();
+    expect(storedTabId).not.toBe(undefined);
+    expect(tabId).toEqual(storedTabId);
+  });
+
+  afterAll(() => {
```
matus-tomlein, line 98: I wouldn't expect to find this at the bottom of the file :) Usually we put them at the top along with the before callbacks. This approach might be better but I think it's more important to be consistent within the project.

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -713,6 +716,23 @@ export function Tracker(
       return state.pageViewId;
     }
 
+    /**
+     * Safe function to get `tabId`.
+     * Generates it if it is not yet initialized. Shared between trackers.
+     */
+    function getTabId() {
```
matus-tomlein, line 723: I'm wondering what should happen in case the `stateStorageStrategy` is `none` (and anonymous tracking is off) – would it make sense to use the in-memory `SharedState` storage as we do for the `pageViewId`? It complicates the logic to store the value in multiple places so I'm not sure if it's worth it. But I think it'd make sense because otherwise, the `pageViewId` (also the `domainUserId` since it is also cached in memory) would be consistent for some events tracked with the same in-memory tracker instance, but the `tabId` would be changing on every event – that could be quite confusing for users.

matus-tomlein (commented):
Awesome, lots of nice refactoring of the code! 👍 Just added a couple comments from a quick pass through.

Also make sure to run `rush change` to generate changelog items as the PR check is failing because of that. The build action is probably just failing due to flaky integration tests...

trackers/browser-tracker/test/tab_id_storage.test.ts
```diff
@@ -0,0 +1,102 @@
+/*
```
matus-tomlein, line 1: I'm a bit confused about the placement of this file, is it intentional that it is in `trackers/browser-tracker`? The feature is implemented in `libraries/browser-tracker-core` so I would expect the tests to be there, but maybe I'm missing some requirement here.

libraries/browser-tracker-core/src/tracker/index.ts
igneel64: Yeap missed that, thanks!

trackers/browser-tracker/test/tab_id_storage.test.ts
igneel64: No particular reason, just the size of the file made it flow to the eye easily. Will move it on top as I agree!

trackers/browser-tracker/test/tab_id_storage.test.ts
igneel64: Actually this took a bit of my time to think about too. This was my train of thought that resulted in my adding it there:
- The tests on `libraries/browser-tracker-core` are about internal implementations of the tracker system. _(E.g. `newTracker|addTracker|Tracker` is never used in these tests.)_
- The tests on `trackers/browser-tracker` had already tests that described high level behaviour.
Based on the above points and the thinking that this is a high level feature to begin with, I added the tests there.
WDYT ?

libraries/browser-tracker-core/src/tracker/index.ts
igneel64: _Went over this as well as we discussed offline._
I totally see the reasoning you stated above. 
What I am thinking is that storing the `tabId` in memory will make it more **consistent** but will not make it more **reliable** if storage is not allowed. This attribute, in my mind at least, is a tightly coupled notion to the browser storage and persistence of the same tab. 
That is the tradeoff I am thinking about for adding it in the `SharedState` and the logic that would be required.

matus-tomlein: That's a good point! I can see that storing it in memory gives an illusion of reliability in the data. On the other hand, we made the trade-off in other places to store other identifiers in memory and I'm also nervous about having different lifespans for different identifiers. Also there might be some value in the in-memory `tabId` for some client-side JS apps (or maybe not, I don't really know what use cases it addresses).

Remind me please again why we decided to generate a new UUID for the `tabId` even if we can't persist it instead of returning `null`? I know we talked about this on the call, but already forgot the reasoning. Returning `null` might be a way to avoid setting the user expectations for the same lifespan compared to `pageViewId` for instance when storage is `none`.

trackers/browser-tracker/test/tab_id_storage.test.ts
matus-tomlein: So these are kind of like in between unit and integration tests? I'm fine with that and it makes sense to also test on this level, but it leaves the `trackers/javascript-tracker` without testing the functionality. We can keep it like this, just something about it is bugging me that we should find a way to distribute the test coverage to multiple levels.

igneel64: Will discuss it internally as soon as we get the chance and we will come up with the correct strategy.

libraries/browser-tracker-core/src/tracker/index.ts
igneel64: From the offline discussion:
- We wil reform the schema to accept null values for the tabId.
- In the cases of anonymous tracking, no storage and access denial on session storage we will return `null`.
- When the web_page context is deactivated and the `getTabId()` API is called, we again will return null.

matus-tomlein:
The last changes look good! Lets just wait for the integration tests to pass and I think we are good.

Speaking of integration tests, there are a few places where the old schema version is referenced, I think those will keep failing even when the schema is available on Iglu central. See [here for instance](https://github.com/snowplow/snowplow-javascript-tracker/blob/issue/815-add-tab-identifier/trackers/javascript-tracker/test/integration/integration.test.ts#L277) – btw, at this place, if you have a bit of time, it would be nice to also add a short integration test case for the `tabId` in addition to the page view ID test. It's not super important, but it'd be nice to have that!

libraries/browser-tracker-core/src/tracker/index.ts
paulboocock: I appreciate the link of this feature to browser storage, however conceptually, it seems strange to have a `null` `tabId` - since all events from a browser must be done from within a tab.

I think when we consider anonymous tracking, having a `null` id makes sense. Perhaps we should introduce a flag like we have done for session tracking - so users can keep the tabId even when anonymous tracking is on but by default we set it to `null`. I think this is reason enough to allow `null` in the schema.

What do we do for session id when the `stateStorageStrategy` is `none`? I feel like the tabId should probably behave similarly to that? Unless however we've implemented is that it rotates all the time (which is pointless but I can't remember), in which case `null` makes more sense. As I type this I'm imagining `null` does actually make sense, especially for non-SPA, as if there's no storage then we can't keep track of the tab between page loads.

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -180,12 +183,15 @@ export function Tracker(
...
-      getAnonymousTracking = (config: TrackerConfiguration) => !!config.anonymousTracking;
+      getAnonymousTracking = (config: TrackerConfiguration) => !!config.anonymousTracking,
+      isWebPageContextAvailable = trackerConfiguration?.contexts?.webPage ?? true;
 
     // Get all injected plugins
     browserPlugins.push(getBrowserDataPlugin());
-    if (trackerConfiguration?.contexts?.webPage ?? true) {
-      browserPlugins.push(getWebPagePlugin()); // Defaults to including the Web Page context
+    /* When including the Web Page context, we add the relevant internal plugins */
+    if (isWebPageContextAvailable) {
```
matus-tomlein, line 192: Question: I am not sure what is the better approach here, but when comparing with the mobile trackers, we have two separate configuration properties for activating the screen context (same as webpage context) and for the mobile context (same as browser context). Why did you decide to have both under the same property?

It is quite theoretical, but I can imagine someone wanting the web page context for the page view ID and not wanting to spend the extra payload size on the browser context entity.

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -773,11 +813,13 @@ export function Tracker(
           updateFirstEventInIdCookie(idCookie, payloadBuilder);
           incrementEventIndexInIdCookie(idCookie);
 
-          payloadBuilder.add('vp', detectViewport());
-          payloadBuilder.add('ds', detectDocumentSize());
+          const { viewport, documentSize } = getBrowserProperties();
+
+          payloadBuilder.add('vp', viewport);
```
matus-tomlein, line 818: Question: there was a discussion whether we should still set the canonical event properties when the browser context is enabled in v4 – where do we land on that? It's probably a big change to stop tracking them by default, but maybe we could let users optionally disable them?

matus-tomlein (approved):
Other than the pending question, this looks good to me! 👍

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -728,6 +752,22 @@ export function Tracker(
...
 
+    function getBrowserContextPlugin() {
+      return {
+        contexts: () => {
+          return [
+            {
+              schema: BROWSER_CONTEXT_SCHEMA,
+              data: {
+                ...getBrowserProperties(),
+                tabId: getTabId(),
```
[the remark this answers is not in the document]
igneel64: Yeap it was something I struggled a bit but decided not to based on the following thought:
- The `getBrowserProperties` in its current form returns platform properties without any additional logic or dependency on the snowplow tracking state.
- Adding the tab identifier there would require injection of three 'state' properties whenever you call the function, `configStateStorageStrategy` , `configAnonymousTracking` and `isWebPageContextAvailable`, adding extra overhead and more 'responsibility' versus returning just browser attributes.
-  There are other places currently that we require some of these attributes where we call the function so until we devise a better way to share the above state, _rather than rely on JS scope_, I kept it as straightforward as possible in my mind.

libraries/browser-tracker-core/src/tracker/index.ts
igneel64: Based on an offline discussion we can make this context an opt-out and remove these properties. However let's discuss it more in the context of v4.

libraries/browser-tracker-core/src/tracker/index.ts
igneel64: Yeap fixed!
`contexts: { browser: true }`

PR #1090, opened 2022-08-24
Fix built-in context typing for newTracker (closes #1089)

### LINKED ISSUE #1089: Typing issue on tracker configuration for built-in contexts
**Describe the bug**
When initializing a new tracker using the `newTracker` method, the configuration object for the built-in contexts expects both options to be provided. This results in a weird experience for the developer since both have a default value.

**To Reproduce**
Any TypeScript capable project.

**Expected behavior**
You can pass any one of the available built-in context options, without explicitly requiring the other ones to be passed as well.

**Screenshots**
context

igneel64:
## Fix `context` option typing for `newTracker`

Allow for passing either `webPage` or `session` options as expexted.
Adds the `BuiltInContexts` type.

ctx

### Notes
- Add the RequireAtLeastOne helper for types. _(Could prove useful in the future for the whole project)_ Rightfully stolen from [@azure](https://docs.microsoft.com/en-us/javascript/api/@azure/keyvault-certificates/requireatleastone?view=azure-node-latest)
- Some typos

closes #1089


snowplowcla:

libraries/browser-tracker-core/src/tracker/types.ts
```diff
@@ -189,15 +199,15 @@ export type TrackerConfiguration = {
    */
   connectionTimeout?: number;
   /**
-   * Condifugration for Anonymous Tracking
```
matus-tomlein, line 192: Some of the typos are quite funny :D

matus-tomlein (approved):
This is really cool, I didn't know about the trick!

I'm just wondering if it is a breaking change? Technically it might break the build for some users in case they set an empty dictionary in their apps? If so, strictly speaking we should add this to the next major release instead of a minor. (btw, don't forget to change the base branch to a release branch before merging)

igneel64:
> This is really cool, I didn't know about the trick!
> 
> I'm just wondering if it is a breaking change? Technically it might break the build for some users in case they set an empty dictionary in their apps? If so, strictly speaking we should add this to the next major release instead of a minor. (btw, don't forget to change the base branch to a release branch before merging)

Accounted for people that could have added `{}` as the input on `contexts`. This makes it a non breaking change.
_(Will add a PR for next major release to remove it completely)_

libraries/browser-tracker-core/src/tracker/types.ts
paulboocock: Have you never tried to Condifugrate your trackers before?

PR #1099, opened 2022-08-31
Add tests to verify Node.js tracker returns eid (event id) attribute (closes #1098)

### LINKED ISSUE #1098: Return event id ( eid ) from calling the `track` method in Node.js
**Describe what you'd like**
Add a way to validate that the event Id is accessible in the Node.js `track` method return value.

**Describe your solution**
Add tests to showcase and validate this behaviour.

**Additional context**
Other trackers implement the same functionality _(some of them only by returning the event_id and not the rest of the payload)_. On the Node.js case, the tracker already returns the values of the full payload including the event_id, we need to validate and showcase this through tests.

igneel64:
### Description
As per other trackers, e.g. [objc](https://github.com/snowplow/snowplow-objc-tracker/pull/711), we implemented returning the event id from the `track` method. 
In the node-tracker, this was already possible as the `tracker.track` method which is used in the package, returns the event payload making `eid` accessible like so:
`const { eid } = tracker.track(...);`

This PR just adds a simple test verifying this behaviour.

#### _Note_
Additionally since the node-tracker uses the `tracker.track()` method from the core directly, a test was added there for completion and future abstraction possibility.

(closes #1098)

### DIFF
--- a/common/changes/@snowplow/node-tracker/feature-1098-verify-eid-returned-node-track_2022-08-31-10-09.json
+++ b/common/changes/@snowplow/node-tracker/feature-1098-verify-eid-returned-node-track_2022-08-31-10-09.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/node-tracker",
+      "comment": "Add tests verifying that the track method returns event id (eid) in the payload (#1098)",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/node-tracker"
+}
--- a/common/changes/@snowplow/tracker-core/feature-1098-verify-eid-returned-node-track_2022-08-31-10-09.json
+++ b/common/changes/@snowplow/tracker-core/feature-1098-verify-eid-returned-node-track_2022-08-31-10-09.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/tracker-core",
+      "comment": "Add tests verifying that the track method returns event id (eid) in the payload (#1098)",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/tracker-core"
+}
--- a/libraries/tracker-core/test/core.ts
+++ b/libraries/tracker-core/test/core.ts
@@ -53,6 +53,8 @@ import {
 } from '../src/core';
 import { Payload } from '../src/payload';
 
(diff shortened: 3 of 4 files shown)


snowplowcla:

matus-tomlein (commented):
It looks good and it's a nice idea to add tests to showcase and validate the behavior!

I am just thinking whether we shouldn't update the NodeJS tracker `tracker.track()` API to return event IDs as strings (similar to Java, C++, iOS and Android trackers). This gives less flexibility to the users, but I am always nervous about exposing more than necessary to the users as they will start depending on it and it will become more difficult to change the internal details of the trackers. I would consider the `Payload` properties an internal detail of how events are sent to the Collector. The users should not rely on the atomic fields and if there is some information that we should expose to them, lets expose it in a more explicit and structured way. This is a topic for a deeper discussion at some point, just writing my thoughts here. The upcoming v4 release could be a good opportunity to bring the NodeJS tracker API closer to the other trackers at the cost of some breaking changes. Btw, also don't like that we surface the `PayloadBuilder` objects to users.

Also one nitpick – please make sure to have a consistent way of closing the referenced issues in the PR titles. Looking at the PRs, you have used a bunch of different formats and capitalizations. Usually we follow the format: `Do something (close #XX)`. This is kind of important because when you select "Stash and merge" in the Github PR to merge the issue branch into a release branch, it will use the PR title in the commit name.

igneel64:
@matus-tomlein 
Yes I already added the discussion point on some of our v4 roadmap. 💪

common/changes/@snowplow/node-tracker/feature-1098-verify-eid-returned-node-track_2022-08-31-10-09.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/node-tracker",
+      "comment": "Add tests verifying that the track method returns event id (eid) in the payload",
```
matus-tomlein, line 5: We usually add the issue number in the changelog to keep the reference.

```suggestion
      "comment": "Add tests verifying that the track method returns event id (eid) in the payload (#1098)",
```

matus-tomlein (approved):
Sounds good! We can add the tests regardless of the v4 API change as they are valuable.

PR #1104, opened 2022-09-02
Remove user and session identifiers from cookies when anonymous tracking becomes enabled (close #1103)

### LINKED ISSUE #1103: Remove user and session identifiers from cookies when anonymous tracking becomes enabled
Currently, when one enables anonymous tracking while having had it disabled previously, the tracker does not clean up existing identifiers in cookies. The identifiers are not tracked in events (they are masked when the events are tracked) but they are still present in the cookies. It is possible to clear them using `snowplow('clearUserData')` but one would expect that enabling anonymous tracking clears them by automatically (also there is no option to only clear the user ID while keeping the session ID the same).

The proposed solution is to clear the `sp_id` cookie when anonymous tracking is enabled. In case anonymous tracking is enabled with session tracking, we should clear the user ID within the `sp_id` cookie.

matus-tomlein:
This addresses issue #1103 with the goal to clear the `sp_id` cookie when anonymous tracking is enabled.

The problem does not occur in case anonymous tracking wasn't disabled before. For example, if `newTracker()` is called with anonymous tracking already enabled, no identifiers are stored in the cookies. However, if `newTracker()` is called without anonymous tracking enabled, and later the user calls `enableAnonymousTracking()`, the user and session identifiers remain in cookies.

The new suggested behaviour can have two forms:

1. If form tracking is enabled without session tracking, the whole `sp_id` cookie should be removed. This is equivalent to calling `clearUserData()` in the current API – the cookie will be removed and in-memory identifiers reset.
2. If form tracking is enabled with session tracking, we shouldn't clear the whole cookie, just remove the `domain_userid` from it while maintaining the same session ID and rest of the `sp_id` cookie.

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/index.ts
+++ b/libraries/browser-tracker-core/src/tracker/index.ts
@@ -642,7 +642,7 @@ export function Tracker(
     /*
      * Load the domain user ID and the session ID
      * Set the cookies (if cookies are enabled)
-     * If anonymous tracking is enabled, it clears the anonymised identifiers
+     * If anonymous tracking is enabled, it clears the stored identifiers that should be anonymous
      */
     function initializeIdsAndCookies() {
       if (configAnonymousTracking && !configAnonymousSessionTracking) {
--- a/libraries/browser-tracker-core/src/tracker/index.ts
+++ b/libraries/browser-tracker-core/src/tracker/index.ts
@@ -642,9 +642,14 @@ export function Tracker(
     /*
      * Load the domain user ID and the session ID
      * Set the cookies (if cookies are enabled)
+     * If anonymous tracking is enabled, it clears the anonymised identifiers
      */
     function initializeIdsAndCookies() {
       if (configAnonymousTracking && !configAnonymousSessionTracking) {
+        clearUserDataAndCookies({
+          preserveSession: false,
+          preserveUser: false,
+        });
         return;
       }
 
@@ -1200,6 +1205,9 @@ export function Tracker(
         trackerConfiguration.anonymousTracking = (configuration && configuration?.options) ?? true;
 
         toggleAnonymousTracking(configuration);
+
+        // Re-initialize the ID cookie which clears the domain_userid (if session tracking) or clears the whole cookie
+        initializeIdsAndCookies();
 
         // Reset the page view, if not tracking the session, so can't stitch user into new events on the page view id
         if (!configAnonymousSessionTracking) {
--- a/common/changes/@snowplow/browser-tracker-core/issue-1103-clear_cookies_anonymous_tracking_2022-09-02-14-33.json
+++ b/common/changes/@snowplow/browser-tracker-core/issue-1103-clear_cookies_anonymous_tracking_2022-09-02-14-33.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
+      "comment": "Remove user and session identifiers from cookies when anonymous tracking becomes enabled (#1103)",
+      "type": "none"
+    }
+  ],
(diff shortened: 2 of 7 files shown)


snowplowcla:

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -642,9 +642,14 @@ export function Tracker(
     /*
      * Load the domain user ID and the session ID
      * Set the cookies (if cookies are enabled)
+     * If anonymous tracking is enabled, it clears the anonymised identifiers
```
igneel64, line 645: Nit:
`If anonymous tracking is enabled, it clears the anonymised identifiers`
to
`If anonymous tracking is enabled, it clears the stored identifiers`

Seems more clear to me on what it does practically. WDYT ?

igneel64 (approved):
LGTM 👍 

One concern though:
In the case of multiple trackers, if one enables anonymous tracking, it breaks the identifiers for all trackers on the page. Do we cater for this case ?

There are a couple of scenarios I can think were this might cause unexpected changes after the release.

libraries/browser-tracker-core/src/tracker/index.ts
matus-tomlein: Sounds better, changed it!

matus-tomlein:
> In the case of multiple trackers, if one enables anonymous tracking, it breaks the identifiers for all trackers on the page. Do we cater for this case ?

That's a good point @igneel64, didn't think about that! I want to say that enabling anonymous tracking should be a global action that should affect all trackers (and Snowplow tags) but I do have some doubts. Thinking about it now, I recall that there were use cases when users had multiple trackers, one with anonymous tracking and one without anonymous tracking for some reason. I can't recall the specific use case, but remember that there was a need for that.

It would still be possible to support that if anonymous tracking state was set in the `newTracker()` call, but changing the anonymous tracking state with `enableAnonymousTracking()` would be more destructive. This might be unintuitive for users with multiple Snowplow tags.

@paulboocock can you recall if there are any use cases that we would break with automatically clearing the user data when anonymous tracking is enabled?

paulboocock:
There's one use case that stuck out for me when you described this. Thats where the user opts in to being tracking (i.e. anonymous tracking is disabled), and then decides to opt out. Some users want to keep hold of the identifiers, in case the user opts back in again. Perhaps holding these in memory (and not in a cookie) so if the user opts back in immediately (i.e. no page refresh) then we can rehydrate the same cookie? I actually think thats what we attempt to do anyway if my memory serves me correctly.

I'm not sure if its our decision to force upon our users what we think is right or wrong in this scenario, but to give our users options. I think my original intention is that if users want all the bits and bobs cleaned up they should call `clearUserData()` after enabling anonymous tracking, which is now less clear on what one should clear given your description of storing different bits depending on the anonymous tracking state.

I wonder if the API would be nicer if we popped a flag in `EnableAnonymousTrackingConfiguration` which allowed the user data to be cleared when enabling it (false by default? or true if we're heading to v4 here).

matus-tomlein:
Thanks @paulboocock. It seems to me that this change is not a clear win and I think we should close it without merging. There are a lot of ways that users might make use of anonymous tracking and we are bound to break some of the use cases by forcing this behaviour. Adding it is a configuration option to the `EnableAnonymousTrackingConfiguration` is a good idea but I think it just complicates the API and can cause more user confusion as there already is a way to clear user data and this is not that different.

Looking at it back, it was an overreaction to a customer ticket where a user was confused about the identifiers in cookies after enabling anonymous tracking. It probably doesn't warrant changing the behaviour as the previous one was not inherently bad.

I will close this PR and issue and instead make the documentation clearer about what happens and when are the identifiers created and cleared.

PR #1108, opened 2022-09-06
Correct initial session_index value and after clearUserData() (closes #1106 #1107)

### LINKED ISSUE #1106: Fix domain_sessionidx starting at 0 after clearUserData is called
If a user calls the `clearUserData()` function that the Web trackers provide, subsequent events are tracked with a session where the visit count (or `session_idx`) is 0.

I think we should never allow the visit count to start at 0, it always needs to start at 1.

### LINKED ISSUE #1107: Domain session index value retrieved from the tracker is wrong if no event has been sent to the tracker.
**Describe the bug**
When creating a tracker, the developer can opt for querying the value of the sessionIndex _(or otherwise visitCount)_ to add some logic based on it. 
When the developer tries to retrieve the value using the `tracker.getDomainSessionIndex()` method, the value returned is the default (_1_) and not the correct one.

**To Reproduce**
Navigate on a domain with an Id type cookie having recorded more than one sessions and try to query the `tracker.getDomainSessionIndex()` before any kind of event has been tracked.
In code this would look like:

```ts
// A customer has a session_idx of 5 stored on the cookie and assuming no new session is created

// The app loads and initiates the tracker
const t = addTracker(...);
t.getDomainSessionIndex(); // Returns 1 from the default

// If an event is being triggered
t.trackPageview(...)
t.getDomainSessionIndex(); // Returns 5
```

**Expected behavior**
`tracker.getDomainSessionIndex()` Returns the correct value.

**Screenshots**
image

igneel64:
### Description
- Correct the session_index value retrieved from `tracker.getDomainSessionIndex` before any event has been sent
- Correct the session_index value after calling `clearUserData()`;

closes #1106
closes #1107

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/index.ts
+++ b/libraries/browser-tracker-core/src/tracker/index.ts
@@ -305,7 +305,8 @@ export function Tracker(
         installed: false, // Guard against installing the activity tracker more than once per Tracker instance
         configurations: {},
       },
-      configSessionContext = trackerConfiguration.contexts?.session ?? false;
+      configSessionContext = trackerConfiguration.contexts?.session ?? false,
+      toOptoutByCookie: string | boolean;
 
     if (trackerConfiguration.hasOwnProperty('discoverRootDomain') && trackerConfiguration.discoverRootDomain) {
       configCookieDomain = findRootDomain(configCookieSameSite, configCookieSecure);
@@ -446,14 +447,6 @@ export function Tracker(
      * Send request
      */
     function sendRequest(request: PayloadBuilder) {
-      // Set to true if Opt-out cookie is defined
-      let toOptoutByCookie;
-      if (configOptOutCookie) {
-        toOptoutByCookie = !!cookie(configOptOutCookie);
-      } else {
-        toOptoutByCookie = false;
-      }
-
       if (!(configDoNotTrack || toOptoutByCookie)) {
         outQueue.enqueueRequest(request.build(), configCollectorUrl);
       }
@@ -610,7 +603,7 @@ export function Tracker(
       deleteCookie(sesname, configCookieDomain, configCookieSameSite, configCookieSecure);
       if (!configuration?.preserveSession) {
         memorizedSessionId = uuid();
-        memorizedVisitCount = 0;
+        memorizedVisitCount = 1;
       }
       if (!configuration?.preserveUser) {
         domainUserId = uuid();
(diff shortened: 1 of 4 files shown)


libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -447,13 +448,6 @@ export function Tracker(
      */
     function sendRequest(request: PayloadBuilder) {
       // Set to true if Opt-out cookie is defined
```
matus-tomlein, line 450: This comment probably should not be here anymore as the local variable was removed.

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -652,6 +646,7 @@ export function Tracker(
       const idCookie = loadDomainUserIdCookie();
 
       domainUserId = initializeDomainUserId(idCookie, configAnonymousTracking);
+      memorizedVisitCount = visitCountFromIdCookie(idCookie);
```
matus-tomlein, line 649: Shouldn't this go a few lines lower behind the if condition? The `startNewIdCookieSession()` function can change (increment) the visit count, so we should access the newer value.

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -610,7 +604,7 @@ export function Tracker(
       deleteCookie(sesname, configCookieDomain, configCookieSameSite, configCookieSecure);
       if (!configuration?.preserveSession) {
         memorizedSessionId = uuid();
-        memorizedVisitCount = 0;
+        memorizedVisitCount = 1;
```
matus-tomlein, line 607: I'm so confused why was this set to 0 before, it seems that it was expected that right after `clearUserData()` is called there would be a new session created that would increment it to 1, but I don't see how that could happen 🤷  It looks good, just makes me a bit nervous that I don't understand the original intention.

libraries/browser-tracker-core/src/tracker/index.ts
igneel64: Looking into the code, the `getBrowserDataPlugin` is always added without any condition on the tracking side and the `beforeTrack` is always called together with the `sendRequest` function.
To declutter the duplication a bit, I took the logic and left it on just one place, which for the moment coincided with my thinking that `sendRequest` should do one thing only. But I totally agree that it is a point of improvement. I will squeez it in on this or the next touch of the file.

igneel64:
@matus-tomlein Yes this is something that we will fix moving forward. I feel the same way and especially with the file size of the core but even more so the `scoping` of variables that we (_ab_)use to change state. We will fix it as we go!

matus-tomlein (approved):
LGTM! 👍

PR #1110, opened 2022-09-28
Introduce Snowplow Ecommerce browser plugin

igneel64:
### Description
Initial draft on the Ecommerce Accelerator browser plugin.

### Initial API

  | Used for: | Expected usage:
-- | -- | --
trackProductView | Tracking a visit to a product page. Known also as product detail view. | `trackProductView({id: 'P1234', price: 25,  currency: 'EUR',  brand: 'Snowplow',  });` 
trackAddToCart | Track an addition to cart. | `trackAddToCart({ total_value: 40, currency: 'EUR', products: [{  id: 'P1234', price: 12,currency: 'EUR' }]});`
trackRemoveFromCart | Track a removal from cart. | `trackRemoveFromCart({ total_value: 40,currency: 'EUR',products: [{ id: 'P1234',price: 12,currency: 'EUR' }]});`
trackProductListView | Track an impression of a product list. The list could be a search results page, recommended products, upsells etc. | `trackProductListView({ name: 'shop the look', products: [ { id: 'P1234', price: 12, currency: 'EUR', position: 1 },{ id: 'P1235', price: 14, currency: 'EUR', position: 2 }, ], });`
trackProductListClick | Track the click/selection of a product from a product list. | `trackProductListClick({ name: 'shop the look', product: { id: 'P1235', price: 14, currency: 'EUR', position: 2 } });`
trackCheckoutStep | Track a checkout step completion in the checkout process together with common step attributes for user choices throughout the checkout funnel. | `trackCheckoutStep({ step: 2, delivery_method: 'express delivery', delivery_provider: 'PostNL' });`
trackTransaction | Track a transaction/purchase completion. | `trackTransaction({ transaction_id: 'T123', currency: 'EUR', revenue: 26, payment_method: 'Alipay', products: [{ id: 'P1234', price: 12, currency: 'EUR', position: 1 }, { id: 'P1235', price: 14, currency: 'EUR', position: 2 } ]});`
setPageType | Set a Page type context which would allow the analyst to discern between types of pages with ecommerce value. E.g. Category Page, Product Page, Cart Page, etc. | `setPageType({ type: 'cart', locale: 'en' });`
setEcommerceUser | Set a User type context with a few standard user attributes. | `setEcommerceUser({ id: 'U123', is_guest: true });`

### DIFF
--- a/plugins/browser-plugin-snowplow-ecommerce/package.json
+++ b/plugins/browser-plugin-snowplow-ecommerce/package.json
@@ -1,0 +1,55 @@
+{
+  "name": "@snowplow/browser-plugin-snowplow-ecommerce",
+  "version": "3.7.0",
+  "description": "Snowplow ecommerce tracking",
+  "homepage": "http://bit.ly/sp-js",
+  "bugs": "https://github.com/snowplow/snowplow-javascript-tracker/issues",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/snowplow/snowplow-javascript-tracker.git"
+  },
+  "license": "BSD-3-Clause",
+  "author": "Peter Perlepes",
+  "sideEffects": false,
+  "main": "./dist/index.umd.js",
+  "module": "./dist/index.module.js",
+  "types": "./dist/index.module.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "rollup -c --silent --failAfterWarnings",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@snowplow/browser-tracker-core": "workspace:*",
+    "@snowplow/tracker-core": "workspace:*",
+    "tslib": "^2.3.1"
+  },
+  "devDependencies": {
+    "@ampproject/rollup-plugin-closure-compiler": "~0.27.0",
+    "@rollup/plugin-commonjs": "~21.0.2",
+    "@rollup/plugin-node-resolve": "~13.1.3",
+    "@types/jest": "~27.4.1",
+    "@types/jsdom": "~16.2.14",
+    "@types/lodash": "~4.14.180",
+    "@typescript-eslint/eslint-plugin": "~5.15.0",
+    "@typescript-eslint/parser": "~5.15.0",
+    "eslint": "~8.11.0",
+    "jest": "~27.5.1",
+    "jest-environment-jsdom": "~27.5.1",
+    "jest-environment-jsdom-global": "~3.0.0",
+    "jest-standard-reporter": "~2.0.0",
+    "lodash": "~4.17.21",
+    "rollup": "~2.70.1",
+    "rollup-plugin-cleanup": "~3.2.1",
+    "rollup-plugin-license": "~2.6.1",
+    "rollup-plugin-terser": "~7.0.2",
+    "rollup-plugin-ts": "~2.0.5",
+    "ts-jest": "~27.1.3",
+    "typescript": "~4.6.2"
+  },
+  "peerDependencies": {
+    "@snowplow/browser-tracker": "~3.5.0"
+  }
+}
--- a/common/changes/@snowplow/browser-plugin-snowplow-ecommerce/feature-ecommerce-accelerator-browser_2022-11-15-10-15.json
+++ b/common/changes/@snowplow/browser-plugin-snowplow-ecommerce/feature-ecommerce-accelerator-browser_2022-11-15-10-15.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-plugin-snowplow-ecommerce",
+      "comment": "Introduce Snowplow Ecommerce browser plugin",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/browser-plugin-snowplow-ecommerce"
+}
--- a/common/changes/@snowplow/javascript-tracker/feature-ecommerce-accelerator-browser_2022-12-21-15-55.json
+++ b/common/changes/@snowplow/javascript-tracker/feature-ecommerce-accelerator-browser_2022-12-21-15-55.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/javascript-tracker",
+      "comment": "Add Snowplow Ecommerce browser plugin",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/javascript-tracker"
+}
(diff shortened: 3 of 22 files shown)


snowplowcla:

common/changes/@snowplow/browser-plugin-accelerator-ecommerce/feature-ecommerce-accelerator-browser_2022-09-28-11-47.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-plugin-accelerator-ecommerce",
```
matus-tomlein, line 4: You might have already discussed and pondered this, so ignore me if so, but the naming of the plugin raises two small concerns for me:

1. Should we tie the plugin to the accelerator? In my view, the plugin is more widely usable, also outside of the accelerator. Using accelerator in the name seems like constraining the plugin a bit to me. For some users, it might just be the ecommerce plugin that they use in their apps long term, not just when they were learning through an accelerator, right?
2. We already have two ecommerce plugins (ecommerce and enhanced-ecommerce). Adding another makes this even more confusing for users. Could we instead have this be the next major version of the ecommerce or enhanced ecommerce plugins?

plugins/browser-plugin-accelerator-ecommerce/src/api.ts
```diff
@@ -0,0 +1,228 @@
...
+  products.forEach((product) => context.push({ schema: PRODUCT_SCHEMA, data: { ...product } }));
+  context.push({ schema: CART_SCHEMA, data: { ...cart } });
+
+  dispatchToTrackersInCollection(trackers, _trackers, (t) => {
+    t.core.track(buildEcommerceActionEvent({ type: 'add_to_cart' }), context, timestamp);
+  });
+}
+
+/**
+ * Track a product removal from cart
```
matus-tomlein, line 119: This is a bit confusing to me. Does one track the state of the cart after the products were removed or does one track the products to be removed? I assume the second, but using the `Cart` type to represent the `event` leads me to think the first.

Also the comment shouldn't probably talk about a singular product since the `Cart` type only has a `products` list.

I think what would help me would be to put examples in these docstrings. I think that'd actually be a good practice and I have been trying to do that elsewhere. We usually use the Snowplow docs to give examples, but putting them right in the docstrings makes it easier for users to understand the API in IDEs (users love examples). I'm thinking something like:

```suggestion
 * Track removal of products from cart
 *
 * Example:
 * ```
 * trackRemoveFromCart({
 *   products: [
 *      { id: 123 } // product to be removed from the cart
 *   ]
 * });
 * ```
```

Maybe I'm just exaggerating the problem, so just leaving for you to consider.

plugins/browser-plugin-accelerator-ecommerce/src/types.ts
```diff
@@ -0,0 +1,230 @@
...
+  /**
+   * The total value of the cart after this interaction
+   */
+  total_value: number;
+  /**
+   * The currency used for this cart (ISO 4217)
+   */
+  currency: string;
+  /**
+   * Array of products added/removed from cart
```
matus-tomlein, line 73: Now I see what confused me above, this explains that the `products` are only the changed products. Maybe naming this entity `CartChange` or something like that to invoke that it's the diff, not the actual "cart" would make it clearer for me.

plugins/browser-plugin-accelerator-ecommerce/src/api.ts
```diff
@@ -0,0 +1,228 @@
...
+}
+
+/**
+ * Track a product list view
+ *
+ * @param event - The list name along with the products which were viewed by the visitor.
+ * @param trackers - The tracker identifiers which the event will be sent to
+ */
+export function trackProductListView(
+  event: ListViewEvent & CommonEcommerceEventProperties,
```
matus-tomlein, line 70: Is the `ListViewEvent` really an event? Strictly speaking there is only one event type, the ecommerce action. Everything else seems like entities to me.

Also a nit: I wouldn't name the first argument `event` in these track functions but instead call them `listView` or `cart` or whatever they represent.

plugins/browser-plugin-accelerator-ecommerce/src/types.ts
```diff
@@ -0,0 +1,230 @@
...
+   */
+  variant?: string | null;
+  /**
+   * The brand of the product
+   */
+  brand?: string | null;
+  /**
+   * The inventory status of the product (e.g. in stock, out of stock, preorder, backorder, etc)
+   */
+  inventory_status?: string | null;
```
matus-tomlein, line 140: Why do some of the optional properties have the `null` as an accepted type while others like `creative_id` don't?

matus-tomlein (commented):
It looks good, very nice API!

common/changes/@snowplow/browser-plugin-accelerator-ecommerce/feature-ecommerce-accelerator-browser_2022-09-28-11-47.json
paulboocock: I agree, I don't want to couple this to the accelerator, but I also don't really fancy calling it `browser-plugin-really-enhanced-ecommerce`.

Maybe `browser-plugin-retail` (? also begs the question if we call this ecommerce or retail as it perhap has broader applications than _just_ ecommerce). This can go out in a `v3` release, where we also mark the other two as deprecated. Then in v4 we remove them and add this new one into sp.js (which is a big move but one I think we can navigate with messaging if this launches in a v3).

paulboocock:
API looks good, I think Matus raises valuable questions and the naming is a pain which we need to sort soon. I think we can have some further discussions on what we want to call this, the dbt model and the accelerator before launch.

plugins/browser-plugin-snowplow-ecommerce/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@snowplow/browser-plugin-snowplow-ecommerce",
+  "version": "3.7.0",
```
Jack-Keene, line 3: Does the version need to be bumped? Or is that done in the release?

Jack-Keene (approved):
LGTM!

plugins/browser-plugin-snowplow-ecommerce/package.json
igneel64: This is done during the release command :)

plugins/browser-plugin-accelerator-ecommerce/src/types.ts
igneel64: Fixed btw

PR #1115, opened 2022-10-09
Use shared logger for Youtube plugin loading and add helpful message for missing plugin initialization ( closes #1113 )

### LINKED ISSUE #1113: Fail gracefully on youtube tracking without plugin enablement
**Describe the bug**
When the client might have forgotten to include the YT plugin and tries to use `enableYouTubeTracking`, the app fails as the `LOG` object is `undefined`. This creates an error thrown from the plugin, which can break the app in many cases, while it could be a helpful error message.

**To Reproduce**
Try to initiate `enableYouTubeTracking` without the plugin and before a the video/iframe target element is in the DOM.

**Expected behavior**
A helpful error message in the console ?

**Screenshots**
image
image

igneel64:
### Description
Use the shared `LOG` function from `@snowplow/tracker-core` to prevent undefined state before plugin initialization.

### Notes
- Added a _somewhat_ helpful message when no tracker has the plugin installed.
- Why did we not use the LOG object initially and we needed to use the `logger` callback ? 🤔  We might need to fix this in a couple more places if there was no valid reason.

closes #1113

### DIFF
--- a/plugins/browser-plugin-youtube-tracking/src/index.ts
+++ b/plugins/browser-plugin-youtube-tracking/src/index.ts
@@ -38,7 +38,7 @@ import {
 } from './constants';
 import { EventData, EventGroup, MediaTrackingOptions, TrackedPlayer, TrackingOptions } from './types';
 import { BrowserPlugin, BrowserTracker, dispatchToTrackersInCollection } from '@snowplow/browser-tracker-core';
-import { buildSelfDescribingEvent, CommonEventProperties, Logger, SelfDescribingJson } from '@snowplow/tracker-core';
+import { buildSelfDescribingEvent, CommonEventProperties, LOG, SelfDescribingJson } from '@snowplow/tracker-core';
 import { SnowplowEvent } from './snowplowEvents';
 import { MediaPlayerEvent } from './contexts';
 import { buildYouTubeEvent } from './buildYouTubeEvent';
@@ -48,15 +48,11 @@ import { DefaultEvents, EventGroups, AllEvents } from './eventGroups';
 const _trackers: Record<string, BrowserTracker> = {};
 const trackedPlayers: Record<string, TrackedPlayer> = {};
 const trackingQueue: Array<TrackingOptions> = [];
-let LOG: Logger;
 
 export function YouTubeTrackingPlugin(): BrowserPlugin {
   return {
     activateBrowserPlugin: (tracker: BrowserTracker) => {
       _trackers[tracker.id] = tracker;
-    },
-    logger: (logger) => {
-      LOG = logger;
     },
   };
 }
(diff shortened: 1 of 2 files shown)


snowplowcla:

greg-el:
> Why did we not use the LOG object initially and we needed to use the logger callback? 

I think I used `Logger` with the callback as it was a pattern that was in [other plugins](https://github.com/snowplow/snowplow-javascript-tracker/blob/4dec01b9a314ca11d88372afa4b2dc67a431ee0d/plugins/browser-plugin-consent/src/index.ts#L97). @paulboocock might be able to clear up if just importing `LOG` has any sneaky detrimental effects.

plugins/browser-plugin-youtube-tracking/src/index.ts
```diff
@@ -126,10 +122,14 @@ export function trackingOptionsParser(mediaId: string, conf?: MediaTrackingOptio
...
 export function enableYouTubeTracking(args: { id: string; options?: MediaTrackingOptions }) {
+  if (!Object.keys(_trackers).length) {
+    LOG.error('You might have forgotten to initialize the YoutubeTrackingPlugin on your tracker configuration.');
+    return;
+  }
   const conf: TrackingOptions = trackingOptionsParser(args.id, args.options);
   const el: HTMLIFrameElement = document.getElementById(args.id) as HTMLIFrameElement;
   if (!el) {
-    LOG.error('Cannot find YouTube iframe');
+    LOG.error('Cannot find YouTube iframe.');
```
paulboocock, line 132: Not sure the full stop is worth the extra bits ;-)

plugins/browser-plugin-youtube-tracking/src/index.ts
```diff
@@ -126,10 +122,14 @@ export function trackingOptionsParser(mediaId: string, conf?: MediaTrackingOptio
 }
 
 export function enableYouTubeTracking(args: { id: string; options?: MediaTrackingOptions }) {
+  if (!Object.keys(_trackers).length) {
+    LOG.error('You might have forgotten to initialize the YoutubeTrackingPlugin on your tracker configuration.');
```
paulboocock, line 126: nit: I try and keep strings as short as possible (as they don't minify). This is a little shorter:

```suggestion
    LOG.error('Check YoutubeTrackingPlugin is initialized in tracker config');
```

paulboocock:
> Why did we not use the LOG object initially and we needed to use the logger callback?

If I recall, I intended to let people inject their own logger. However, I think this is one of those things I threw around a few ways of implementing, ended up with an exported constant from `core` and never got around to implementing anything that lets people inject the `core` logger.

Two choices:
- I think you can switch to simply importing `LOG` everywhere and then in `v4` removing the `logger` callback.
- Implement the idea of letting people push their own logging implementation if they wish.

igneel64:
@paulboocock Fixed the dots 😆 

In this PR I will keep the change as it is for now since it was causing a cryptic error for users. Moving forwards let's discuss the details internally.

greg-el (approved):
LGTM!

PR #1117, opened 2022-10-12
Support JWPlayer in browser-plugin-media-tracking (closes #1116)

### LINKED ISSUE #1116: Support JWPlayer on media-tracking plugin
### Description
JWPlayer is from the most common video players out there. It is expected that our media-tracking plugin should work out of the box it. Currently there are a couple of issues that prevent us tracking that properly.

1. The video elements in the JWPlayer, do not have an id set, only the target of the `jwplayer.setup` call. That is currently breaking tracking as the `htmlId` attribute is required on our media_element schema.
2. When the [poster functionality](https://docs.jwplayer.com/platform/reference/vdh-manage-a-media-poster) is used, JWPlayer internally creates two `video` elements under the container. This again breaks tracking as we do not allow more than one media elements under the target container.

Find a way to make it work seamlessly.

Changing player seems not a valid 'alternative'.

igneel64:
This PR adds a minimal set of changes to be able to support JWPlayer media. Additionally the JWPlayer media with poster functionality.

### Notes
- The check that has been added for the length of the detected media could be in some way enhanced with some `classname` search, but I do not think that it is required in this case.
- JWplayer adds the poster video on an additional 'second' video element, that is why we select the first which is the functional one.
- Do you think we should create an exhaustive jwplayer check ?

- There is an edge case for existing clients will see new events if they update to the latest version that we will publish.
This edge case is described in the following scenario:
1. A client has added media tracking.
2. The client had targeted a container element with exactly two video elements.
3. The tracking would not work and a console error would be shown continouusly until we publish this release.
4. He updates to the latest version of the tracker.
If all the above happen, he will start seeing some video events from the 'first' video element.
 
closes #1116

### DIFF
--- a/plugins/browser-plugin-media-tracking/src/buildMediaEvent.ts
+++ b/plugins/browser-plugin-media-tracking/src/buildMediaEvent.ts
@@ -9,7 +9,7 @@ export function buildMediaEvent(
   conf: TrackingOptions,
   detail?: EventDetail
 ): MediaEventData {
-  const context = [getHTMLMediaElementEntities(el), getMediaPlayerEntities(el, detail)];
+  const context = [getHTMLMediaElementEntities(el, conf), getMediaPlayerEntities(el, detail)];
   if (el instanceof HTMLVideoElement) context.push(getHTMLVideoElementEntities(el as HTMLVideoElement));
   const data: MediaPlayerEvent = { type: e in eventNames ? eventNames[e] : e };
   if (conf.label) data.label = conf.label;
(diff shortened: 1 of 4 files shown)


snowplowcla:

plugins/browser-plugin-media-tracking/src/findMediaElement.ts
```diff
@@ -25,6 +25,10 @@ function findMediaElementChild(el: Element): SearchResult {
     if (elem.length === 1) {
       if (isAudioElement(elem[0])) return { el: elem[0] };
       if (isVideoElement(elem[0])) return { el: elem[0] };
+    } else if (elem.length === 2 && tag === 'VIDEO' && isVideoElement(elem[0])) {
```
igneel64, line 28: If we want to make it really specific we can check for classnames in the container containing the string `jwplayer` etc.

paulboocock: This _might_ catch other players similar to jwplayer so I'd probably leave it generic and say a little prayer. Can always tighten it up if this leads to unexpected consequences for other players, seems pretty safe tho.

plugins/browser-plugin-media-tracking/src/buildMediaEvent.ts
```diff
@@ -40,9 +40,11 @@ function getMediaPlayerEntities(el: HTMLAudioElement | HTMLVideoElement, detail?
   };
 }
 
-function getHTMLMediaElementEntities(el: HTMLAudioElement | HTMLVideoElement): MediaEntities {
+function getHTMLMediaElementEntities(el: HTMLAudioElement | HTMLVideoElement, conf: TrackingOptions): MediaEntities {
+  // In cases where the media does not have explicit id, we use the container id passed from the configuration.
```
paulboocock, line 44: Nice idea!

paulboocock (approved):
LGTM

PR #1119, opened 2022-10-17
Use currentSrc as src when source element in used in media-tracking (closes #1118)

### LINKED ISSUE #1118: Support src coming from the source element in media-tracking
Currently the media-tracking plugin does correctly retrieve the media source from a setup using the `<source />` element.
E.g. for the below case, the `src` attribute will end up as empty which is not allowed on the media schema.
```html
 <audio id='audio-id'>
      <source
        src="https://snowplow.com/test.mp3"
        type="audio/mp3"
      />
</audio>
```

Support this case properly. We know that the `src` and `currentSrc` will end up the same in all those cases, but this is expected.

Changing media structures is daunting for a setup.

igneel64:
Use `currentSrc` as `src` when src is not available. This is the case where there is a `source` element under our audio or video we wish to track.

### Notes
- Unfortunately testing will require a bit more time since `jsdom` does not have the functionality to work properly with `currentSrc`

(closes #1118)

### DIFF
--- a/plugins/browser-plugin-media-tracking/src/buildMediaEvent.ts
+++ b/plugins/browser-plugin-media-tracking/src/buildMediaEvent.ts
@@ -56,7 +56,7 @@ function getHTMLMediaElementEntities(el: HTMLAudioElement | HTMLVideoElement): M
     readyState: READY_STATE[el.readyState] as MediaElement['readyState'],
     seekable: timeRangesToObjectArray(el.seekable),
     seeking: el.seeking,
-    src: dataUrlHandler(el.src),
+    src: dataUrlHandler(el.src || el.currentSrc),
     textTracks: textTrackListToJson(el.textTracks),
     fileExtension: el.currentSrc.split('.').pop() as string,
     fullscreen: isElementFullScreen(el.id),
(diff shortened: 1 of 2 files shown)


snowplowcla:

plugins/browser-plugin-media-tracking/src/buildMediaEvent.ts
```diff
@@ -56,7 +56,7 @@ function getHTMLMediaElementEntities(el: HTMLAudioElement | HTMLVideoElement): M
     readyState: READY_STATE[el.readyState] as MediaElement['readyState'],
     seekable: timeRangesToObjectArray(el.seekable),
     seeking: el.seeking,
-    src: dataUrlHandler(el.src),
+    src: dataUrlHandler(el.src || el.currentSrc),
```
greg-el, line 59: Consider: It appears that `src` is in fact already optional in the [schema](https://github.com/snowplow/iglu-central/blob/be84568d2dda1c272a15fb49cc63a916c180dd64/schemas/org.whatwg/media_element/jsonschema/1-0-0#L247) (it isn't in `required`), so I'm wondering if a null fallback is worth putting in here too, to prevent the empty string issue. I'm not quite sure if there's a situation where `src` and `currentSrc` will both be empty strings though.

greg-el (approved):
LGTM! Just a comment for your consideration, but this seems like a good solution.

plugins/browser-plugin-media-tracking/src/buildMediaEvent.ts
igneel64: That's a great observation! 
In that case we indeed would want the schema to fail. IMO `src` should have been required so it could signify problematic cases.

PR #1123, opened 2022-10-24
Remove fixed values for playbackQuality in YT integration tests plus some type fixes

### LINKED ISSUE #1122: End to end tests failing at `getPlaybackQuality` for Youtube tracking.
**Describe the bug**
Currently tests for the youtube tracking plugin [youtube.test.ts](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/trackers/javascript-tracker/test/integration/youtube.test.ts#L44) and specifically the `playbackQuality` attribute, seems to be restricting the available values expected of the `getPlaybackQuality` YT iFrame API.

This leads to unexpected E2E test failures as the results might depend on changes in the YT iFrame implementation which is external. E.g. there are cases where a new value called `tiny` might appear. This is breaking the whole end to end suite.

Since this is an external API and the attribute on our youtube schema does not restrict the values to be sent on `playbackQuality`, we can be more lenient on the values we expect and save some end to end testing time.

Example build: https://github.com/snowplow/snowplow-javascript-tracker/actions/runs/3288427884/jobs/5437298652#step:10:7958

**Screenshots**
image

igneel64:
- Remove fixed test values for youtube integration tests.
- Add missing types in the Youtube interface

### Notes
- The `Youtube` interface is not exposed on the plugin typings (_or arguments_) so this can be regarded a safe change.

(closes #1122)

### DIFF
--- a/common/changes/@snowplow/browser-plugin-youtube-tracking/issue-1122-fix-youtube-e2e-playbackQuality-values_2022-10-24-08-31.json
+++ b/common/changes/@snowplow/browser-plugin-youtube-tracking/issue-1122-fix-youtube-e2e-playbackQuality-values_2022-10-24-08-31.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-plugin-youtube-tracking",
+      "comment": "Remove fixed values for playbackQuality in YT integrations plus type fixes (closes #1122)",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/browser-plugin-youtube-tracking"
+}
--- a/common/changes/@snowplow/javascript-tracker/issue-1122-fix-youtube-e2e-playbackQuality-values_2022-10-24-08-31.json
+++ b/common/changes/@snowplow/javascript-tracker/issue-1122-fix-youtube-e2e-playbackQuality-values_2022-10-24-08-31.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/javascript-tracker",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/javascript-tracker"
+}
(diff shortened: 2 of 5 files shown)


snowplowcla:

plugins/browser-plugin-youtube-tracking/src/index.ts
```diff
@@ -243,6 +243,8 @@ function youtubeEvent(player: YT.Player, eventName: string, conf: TrackingOption
   }
 
   const event = buildYouTubeEvent(player, eventName, conf, eventData);
+  /* FIXME: Convert CommonEventProperties interface to accept generics for SelfDescribingJson */
+  // @ts-expect-error
```
paulboocock, line 247: Can we not add an index signature to `interface YouTube` (https://github.com/snowplow/snowplow-javascript-tracker/blob/0dbff387084ea714b75c770dbd8fc0781b6d6981/plugins/browser-plugin-youtube-tracking/src/contexts.ts#L95) so that is can assigned to `Record<string, unknown>`? Seems nicer to me than ignoring the TS error and it matches the other interfaces like `interface SnowplowMediaPlayer` (https://github.com/snowplow/snowplow-javascript-tracker/blob/0dbff387084ea714b75c770dbd8fc0781b6d6981/plugins/browser-plugin-youtube-tracking/src/types.ts#L53).

You could even add `[key: string]: string | string[] | boolean | number | number[] | null | undefined;` if you really wanted to 😂 but I'd be tempted to just go for `[key: string]: unknown;`

igneel64: The thing is that `[key: string]: unknown;` is kneecapping TypeScript and imo it should be used mostly in cases where our library provides APIs on the clients with uknown requirements. I cannot see the benefit to do this on our interfaces, especially the ones with standard properties. 

The kneecap is visible just by the fact that we were missing an attribute and TS did not even complain. That is `avaliableQualityLevels` which is fixed in this PR.

The good practise here is mentioned on the FIXME comment above the error expectation. The only reason I chose not to do the fix in this PR is it would make it a bit large and out of scope for the fix.

WDYT ? _hope I understood your comment_

paulboocock: I find TODOs and FIXMEs a little triggering, and would prefer to see it fixed properly if we need to. I appreciate adding the unknown index type isn't a wonderful solution. I'd be a bit fan of adding generics in (in a non breaking change if possible) and I don't mind a little scope creep as its starting to feel like the right time to solve this. The more new plugins we add that track specific entities (ecom, consent, etc) the more we'll likely need this.

igneel64: Will add in a PR for v4. Removing the change here.

matus-tomlein (approved):
LGTM!

PR #1132, opened 2022-12-22
Update browser targets for compilation (closes #1131)

### LINKED ISSUE #1131: Update Browser/JavaScript compilation target for v4
Planning for the version 4 major update, there would be the expected bump in the browser support for our libraries.

The motivation for this change is to allow for our users to enjoy the benefits of more modern JavaScript and allow us to take advantage of more recent but mature browser capabilities when building new features.

igneel64:
Based on our internal data and [open source information](https://caniuse.com/usage-table) the first proposal for the v4 compilation targets is the one presented.

### Highlights
- Dropping support for IE9 and IE10
- Minimum Safari is 9.1
- 'Evergreen' type browsers are bumped

### Notes
- Our compiled artifacts do not seem to change after the bump.


matus-tomlein (approved):
LGTM!

Jack-Keene (approved):
LGTM!

.browserslistrc
```diff
@@ -1,5 +1,5 @@
-chrome >= 32,
-ie >= 9,
-edge >= 13,
-firefox >= 27,
-safari >= 8,
+chrome >= 72,
+ie >= 11,
+edge >= 100,
```
matus-tomlein, line 3: Nit: Sorry for going back to this later, for some reason I opened this again and a thought came to my mind whether we should keep the Chrome and Edge versions in sync? Edge 100 was released in March 2022 which seems a bit too recent (I understand that it is evergreen and doesn't show up in our event sample, but makes more sense to me to handle it the same way as Chrome since it is basically the same).

igneel64: Yeap was a bit skeptical on this one too. 

- The main reason for not matching them is based on our data sources, both the internal and the OS information, which are different for the two browsers.
- Both browsers have the same upstream but different release schedules so they are bound to mismatch.

Trying to find a benefit for pulling it back.

matus-tomlein: From what I could find out, it seems to me that Edge X is built on Blink X (it appears to be that way [based on this table](https://en.wikipedia.org/wiki/Microsoft_Edge#New_Edge_release_history)) and the same is true for Chrome.

The benefit could be that we can include a larger pool of browsers without doing any more work – since we supporting Chrome since version 72, my understanding is that we don't need to check anything else to claim that we also support Edge 79.

igneel64: Edge and Chrome have the same OS base which is Chromium _(they have the same upstream)_. The rendering engine for Chromium is Blink.

The release schedule for Chrome and Edge even if converges at some point, can change independent of each other. 
- Edge https://learn.microsoft.com/en-us/deployedge/microsoft-edge-release-schedule#release-cadence
- Chrome has changed the release schedule from six weeks to four (1 day ago some news came up again, [link](https://developer.chrome.com/blog/early-stable/) 😆 )

Yeah I see your point, I will find the edge version corresponding, in time period, with Chrome 72 and update accordingly.

paulboocock (approved):
LGTM. Out of scope for this PR, but there are a number of places where we can also now improve the code by removing IE<11.

A couple that quickly spring to mind:
- The out queue, no need for Image any more, it can all be XHR.
- Integration Tests ;-)
- Bumping UUID library to latest

I'm also a little surprised the output didn't change. The underlying library uses this file to figure out if typescript should transpile to ES3 or ES5. It always picked ES3, but I'd have thought we'd have ended up with ES5 now. I can't quite remember how, but using the TS lib we have and running it in node, you can feed this file into it and it outputs the TS version it selects.

igneel64:
@paulboocock 
About ES3-ES5, even with the old settings we were in ES5 supported environments https://caniuse.com/es5 .
Did you find any hint that this compiled down to ES3 in the past ?

PR #1134, opened 2022-12-22
Add plugin to send requests with user ID to a Kantar FocalMeter endpoint (close #1133)

### LINKED ISSUE #1133: Add plugin to send requests with user ID to a Kantar FocalMeter endpoint
This will integrate the tracker with the FocalMeter system by Kantar that measures audience of content through a router meter.

The tracker should make a request with the domain user ID whenever that changes. This will be implemented through a plugin that will check the ID in the events in the `afterTrack` callback. The URL endpoint to send the requests to should be configurable when initializing the plugin.

matus-tomlein:
Issue #1133

Implements a plugin for integration with [Kantar FocalMeter](https://www.virtualmeter.co.uk/focalmeter) system for measuring audience of content using a router on local network.

The solution works by sending an HTTP request with the current domain user ID to a configurable Kantar endpoint. This enables the FocalMeter to associate the traffic from the device with the Snowplow ID.

The request is made from a plugin in the `afterTrack` callback. It checks whether the user ID in the trackerd event changed from the last observed one and if so, it makes the request. The Kantar endpoint to make the request to is passed when creating the plugin.

There is also an option to store the last observed user ID in local storage so that the same request doesn't need to be made on each page load. However, I decided to make the local storage disabled by default because I assume that sometimes it is necessary to make the request with the ID multiple times. For example, in case one loads the page in a place where they don't have the Kantar router but then go to a different place with the router, we should make the request in the second place as well.

### DIFF
--- a/plugins/browser-plugin-focalmeter/package.json
+++ b/plugins/browser-plugin-focalmeter/package.json
@@ -1,0 +1,53 @@
+{
+  "name": "@snowplow/browser-plugin-focalmeter",
+  "version": "3.7.0",
+  "description": "Kantar FocalMeter integration for Snowplow",
+  "homepage": "http://bit.ly/sp-js",
+  "bugs": "https://github.com/snowplow/snowplow-javascript-tracker/issues",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/snowplow/snowplow-javascript-tracker.git"
+  },
+  "license": "BSD-3-Clause",
+  "author": "Matus Tomlein",
+  "sideEffects": false,
+  "main": "./dist/index.umd.js",
+  "module": "./dist/index.module.js",
+  "types": "./dist/index.module.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "rollup -c --silent --failAfterWarnings",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@snowplow/browser-tracker-core": "workspace:*",
+    "@snowplow/tracker-core": "workspace:*",
+    "tslib": "^2.3.1"
+  },
+  "devDependencies": {
+    "@ampproject/rollup-plugin-closure-compiler": "~0.27.0",
+    "@rollup/plugin-commonjs": "~21.0.2",
+    "@rollup/plugin-node-resolve": "~13.1.3",
+    "@types/jest": "~27.4.1",
+    "@types/jsdom": "~16.2.14",
+    "@typescript-eslint/eslint-plugin": "~5.15.0",
+    "@typescript-eslint/parser": "~5.15.0",
+    "eslint": "~8.11.0",
+    "jest": "~27.5.1",
+    "jest-environment-jsdom": "~27.5.1",
+    "jest-environment-jsdom-global": "~3.0.0",
+    "jest-standard-reporter": "~2.0.0",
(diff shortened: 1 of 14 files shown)


plugins/browser-plugin-focalmeter/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+ * The plugin sends requests with the domain user ID to a Kantar endpoint used with the FocalMeter system.
+ * A request is made when the first event with a new user ID is tracked.
+ *
+ * @param kantarEndpoint URL of the Kantar endpoint to send the requests to (including protocol)
+ * @param useLocalStorage Whether to store information about the last submitted user ID in local storage to prevent sending it again on next load
+ * @param logLevel Log message level
+ * @returns The plugin
+ */
+export function FocalMeterPlugin(
+  kantarEndpoint: string,
```
igneel64, line 52: ❓ 
Can we make this accept an options object instead of positional arguments ? 
IMO it will help with the maintenance of the plugin and adding new things if needed _(probably this might not need any updates but I would consider it a good practise)_. WDYT

matus-tomlein: Sounds like a good idea! I'll change it.

matus-tomlein: Added the `FocalMeterConfiguration` type to wrap the options. Also added an `enableFocalMeterIntegration` function to pass the configuration object since that is how it is done in other plugins (e.g., form tracking). Also passing the options when creating the plugin would not be possible in the JS tag interface which does not have a way to do that, so using the enable function seems better.

plugins/browser-plugin-focalmeter/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "rollup": "~2.70.1",
+    "rollup-plugin-cleanup": "~3.2.1",
+    "rollup-plugin-license": "~2.6.1",
+    "rollup-plugin-terser": "~7.0.2",
+    "rollup-plugin-ts": "~2.0.5",
+    "ts-jest": "~27.1.3",
+    "typescript": "~4.6.2"
+  },
+  "peerDependencies": {
+    "@snowplow/browser-tracker": "~3.7.0"
```
igneel64, line 51: I think this should be bumped into `3.8.0` . 
Btw could we maybe use `workspace:*` in the peerDependencies too.

plugins/browser-plugin-focalmeter/src/index.ts
```diff
@@ -0,0 +1,156 @@
...
+      _configurations[id] = configuration;
+    }
+  }
+}
+
+function getLocalStorageKey(trackerId: string): string {
+  return `sp-fclmtr-${trackerId}`;
+}
+
+function sendRequest(url: string, userId: string, LOG: Logger, successCallback: () => void): void {
```
igneel64, line 124: At some point this should be extracted _(just leaving the comment here as a thought moving forward)_

plugins/browser-plugin-focalmeter/package.json
matus-tomlein: Good catch! I changed it to 3.8.0. Not sure about using the `workspace:*` in peerDependencies, all the other plugins specify the version so didn't want to do something else here. Also it should be managed automatically on release so not a big problem.

PR #1135, opened 2022-12-23
Add client_session context entity if anonymous tracking with session tracking is enabled (close #1124)

### LINKED ISSUE #1124: Add client_session context entity if anonymous tracking with session tracking is enabled
In case one creates tracker with the following configuration:

```js
snowplow('newTracker', 'sp', '{{collector_url_here}}', {
  ...
  anonymousTracking: { withSessionTracking: true },
  contexts: {
    session: true
  }
);
```

The session context entity is not added to events even though sessions are tracked.

This is because of the [check on this line](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/libraries/browser-tracker-core/src/tracker/index.ts#L790) which doesn't allow the session entity to be added in case anonymous tracking is enabled.

Session entity should be added if anonymous tracking with session tracking is enabled. However, the `userId` property in the entity should be a null UUID: `00000000-0000-0000-0000-000000000000`

matus-tomlein:
Issue #1124

Adds the `client_session` context entity to events even in case anonymous tracking with session tracking is enabled. That means with this configuration:

```js
snowplow('newTracker', 'sp', '{{collector_url_here}}', {
  ...
  anonymousTracking: { withSessionTracking: true },
  contexts: { session: true }
);
```

In that case, the `userId` and `previousSessionId` properties of the entity are anonymised (we do the same on the mobile trackers). The `userId` is assigned a zero UUID (`00000000-0000-0000-0000-000000000000`) while the `previousSessionId` is null.

### DIFF
--- a/libraries/browser-tracker-core/test/tracker/session_data.test.ts
+++ b/libraries/browser-tracker-core/test/tracker/session_data.test.ts
@@ -139,6 +139,7 @@ describe('Tracker API: ', () => {
             let context = payload.co as string;
             expect(context).toContain('client_session');
             expect(context).toContain('"userId":"00000000-0000-0000-0000-000000000000"');
+            expect(context).toContain('"previousSessionId":null');
             done();
           },
         },
--- a/libraries/browser-tracker-core/test/tracker/session_data.test.ts
+++ b/libraries/browser-tracker-core/test/tracker/session_data.test.ts
@@ -28,6 +28,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+import { TrackerConfiguration } from '../../dist/index.module';
 import { addTracker, SharedState } from '../../src';
 import { createTestIdCookie, createTestSessionIdCookie } from '../helpers';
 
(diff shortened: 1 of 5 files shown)


libraries/browser-tracker-core/test/tracker/session_data.test.ts
```diff
@@ -102,10 +103,72 @@ describe('Tracker API: ', () => {
...
+  it('Adds the client session context entity when anonymous session tracking', (done) => {
+    const tracker = createTracker({
+      contexts: { session: true },
+      encodeBase64: false,
+      anonymousTracking: { withSessionTracking: true },
+      plugins: [
+        {
+          afterTrack: (payload) => {
+            let context = payload.co as string;
+            expect(context).toContain('client_session');
```
igneel64, line 140: Could we also check the previous session id property ?

libraries/browser-tracker-core/test/tracker/session_data.test.ts
```diff
@@ -102,10 +103,72 @@ describe('Tracker API: ', () => {
...
+          },
+        },
+      ],
+    });
+
+    tracker?.trackPageView();
+  });
 });
+
+function createTracker(configuration?: TrackerConfiguration) {
```
igneel64, line 171: 👍

igneel64 (approved):

PR #1143, opened 2023-01-24
Make sure brands array in client hints plugin is serialized as an array in context entity (close #1082)

### LINKED ISSUE #1082: Make sure arrays retrieved from browser APIs are serialized as arrays in context entities
**Describe the bug**
The client hints plugin can produce invalid `http_client_hints` context entity due to assigning the `brands` array as an object in case browsers return such value from `navigator.userAgentData.brands`. This is unexpected behaviour from the browser but does seem to happen in some cases.

For example, the brands property may look like this: `"brands":{"0":{"brand":" Not A;Brand","version":"99"}}` instead of `"brands":[{"brand":" Not A;Brand","version":"99"}]`.

**To Reproduce**
Unfortunately, I can't provide clear instructions on how to reproduce this as it only happens under certain cases. However, we do have examples of such badly formed events.

**Expected behavior**
The `brands` property should be an array instead of the object.

**Suggested solution**
We should process the value returned from `navigator.userAgentData.brands` into an array to make sure that it is an array, e.g. using:

```js
let brands = Object.keys(navigator.userAgentData.brands).map(function(e) {
  return navigator.userAgentData.brands[e];
})
```

There might be other cases where the same problem occurs. So we should also check for those.

matus-tomlein:
Issue #1082

The PR addresses the problem where the `brands` array in the `http_client_hints` context entity produced by the client hints plugin was malformed. In some cases reported in the issue, it was in the form of an object instead of an array. I suspect that this is due to some browser bugs or other influences but I wasn't able to replicate the issue myself.

The fix just converts the `brands` object or array into an array by iterating over its values. Since I wasn't able to replicate the problem, I can't verify it helps but it should be able to convert an object to an array.

### DIFF
--- a/plugins/browser-plugin-client-hints/src/index.ts
+++ b/plugins/browser-plugin-client-hints/src/index.ts
@@ -57,6 +57,16 @@ declare global {
 let uaClientHints: HttpClientHints;
 
 /**
+ * This function makes sure that the expected array is returned as an array instead of an object.
+ * It handles a problem that in some cases the `navigator.userAgentData.brands` was returned as an object instead of array.
+ */
+function forceArray<T>(array: T[]): T[] {
+  return Object.keys(array).map((e) => {
+    return (array as any)[e];
+  });
+}
+
+/**
  * Attaches Client Hint information where available
  * @param includeHighEntropy - Should high entropy values be included
  */
@@ -67,7 +77,7 @@ export function ClientHintsPlugin(includeHighEntropy?: boolean): BrowserPlugin {
     if (navigatorAlias.userAgentData) {
       uaClientHints = {
         isMobile: navigatorAlias.userAgentData.mobile,
-        brands: navigatorAlias.userAgentData.brands,
+        brands: forceArray(navigatorAlias.userAgentData.brands),
       };
       if (includeHighEntropy && navigatorAlias.userAgentData.getHighEntropyValues) {
         navigatorAlias.userAgentData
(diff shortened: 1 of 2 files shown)


igneel64 (approved):

plugins/browser-plugin-client-hints/src/index.ts
```diff
@@ -56,6 +56,16 @@ declare global {
...
+/**
+ * This function makes sure that the expected array is returned as an array instead of an object.
+ * It handles a problem that in some cases the `navigator.userAgentData.brands` was returned as an object instead of array.
+ */
+function forceArray<T>(array: T[]): T[] {
+  return Object.keys(array).map((e) => {
+    return (array as any)[e];
+  });
+}
+
```
paulboocock, line 68: Do you have an example of one of the malformed objects? I'll be honest, I'm struggling to imagine a malformed object that passes through this and ends up in the same format as the expected array of objects.

matus-tomlein: Yes, we do have an example, should have put it in the PR description:

```json
{ "0":{"brand":" Not A;Brand","version":"99"}, "1":{"brand":"Chromium","version":"102"}, "2":{"brand":"Google Chrome","version":"102"} }
```

It was [reported here](https://github.com/snowplow/snowplow-javascript-tracker/issues/1082#issuecomment-1162392887).

paulboocock: I should have looked at the issue 🤦

paulboocock: Looks good. If you want to make TypeScript happy, you could do something like this:

```
function forceArray<T>(array: T[] | Record<string, T>): T[] {
  if (Array.isArray(array)) return array;

  return Object.keys(array).map((e) => {
    return array[e];
  });
}
```

paulboocock (approved):

plugins/browser-plugin-client-hints/src/index.ts
matus-tomlein: Looks better, thanks!

PR #1144, opened 2023-01-25
Change default tracker initialization parameters for JS/Browser v4

igneel64:
This PR includes the proposed changes for the JS/Browser tracker initialization parameters:

Currently it includes:
- `discoverRootDomain`: The default becomes `true`
- `configCookieSameSite`: The default becomes `Lax`

A bit more info:

### `discoverRootDomain`
On the previous version this value defaulted to `false`. 

This had an important effect when used as a default. If the tracker was installed in a subdomain (_foo.example.com_), or any number of subdomains _(bar.example.com etc),_ and a TLD _(example.com)_, then all instances had different domain userIds.

That would make the analysis a bit more challenging when trying to count unique users, without the existance of user ID _(`uid`),_ and cross TLD-subdomain behaviours/attribution.

The behaviour now mimics what other tools like Google Analytics have as a default.

### `configCookieSameSite`
This was set as the recommended value of `Lax` but defaults to `None` which seems to have caused some issues for unsupported browsers and tracking use cases. 
Since our v4 compatibility does not allow us to follow the changes made in latest browser versions, this would be a good default to follow for this version:
More info:
- https://snowplow.io/blog/understanding-the-samesite-cookie-update/
- https://www.chromium.org/updates/same-site/incompatible-clients/

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/feature-v4-change-default-initialization-params-browser_2023-01-26-08-53.json
+++ b/common/changes/@snowplow/browser-tracker-core/feature-v4-change-default-initialization-params-browser_2023-01-26-08-53.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
+      "comment": "Change default tracker initialization params (discoverRootDomain, cookieSameSite)",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/browser-tracker-core"
+}
--- a/common/changes/@snowplow/javascript-tracker/feature-v4-change-default-initialization-params-browser_2023-01-26-08-53.json
+++ b/common/changes/@snowplow/javascript-tracker/feature-v4-change-default-initialization-params-browser_2023-01-26-08-53.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/javascript-tracker",
+      "comment": "",
+      "type": "none"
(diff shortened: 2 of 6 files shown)


matus-tomlein (approved):
LGTM, nice change! 👍 

I can't comment in the file directly, but can you fix the API comments in the `types.ts` file? It still says that the default for `discoverRootDomain` is false and for `cookieSameSite` is None.

igneel64:
Sure thanks!

paulboocock:
On `discoverRootDomain`, I've never been a big fan of how it tries to keep setting cookies. This usually leads of console warnings, which is always unfortunate. I'd prefer to not have a default value which commonly produces warnings.

However, I do see the benefit in making the default cookie domain the root... maybe we can try and find it another way - trim `window.location` to the typical value and try to set the cookie? Maybe a little regex solves this. Worst case, we fallback to cycle through the other options if we don't find the root straight away?

igneel64:
> On `discoverRootDomain`, I've never been a big fan of how it tries to keep setting cookies. This usually leads of console warnings, which is always unfortunate. I'd prefer to not have a default value which commonly produces warnings.
> 
> However, I do see the benefit in making the default cookie domain the root... maybe we can try and find it another way - trim `window.location` to the typical value and try to set the cookie? Maybe a little regex solves this. Worst case, we fallback to cycle through the other options if we don't find the root straight away?

Addressed at https://github.com/snowplow/snowplow-javascript-tracker/pull/1163

PR #1145, opened 2023-01-30
v4: Revise Node.js initialization

igneel64:
Node.js tracker API revision:

## API revision
This is a v4 revision of the Node.js API. 

**From**
```ts
function tracker( 
  emitters: Emitter | Array<Emitter>, 
  namespace: string, 
  appId: string, 
  encodeBase64: boolean )
: Tracker
```
**To**
```ts
function tracker(
  trackerConfiguration: NodejsTrackerConfiguration,
  emitterConfiguration: NodejsEmitterConfiguration
): Tracker 
```

It mainly removes the need to initialize the `gotEmitter` separately and now allows all the available options under the `NodejsEmitterConfiguration` options.
If clients want to initialize a different or more than one emitter, they can use the `customEmitter` option.

```ts
// v3
import { tracker, gotEmitter } from '@snowplow/node-tracker';

const e = gotEmitter(
  'collector.mydomain.net', // Collector endpoint
  snowplow.HttpProtocol.HTTPS,
  snowplow.HttpMethod.POST
);

const t = tracker(e, 'myTracker', 'myApp', false);

// v4
import { tracker } from '@snowplow/node-tracker';
const t = tracker(
  { namespace: 'myTracker', appId: 'myApp', encodeBase64: false }, 
  { endpoint:  'collector.mydomain.net' }
);

// v4 multiple (got)emitters
import { tracker } from '@snowplow/node-tracker';
const t = tracker(
  { namespace: 'myTracker', appId: 'myApp', encodeBase64: false }, 
  [ { endpoint:  'collector.mydomain.net' } ,  { endpoint:  'collector.mydomain.net' } ]
);

// v4 custom emitter
import { tracker } from '@snowplow/node-tracker';
const t = tracker(
  { namespace: 'myTracker', appId: 'myApp', encodeBase64: false }, 
  { customEmitter: () => (/* Return custom emitter/s */) }
);
```

## Exposed fields
- `gotEmitter` is no longer exported.
- Extraneous interfaces e.g. HTTPMethod are no longer exported.

## Notes
- Default configuration changes
  - Protocol defaults to `https`
  - Method defaults to `post`
- Some TS magic to only allow `customEmitter` when available and not other options, to avoid confusion.
- Upgrade ava to @5

### DIFF
--- a/trackers/node-tracker/src/tracker.ts
+++ b/trackers/node-tracker/src/tracker.ts
@@ -31,6 +31,7 @@
 import { trackerCore, PayloadBuilder, TrackerCore, version } from '@snowplow/tracker-core';
 
 import { Emitter } from './emitter';
+import { gotEmitter, GotEmitterConfiguration } from './got_emitter';
 
 export interface Tracker extends TrackerCore {
   /**
@@ -48,31 +49,40 @@ export interface Tracker extends TrackerCore {
   setNetworkUserId: (userId: string) => void;
 }
 
-/**
- * Snowplow Node.js Tracker
- *
- * @param string - or array emitters The emitter or emitters to which events will be sent
- * @param string - namespace The namespace of the tracker
- * @param string - appId The application ID
- * @param boolean - encodeBase64 Whether unstructured events and custom contexts should be base 64 encoded
- */
-export function tracker(
-  emitters: Emitter | Array<Emitter>,
-  namespace: string,
-  appId: string,
-  encodeBase64: boolean
-): Tracker {
-  let domainUserId: string;
-  let networkUserId: string;
-  let allEmitters: Array<Emitter>;
+interface TrackerConfiguration {
+  /* The namespace of the tracker */
+  namespace: string;
+  /* The application ID */
+  appId: string;
+  /* Whether unstructured events and custom contexts should be base64 encoded. */
+  encodeBase64: boolean;
+}
 
-  if (Array.isArray(emitters)) {
-    allEmitters = emitters;
+type CustomEmitter = {
+  /* Function returning custom Emitter or Emitter[] to be used. If set, other options are irrelevant */
+  customEmitter: () => Emitter | Array<Emitter>;
+};
+
+type EmitterConfiguration = CustomEmitter | GotEmitterConfiguration;
+
+export function newTracker(
+  trackerConfiguration: TrackerConfiguration,
+  emitterConfiguration: EmitterConfiguration | EmitterConfiguration[]
+): Tracker {
+  const { namespace, appId, encodeBase64 = true } = trackerConfiguration;
+
+  let allEmitters: Emitter[] = [];
+  if (Array.isArray(emitterConfiguration)) {
+    allEmitters = emitterConfiguration.map((config) => gotEmitter(config as GotEmitterConfiguration));
+  } else if (emitterConfiguration && emitterConfiguration.hasOwnProperty('customEmitter')) {
+    const customEmitters = (emitterConfiguration as CustomEmitter).customEmitter();
+    allEmitters = Array.isArray(customEmitters) ? customEmitters : [customEmitters];
   } else {
-    allEmitters = [emitters];
+    allEmitters = [gotEmitter(emitterConfiguration as GotEmitterConfiguration)];
   }
 
-  encodeBase64 = encodeBase64 !== false;
+  let domainUserId: string;
+  let networkUserId: string;
 
   const addUserInformation = (payload: PayloadBuilder): void => {
     payload.add('duid', domainUserId);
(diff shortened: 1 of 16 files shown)


trackers/node-tracker/src/tracker.ts
```diff
@@ -48,31 +49,33 @@ export interface Tracker extends TrackerCore {
...
+
+type NodejsEmitterConfiguration = CustomEmitter | GotEmitterConfiguration;
+
 export function tracker(
-  emitters: Emitter | Array<Emitter>,
-  namespace: string,
-  appId: string,
-  encodeBase64: boolean
+  trackerConfiguration: NodejsTrackerConfiguration,
+  emitterConfiguration: NodejsEmitterConfiguration
```
matus-tomlein, line 70: Wouldn't it be better to also accept an array of `NodejsEmitterConfiguration` here instead of only through the `customEmitter` function? Because now I think users can't easily make multiple got emitter instances like they could before – if they want to have multiple emitters, they need to implement custom ones.

matus-tomlein (approved):
Looks very nice, so much better than before!

Btw, NodeJS 18 supports the fetch API so there is a hope that we might be able to drop the got dependency in the future and make the NodeJS tracker work in any JS environment 🤩  Given this configuration API, it should be just a matter of swapping the internals.

trackers/node-tracker/src/tracker.ts
```diff
@@ -48,31 +49,33 @@ export interface Tracker extends TrackerCore {
...
 
-/**
- * Snowplow Node.js Tracker
- *
- * @param string - or array emitters The emitter or emitters to which events will be sent
- * @param string - namespace The namespace of the tracker
- * @param string - appId The application ID
- * @param boolean - encodeBase64 Whether unstructured events and custom contexts should be base 64 encoded
- */
+interface NodejsTrackerConfiguration {
```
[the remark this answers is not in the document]
igneel64: The reason I added the additional name is mostly about making it faster to search/autocomplete and discern in the monorepo. Nothing else in particular.
WDYT ?

trackers/node-tracker/src/tracker.ts
matus-tomlein: But you said that the `gotEmitter` function is no longer exported. If it is, then I have no problem with this. I was just worried that if people want to have multiple emitters, they need to implement their own emitter from scratch.

trackers/node-tracker/src/tracker.ts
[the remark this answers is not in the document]
matus-tomlein: I understand that point that but it still looks quite ugly to me :D But it's no big deal, just a nitpick.

trackers/node-tracker/src/tracker.ts
igneel64: Yes the `gotEmitter` function is not exported now _(I left the comment `/* For the client, this would be any kind of function they want */` to make it clear that this would be custom made)._

> if people want to have multiple emitters, they need to implement their own emitter from scratch

Yes that is the case, but we provide the interface and point them to our own implementation. They can even copy the code and use it exactly as stated above. I considered this a fair change to remove. We can provide this convenience but in my view it adds more complication to our API rather than a clear benefit.

I might be wrong since I might be missing past context, but is this a commonly used feature ?

matus-tomlein: But the complication from our side is like adding a couple of extra lines of code while for the user it's adding maybe 100 lines so it's quite disproportionate.

I don't think the feature is used much but I also don't have any insight into how people use the tracker. Also our other trackers don't have the same feature. I am just arguing for it because some users of v3 might have adopted it and will be sad when they upgrade to v4. But I don't have a strong opinion either way.

trackers/node-tracker/src/tracker.ts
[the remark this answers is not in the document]
paulboocock: I'd go for TrackerConfiguration. It's cleaner and more concise, still describes what it does and the import shows it's from the "node-tracker". 

You could rename it on import or use a global import if you wanted it "namespaced".

trackers/node-tracker/src/tracker.ts
paulboocock: This is one of those features which when its used, its a strong requirement, but it is used quite rarely. Initialising multiple trackers is miserable, so I think keeping a way of creating two "core" emitters is a feature we need to leave in.

We could make it `emitterConfiguration: EmitterConfiguration | Array<EmitterConfiguration>`. It doesn't make our side much worse, but it gives great flexibility for the end user. 

In reality no one is going to build their own emitter, it's probably a risky venture for most and they'd prefer us to deal with the issues and bugs. So if anything, putting the work into this over customEmitter is probably more important (although leaving both in is good too for flexibility).

+1 for not exporting `gotEmitter` thought and moving to `EmitterConfiguration`s.

trackers/node-tracker/src/tracker.ts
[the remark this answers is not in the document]
igneel64: Fixed as well!

PR #1147, opened 2023-02-06
Revise default plugins included on the JavaScript tracker

igneel64:
Revise the 'by default' included plugins on `sp.js`. _(Reduction is about 1.5kb)_

- Remove by default:
  - ecommerce
  - enhanced ecommerce
  - consent
  - optimizelyX
- Add by default
  - enhanced consent
  - snowplow ecommerce
- Add deprecation warning on ecommerce/consent packages since they will be replaced by `snowplow ecommerce` and `enhanced consent`.

_This will be paired with changes in the docs as well_

### DIFF
--- a/plugins/browser-plugin-consent/README.md
+++ b/plugins/browser-plugin-consent/README.md
@@ -3,6 +3,10 @@
 [![npm version][npm-image]][npm-url]
 [![License][license-image]](LICENSE)
 
+<h3>⚠️ This package is deprecated, please use <a href="https://www.npmjs.com/package/@snowplow/browser-plugin-enhanced-consent">@snowplow/browser-plugin-enhanced-consent</a> instead. ⚠️</h3>
+
+---
+
 Browser Plugin to be used with `@snowplow/browser-tracker`.
 
 Adds consent information to your Snowplow tracking.
--- a/plugins/browser-plugin-ecommerce/README.md
+++ b/plugins/browser-plugin-ecommerce/README.md
@@ -3,6 +3,10 @@
 [![npm version][npm-image]][npm-url]
 [![License][license-image]](LICENSE)
 
+<h3>⚠️ This package is deprecated, please use <a href="https://www.npmjs.com/package/@snowplow/browser-plugin-snowplow-ecommerce">@snowplow/browser-plugin-snowplow-ecommerce</a> instead. ⚠️</h3>
+
+---
+
 Browser Plugin to be used with `@snowplow/browser-tracker`.
 
 Adds ecommerce events to your Snowplow tracking.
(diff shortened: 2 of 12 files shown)


plugins/browser-plugin-consent/README.md
```diff
@@ -3,6 +3,10 @@
 [![npm version][npm-image]][npm-url]
 [![License][license-image]](LICENSE)
 
+<h3>⚠️ This package will soon be deprecated, please use <a href="https://www.npmjs.com/package/@snowplow/browser-plugin-enhanced-consent">@snowplow/browser-plugin-enhanced-consent</a> instead. ⚠️</h3>
```
paulboocock, line 6: This package _is_ now deprecated. Just because it's deprecated doesn't mean we remove it yet, but they we simply won't be updating it going forward.

plugins/browser-plugin-ecommerce/README.md
```diff
@@ -3,6 +3,10 @@
 [![npm version][npm-image]][npm-url]
 [![License][license-image]](LICENSE)
 
+<h3>⚠️ This package will soon be deprecated, please use <a href="https://www.npmjs.com/package/@snowplow/browser-plugin-snowplow-ecommerce">@snowplow/browser-plugin-snowplow-ecommerce</a> instead. ⚠️</h3>
```
paulboocock, line 6: Same here

trackers/javascript-tracker/tracker.config.ts
```diff
@@ -28,23 +28,26 @@
...
+
+/* By default excluded plugins in sp.js */
 export const browserFeatures = false;
+export const consent = false;
+export const ecommerce = false;
+export const enhancedEcommerce = false;
 export const mediaTracking = false;
+export const optimizely = false;
+export const optimizelyX = false;
 export const youtubeTracking = false;
```
paulboocock, line 53: I like this separation. I wonder if it's worth making a "Included", "Excluded", "Deprecated" sections for anyone who does stumble into this file.

I'd say:
browserFeatures, consent, ecommerce, enhancedEcommerce and optimisely (X can stay undeprecated but non-default) are all (and if not should be) deprecated now.

plugins/browser-plugin-consent/README.md
igneel64: Yeap, was a bit hesitant on the language because I thought of not using `npm deprecate`.

I now think we shoud do that for the list you mention on the other comment below. WDYT ?

paulboocock: Yeah, go for it. The ones we're suggesting are either super old or we have better replacements.

igneel64: Done.
The deprecation will happen along with the release through `npm` directly.

trackers/javascript-tracker/tracker.config.ts
igneel64: Done

matus-tomlein (approved):
LGTM!

PR #1152, opened 2023-02-09
Fix flaky end-to-end tests

matus-tomlein:
The tests seem much less flaky now – in the last build only 1 test failed and had to be retried.

**Dependency updates**

* updated all Webdriver dependencies to the latest version and deal with breaking changes
* update Micro to v1.4

**Changes in tests**

* skip IE in activity tracking and integration tests because of flakiness
* for the youtube test also skip IE 11 and Safari 12 (kept crashing) and do not check `percentprogress` events under Safari 14
* simplified the `beforeAll` in youtube tests not to poll Micro to check for tracked events since this caused failures that were difficult to understand from test results – it's nicer when we see the failures in the specific test cases

**Updated browsers used in tests**

* Remove legacy browsers no longer available on Saucelabs: Safari 8, IE 9, Chrome 60
* Remove Firefox 96 on Windows, keep on macOS
* Add Chrome 109
* Remove Safari 13, keep 12 and 14 (just to have less browsers to test)
* Changes to reflect our v4 targets:
  * Remove Firefox 55 and add 78
  * Remove Edge 13
  * Remove IE 10

**TODO (in a future PR)**

* Update Firefox to 109 – need to deal with some failing tests
* Also update Safari to 16 – but I suspect there are some bugs on Saucelabs because this resulted in some empty tests where nothing happened but they passed
* Investigate why are `percentprogress` events not tracked on Safari
* Fix running the tests locally


greg-el (approved):
Looks good! Nice to see the little green tick is back :smile:

Running integration tests locally still seems to be having a bit of trouble, but I think we can address that in another PR. Getting them more reliable in CI is a great step.

igneel64 (approved):

PR #1155, opened 2023-02-15
Accept `YT.Player` instance in `enableYouTubeTracking` (close #1141)

### LINKED ISSUE #1141: YouTube plugin can't track players created with IFrame API
Internally, [we create a `YT.Player`](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/plugins/browser-plugin-youtube-tracking/src/index.ts#L190) for the youtube video iframe, and then wait for the `onReady` callback to fire, so we can be sure the player is ready to have event listeners attached. If a youtube video is created using the API, the `onReady` will have likely fired before the plugin has a chance to create its own `YT.Player` for that iframe, meaning the event listeners are never attached.

We need to look into whether there's a reliable way to detect if a player is "ready" without always relying on the `onReady` callback.

greg-el:
We currently only accept a string id for an Iframe to signify which element to track and attach our listeners on the `onReady` callback provided by the player.

This doesn't work if a user has already created an instance of `YT.Player` on their Iframe, as the `onReady` callback is never fired. To mitigate this we can accept an instance of `YT.Player` in our `id` field.


plugins/browser-plugin-youtube-tracking/src/index.ts
```diff
@@ -66,9 +66,13 @@ function trackEvent(
   });
 }
 
-export function trackingOptionsParser(mediaId: string, conf?: MediaTrackingOptions): TrackingOptions {
+export function trackingOptionsParser(id: string | YT.Player, conf?: MediaTrackingOptions): TrackingOptions {
+  const player: YT.Player | undefined = typeof id === 'string' ? undefined : id;
```
igneel64, line 70: I think the compiler understands the type without `: YT.Player | undefined`
Something like: `const player = typeof id === 'string' ? undefined : id;` would suffice

igneel64 (commented):
Looks good! Could we get an additional example on the `javascript-tracker-examples` for this one ?

PR #1163, opened 2023-03-03
Fix root domain algorithm initial attempt

### LINKED ISSUE #1162: Fix top level domain grouping initial cookie set attempt
**Describe the bug**
The algorithm behind the `discoverRootDomain` functionality checks if the domain is valid to set a cookie at the highest level domain. This algorithm currently starts from a eTLD which will not be valid in any case, e.g. `.com`, `.io` etc.

This behaviour will create a console warning every time in some browsers e.g. Firefox
image

The fix would just be to start trying from the TLD and not the eTLD.

**Note**
Even with the fix, if we have an eTLD with more than one part e.g. co.uk, then the warning will still be there. This happens with other trackers as well e.g. analytics.js

igneel64:
close #1162


matus-tomlein (approved):
This is a very nice fix!

PR #1170, opened 2023-04-03
Add Google Analytics 4 and Universal Analytics Ecommerce translation layer

igneel64:


miike:
This looks good to me, adding @jethron to have a look through too.

plugins/browser-plugin-snowplow-ecommerce/src/ga4/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function transformG4ItemsToSPProducts({
+  items,
+  currency = 'USD',
+  categorySeparator = '/',
+}: GA4ItemTransformation): Product[] {
+  return items.map((ga4Item) => {
+    const { item_category, item_category2, item_category3, item_category4, item_category5 } = ga4Item;
+    const category = [item_category, item_category2, item_category3, item_category4, item_category5]
+      .filter(Boolean)
```
jethron, line 18: I thought I was the only one that did this. :heart:

plugins/browser-plugin-snowplow-ecommerce/src/ua/api.ts
```diff
@@ -0,0 +1,98 @@
...
+  });
+  trackProductView({ ...products[0] });
+}
+
+export function trackEEProductListView(ecommerce: UAEnhancedEcommerceObject & ProductListViewEvent) {
+  const products = transformUAProductsToSPProducts({
+    currencyCode: ecommerce.currencyCode,
+    products: ecommerce.impressions,
+  });
+  trackProductListView({ name: ecommerce.impressions[0].list as string, products });
```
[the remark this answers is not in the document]
jethron: Also I'd generally prefer `val!` to `val as string` as I think the intent is clearer: (despite the type we really expect this to be defined at this point) vs (shut up typescript, this is a string no matter what I said before)

jethron (commented):
Looks like handy wrappers, thanks @igneel64! I think the single object arg vs multiple args is the only big thing I'd change. Maybe currency after that.

plugins/browser-plugin-snowplow-ecommerce/src/ua/types.ts
```diff
@@ -0,0 +1,90 @@
...
+      tax?: number;
+      shipping?: number;
+      coupon?: string;
+      affiliation?: string;
+    };
+    products: EEProduct[];
+  };
+}
+
+export interface EEProduct {
```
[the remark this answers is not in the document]
igneel64: I would not prefer to cross import just to keep things a bit more straightforward. 

The definitions from the `browser-plugin-enhanced-ecommerce` are there to support the full API of EE. For the translation layers, I kept the typings to the attributes we can support. E.g. we do not support `coupon` at the product level yet.

plugins/browser-plugin-snowplow-ecommerce/src/ua/api.ts
```diff
@@ -0,0 +1,98 @@
...
+  PurchaseEvent,
+  RemoveFromCartEvent,
+  UAEnhancedEcommerceObject,
+} from './types';
+
+import { transformUAProductsToSPProducts } from './utils';
+
+const DEFAULT_PAYMENT_METHOD = 'unknown';
+
+export function trackEEPromoView(ecommerce: UAEnhancedEcommerceObject & PromoViewEvent) {}
```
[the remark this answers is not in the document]
igneel64: Will ask @matus-tomlein for an additional opinion as I was torn as well.

plugins/browser-plugin-snowplow-ecommerce/src/ga4/api.ts
```diff
@@ -0,0 +1,70 @@
...
+  trackRemoveFromCart,
+} from '../api';
+import { Currency, GA4EcommerceObject, ItemList, Transaction } from './types';
+import { transformG4ItemsToSPProducts } from './utils';
+
+const DEFAULT_PAYMENT_METHOD = 'unknown';
+
+export function trackGA4ViewPromotion(ecommerce: GA4EcommerceObject) {}
+export function trackGA4SelectPromotion(ecommerce: GA4EcommerceObject) {}
+export function trackGA4ViewCart() {}
```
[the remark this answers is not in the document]
igneel64: Added :)

plugins/browser-plugin-snowplow-ecommerce/src/ua/api.ts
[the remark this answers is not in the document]
matus-tomlein: I don't really have a strong opinion but I agree with the point to make it consistent EE plugin so would prefer `trackEnhancedEcommercePromoView`. `EE` as an abbreviation is probably not very well understood for people not very familiar with the tracker.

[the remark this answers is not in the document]
igneel64: Done.

plugins/browser-plugin-snowplow-ecommerce/src/ga4/api.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export function trackGA4AddToCart(ecommerce: GA4EcommerceObject & Currency, finalCartValue: number) {
```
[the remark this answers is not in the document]
igneel64: Yeap, as advertised in the accelerator page, this is a 'quick' way to migrate your most important GA ecommerce events to snowplow using these wrappers. If you are at a level of custom contexts on ecommerce events, then I think it is a good place to start the native migration.

_I think this would be the point where the user would like to add their custom dimensions in a context, sent together with the standard snowplow ecommerce attributes._

Kinda torn again, it would be really easy make the second argument an `options` object which would include the required attributes, where needed e.g. `finalCartValue`, and the addition of custom contexts and tracker selection.

plugins/browser-plugin-snowplow-ecommerce/src/ua/api.ts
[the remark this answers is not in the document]
igneel64: In my view if the list is empty on the `trackEnhancedEcommerceProductListView` call, then I would think it should fail as a bad event on the pipeline. This can then unearth any product list issues that were previously uknown to the user due to GA not indicating an otherwise invalid event.

On the note of `val!` I agree as well. Fixed

plugins/browser-plugin-snowplow-ecommerce/src/ga4/api.ts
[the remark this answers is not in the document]
igneel64: @jethron 
Added a simple options argument which for now reference the currency where required and any other attribute which is required in some cases. I believe that for now it is good enough for the simple translation layer and the extra contexts etc. could be a ticket to the native implementation.

plugins/browser-plugin-snowplow-ecommerce/src/ua/api.ts
```diff
@@ -0,0 +1,114 @@
...
+  opts: Options
+) {
+  const currency = (ecommerce.currencyCode || opts.currency)!;
+  const products = transformUAProductsToSPProducts(ecommerce.impressions, currency);
+  trackProductListView({ name: ecommerce.impressions[0].list!, products });
+}
+
+export function trackEnhancedEcommerceProductListClick(
+  ecommerce: UAEnhancedEcommerceObject & ProductListClickEvent,
+  opts: Options
```
jethron, line 85: For `trackEnhancedEcommerceProductDetail`, `trackEnhancedEcommerceProductListView` and `trackEnhancedEcommerceProductListClick`, `opts` being undefined may throw JS errors, should we just default it to empty object or something so it's easy to use if you need it, but won't break the code if you don't?

igneel64: Yeap fixed, thanks :)

plugins/browser-plugin-snowplow-ecommerce/src/ua/api.ts
```diff
@@ -0,0 +1,114 @@
...
+  currency?: string;
+};
+
+const DEFAULT_PAYMENT_METHOD = 'unknown';
+
+export function trackEnhancedEcommercePromoView(ecommerce: UAEnhancedEcommerceObject & PromoViewEvent) {
+  const promotions = transformUAPromotionsToSPPromotions(ecommerce.promoView.promotions);
+  /* Enhanced Ecommerce allows multiple promotions to be sent on a single event. */
+  for (let i = 0; i < promotions.length; i++) {
+    trackPromotionView(promotions[i]);
```
jethron, line 38: Think this can just be a `forEach`?

plugins/browser-plugin-snowplow-ecommerce/src/ua/api.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+}
+
+export function trackEnhancedEcommercePromoClick(ecommerce: UAEnhancedEcommerceObject & PromoClickEvent) {
+  const [promotion] = transformUAPromotionsToSPPromotions(ecommerce.promoClick.promotions);
```
jethron, line 43: Would this ever have >1 promotion that we'd be throwing away here? Should we just `forEach` that too?

plugins/browser-plugin-snowplow-ecommerce/src/ua/api.ts
```diff
@@ -0,0 +1,114 @@
...
+  trackPromotionClick(promotion);
+}
+
+export function trackEnhancedEcommerceAddToCart(
+  ecommerce: UAEnhancedEcommerceObject & AddToCartEvent,
+  opts: Options & { finalCartValue: number }
+) {
+  const currency = (ecommerce.currencyCode || opts.currency)!;
```
jethron, line 51: I think now that we have this opts arg for ergonomics, we can blame the user if it falls back incorrectly. So I think we're now safe to return to your old version with like `const DEFAULT_CURRENCY_CODE = "USD";` or whatever as a default and remove the `!`. `const currency = ecommerce.currencyCode || opts.currency || DEFAULT_CURRENCY_CODE;`

plugins/browser-plugin-snowplow-ecommerce/src/ua/utils.ts
```diff
@@ -0,0 +1,29 @@
+import { Product, Promotion } from '../types';
+import { EEProduct, EEPromo } from './types';
+
+export function transformUAProductsToSPProducts(products: EEProduct[], currency: string): Product[] {
+  return products.map((product) => {
+    return {
+      currency,
+      /* Id or name is required, but we in SP-plugin we require id to be present */
+      id: product.id! || product.name!,
```
jethron, line 9: If we're sure one of the values is defined we shouldn't need the `!`?

plugins/browser-plugin-snowplow-ecommerce/src/ga4/api.ts
```diff
@@ -0,0 +1,99 @@
...
+export function trackGA4SelectPromotion(ecommerce: GA4EcommerceObject & Promotion) {
+  const promotion = transformGA4PromotionToSPPromotion(ecommerce);
+  trackPromotionClick(promotion);
+}
+
+export function trackGA4AddToCart(
+  ecommerce: GA4EcommerceObject & Currency,
+  opts: Options & { finalCartValue: number }
+) {
+  const currency = (ecommerce.currency || opts.currency)!;
```
jethron, line 39: Same as for UA, I think now that we have `opts` it's easy enough to specify a currency if you don't have it in `ecommerce` already, so having an explicit fallback default should be OK now and we don't need to risk undefined currency with `!`.

jethron (commented):
_Almost_ there... :stuck_out_tongue:

plugins/browser-plugin-snowplow-ecommerce/src/ua/api.ts
jethron: Oh derp, will need a wrapper function because of the `trackers` param, this is probably fine, then.

plugins/browser-plugin-snowplow-ecommerce/src/ua/api.ts
igneel64: Based on the API and the interaction it should not, it is just served as an array. https://developers.google.com/analytics/devguides/collection/ua/gtm/enhanced-ecommerce#promo-clicks

plugins/browser-plugin-snowplow-ecommerce/src/ua/api.ts
igneel64: Sure!

plugins/browser-plugin-snowplow-ecommerce/src/ga4/api.ts
igneel64: Same comment as with GA.

plugins/browser-plugin-snowplow-ecommerce/src/ua/utils.ts
igneel64: The `product` in this case comes from the type `EEProduct` which has both `id` and `name` as optional. 

The type inference of the expression in the `id` attribute results in possible `string | undefined` which is not valid for our _(Snowplow Ecommerce)_ API. 

Instead of creating a type helper chose to be explicit about the inference of "one of the two will be there", using the `!` mark. How would you go about it ?

jethron: I would probably change the type and define `EEProduct` as `({id: string; name?: string} | {id?: string: name: string}) & {...}` to more closely match its actual type. But I think this is fine, I'm just being pedantic; we know it'll be fine but two `!` on a single line made me uncomfortable. :stuck_out_tongue:

plugins/browser-plugin-snowplow-ecommerce/src/ua/api.ts
jethron: Sounds good!

I feel like an official plugin for the official SDK should basically never send data to bad if we can help it, but I think since this is a quick migration path it's better to preserve the (lack of) data quality and it's acceptable in this case.

jethron (approved):

PR #1175, opened 2023-04-17
Fix newTracker typing to accurately return null or undefined (fix #1167)

### LINKED ISSUE #1167: Wrongly typed `newTracker()` function within the "@snowplow/browser-tracker" package
**Description**
The exposed `newTracker()` function from the [@snowplow/browser-tracker package](https://github.com/snowplow/snowplow-javascript-tracker/tree/master/trackers/browser-tracker) returns `BrowserTracker | null`. The current interface definition hints to a return type of `BrowserTracker` only.

This is misleading and does not represent the current implementation correctly.

**To Reproduce**
- Add a new tracker by calling `newTracker()` with a specified tracker identifier e.g. "my-test"
- Add another tracker by calling `newTracker()` with the same `trackerId`
- The function should have returned `null`

**Expected behavior**
I would expect that either the types will tell that the function could possibly return null.
Or that the function would return the already initialized tracker, even though  this could be misleading due to the function naming "new".

**"Screenshots"**
@snowplow/browser-tracker: `newTracker()`
https://github.com/snowplow/snowplow-javascript-tracker/blob/master/trackers/browser-tracker/src/index.ts#L59

@snowplow/browser-tracker-core: `addTracker()`
https://github1s.com/snowplow/snowplow-javascript-tracker/blob/master/libraries/browser-tracker-core/src/snowplow.ts#L88

**Additional context**
We implemented the browser tracker within a React ecosystem and run into an issue where we called `newTracker()` twice due to an unexpected rerender. We used the returned tracker to add further tracker configurations; but were not able to do so after the second render / initialization.
We prevent calling the `newTracker()` function twice; and realized while debugging that the provided interface / type definitions are currently wrong.

igneel64:
Fix correct typing issue for `newTracker` possibly returning:
- `null` when a tracker is already initialized with the same tracker id.
- `undefined` when we are not in a browser context.

(close #1167)

### DIFF
--- a/common/changes/@snowplow/browser-tracker/fix-1167-correct-newTracker-function-on-browser-tracker_2023-04-17-07-28.json
+++ b/common/changes/@snowplow/browser-tracker/fix-1167-correct-newTracker-function-on-browser-tracker_2023-04-17-07-28.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker",
+      "comment": "Fix newTracker typing to accurately return null or undefined (fix #1167)",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/browser-tracker"
+}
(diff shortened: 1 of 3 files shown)


igneel64:
@matus-tomlein 
Yeap had the same thought and was waiting for some feedback. Will point to v4

PR #1186, opened 2023-05-11
Add the idService option to the tracker configuration (close #1185)

### LINKED ISSUE #1185: Add an `idService` option
Add an `idService` option to the JavaScript tracker. This option will expect a string URL as input. This URL will be added to the queue and will be called using a GET method. 

This option is there to allow the service URL to be called in order to set any required identifiers e.g. extra cookies. 

This call will be added on the execution queue as any other kind of tracker request. This will make sure that there are no race conditions when setting/retrieving identifiers between tracker and idService requests.

igneel64:
_This was first released in a dev version._
More in #1185

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/index.ts
+++ b/libraries/browser-tracker-core/src/tracker/index.ts
@@ -302,7 +302,8 @@ export function Tracker(
         trackerConfiguration.customHeaders ?? {},
         trackerConfiguration.withCredentials ?? true,
         trackerConfiguration.retryStatusCodes ?? [],
-        (trackerConfiguration.dontRetryStatusCodes ?? []).concat([400, 401, 403, 410, 422])
+        (trackerConfiguration.dontRetryStatusCodes ?? []).concat([400, 401, 403, 410, 422]),
+        trackerConfiguration.idService
       ),
       // Whether pageViewId should be regenerated after each trackPageView. Affect web_page context
       preservePageViewId = false,
--- a/libraries/browser-tracker-core/src/tracker/out_queue.ts
+++ b/libraries/browser-tracker-core/src/tracker/out_queue.ts
@@ -63,6 +63,7 @@ export interface OutQueue {
  * @param withCredentials - Sets the value of the withCredentials flag on XMLHttpRequest (GET and POST) requests
  * @param retryStatusCodes – Failure HTTP response status codes from Collector for which sending events should be retried (they can override the `dontRetryStatusCodes`)
  * @param dontRetryStatusCodes – Failure HTTP response status codes from Collector for which sending events should not be retried
+ * @param idService - Id service full URL. This URL will be added to the queue and will be called using a GET method.
  * @returns object OutQueueManager instance
  */
 export function OutQueueManager(
@@ -81,7 +82,8 @@ export function OutQueueManager(
   customHeaders: Record<string, string>,
   withCredentials: boolean,
   retryStatusCodes: number[],
-  dontRetryStatusCodes: number[]
+  dontRetryStatusCodes: number[],
+  idService?: string
 ): OutQueue {
   type PostEvent = {
     evt: Record<string, unknown>;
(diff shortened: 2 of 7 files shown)


libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -81,16 +82,21 @@ export function OutQueueManager(
...
   };
 
+  type IdServiceCall = {
+    url: string;
+  };
+
   let executingQueue = false,
     configCollectorUrl: string,
-    outQueue: Array<PostEvent> | Array<string> = [];
+    outQueue: Array<string | PostEvent | IdServiceCall> = [];
```
jethron, line 99: I think this type change hides some assumptions other parts of the code make, per the below.

`PostEvent[] | string[]` > `(PostEvent | string | IdServiceCall)[]`

We might be better off just having a flag for if we've contacted the idService yet, and when we start processing the queue, and handle it if it's not set?

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -322,12 +332,16 @@ export function OutQueueManager(
...
+        url = outQueue[0].url;
+        xhr = initializeXMLHttpRequest(outQueue[0].url, false, sync);
+        numberToSend = 1;
+      } else if (postable(outQueue)) {
         url = configCollectorUrl;
         xhr = initializeXMLHttpRequest(url, true, sync);
         numberToSend = chooseHowManyToSend(outQueue);
       } else {
-        url = createGetUrl(outQueue[0]);
+        url = createGetUrl(outQueue[0] as string);
```
jethron, line 344: :grimacing: Migrating to a flag should let us remove the as casts, too.

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -301,7 +301,8 @@ export function Tracker(
         trackerConfiguration.customHeaders ?? {},
         trackerConfiguration.withCredentials ?? true,
         trackerConfiguration.retryStatusCodes ?? [],
-        (trackerConfiguration.dontRetryStatusCodes ?? []).concat([400, 401, 403, 410, 422])
+        (trackerConfiguration.dontRetryStatusCodes ?? []).concat([400, 401, 403, 410, 422]),
+        trackerConfiguration.idService || ''
```
matus-tomlein, line 305: Nit: the `|| ''` part is not necessary, right? It's an optional argument in the constructor.

libraries/browser-tracker-core/src/tracker/out_queue.ts
matus-tomlein: I agree that having a flag would make the implementation more straightforward. I think it wouldn't change the number of if conditions we have to make and would make the type of the event queue simpler.

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -222,7 +225,7 @@ export function OutQueueManager(
   }
 
   const postable = (queue: Array<PostEvent> | Array<string>): queue is Array<PostEvent> => {
-    return typeof queue[0] === 'object';
+    return typeof queue[0] === 'object' && 'evt' in queue[0];
```
matus-tomlein, line 228: This is probably no longer necessary:

```suggestion
    return typeof queue[0] === 'object';
```

igneel64: I added that as an extra sanity check of the POST type requests.

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -305,6 +308,18 @@ export function OutQueueManager(
 
     executingQueue = true;
 
+    if (idService && !idServiceCalled) {
+      const xhr = initializeXMLHttpRequest(idService, false, sync);
```
matus-tomlein, line 312: This looks good but should we add a connection timeout here like in the other collector requests? In case the ID service doesn't respond so that it doesn't block sending events forever.

igneel64: Added. Removed the status check and added the same `connectionTimeout` proeprty for the value.

matus-tomlein (approved):
LGTM!

matus-tomlein (approved):
LGTM!

jethron (approved):
One tiny thing that probably doesn't matter. LGTM!

PR #1190, opened 2023-05-16
Add browser-plugin-web-vitals to track Web Vitals metrics

### LINKED ISSUE #1189: Add plugin to track web vitals
Create a JavaScript tracker plugin to allow tracking of [Web Vitals](https://web.dev/vitals/) and other high level metrics related to measuring web page performance.

igneel64:
Add a new Web Vitals plugin!

close #1189

### DIFF
--- a/plugins/browser-plugin-web-vitals/README.md
+++ b/plugins/browser-plugin-web-vitals/README.md
@@ -1,0 +1,68 @@
+# Snowplow Web Vitals Tracking
+
+[![npm version][npm-image]][npm-url]
+[![License][license-image]](LICENSE)
+
+Browser Plugin to be used with `@snowplow/browser-tracker`.
+
+The plugin adds the capability to track web performance metrics categorized as [Web Vitals](https://web.dev/vitals/). These metrics are tracked with an event based on the [web_vitals schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/web_vitals/jsonschema/).
+
+## Maintainer quick start
+
+Part of the Snowplow JavaScript Tracker monorepo.  
+Build with [Node.js](https://nodejs.org/en/) (14 or 16) and [Rush](https://rushjs.io/).
+
+### Setup repository
+
+```bash
+npm install -g @microsoft/rush 
+git clone https://github.com/snowplow/snowplow-javascript-tracker.git
+rush update
+```
+
+## Package Installation
+
+With npm:
+
+```bash
+npm install @snowplow/browser-plugin-web-vitals
+```
+
+## Usage
+
+Initialize your tracker with the `WebVitalsPlugin`:
+
+```js
+import { newTracker } from '@snowplow/browser-tracker';
+import { WebVitalsPlugin } from '@snowplow/browser-plugin-web-vitals';
+
+newTracker('sp1', '{{collector}}', { plugins: [ WebVitalsPlugin(/* pluginOptions */) ] });
+
+/* 
+ * Available plugin options `WebVitalsPluginOptions`:
+ * {
+ *  loadWebVitalsScript: Should the plugin immediately load the Core Web Vitals measurement script from UNPKG CDN.
+ *  webVitalsSource: The URL endpoint the Web Vitals script should be loaded from. Defaults to the UNPKG CDN.
+ * } 
+ */
+```
+
+### Choosing a Web Vitals measurement source
+
+The default Web Vitals measurement script is loaded from the [UNPKG](https://www.unpkg.com/) CDN. This choice is chosen as a default but you should consider your own setup when choosing the script source. Selecting a script source from a CDN which might already be used in your website might save you from yet another connection startup time (_Queueing_,_DNS lookup_,_TCP_, _SSL_).
+
+Another reasonable choice could be [jsDelivr](https://cdn.jsdelivr.net/npm/web-vitals@3/dist/web-vitals.iife.js).
+
+## Copyright and license
+
+Licensed and distributed under the [BSD 3-Clause License](LICENSE) ([An OSI Approved License][osi]).
+
+Copyright (c) 2023 Snowplow Analytics Ltd.
+
+All rights reserved.
+
+[npm-url]: https://www.npmjs.com/package/@snowplow/browser-plugin-web-vitals
+[npm-image]: https://img.shields.io/npm/v/@snowplow/browser-plugin-web-vitals
+[docs]: https://docs.snowplowanalytics.com/docs/collecting-data/collecting-from-own-applications/javascript-tracker/
+[osi]: https://opensource.org/licenses/BSD-3-Clause
+[license-image]: https://img.shields.io/npm/l/@snowplow/browser-plugin-web-vitals
--- a/plugins/browser-plugin-web-vitals/package.json
+++ b/plugins/browser-plugin-web-vitals/package.json
@@ -1,0 +1,54 @@
+{
+  "name": "@snowplow/browser-plugin-web-vitals",
+  "version": "3.11.0",
+  "description": "Adds the capability to track web performance metrics categorized as Web Vitals.",
+  "homepage": "https://github.com/snowplow/snowplow-javascript-tracker",
+  "bugs": "https://github.com/snowplow/snowplow-javascript-tracker/issues",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/snowplow/snowplow-javascript-tracker.git"
+  },
+  "license": "BSD-3-Clause",
+  "author": "Snowplow Analytics Ltd (https://snowplow.io/)",
+  "sideEffects": false,
+  "main": "./dist/index.umd.js",
+  "module": "./dist/index.module.js",
+  "types": "./dist/index.module.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "rollup -c --silent --failAfterWarnings",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@snowplow/browser-tracker-core": "workspace:*",
+    "@snowplow/tracker-core": "workspace:*",
+    "tslib": "^2.3.1",
+    "web-vitals": "~3.3.2"
+  },
+  "devDependencies": {
+    "@ampproject/rollup-plugin-closure-compiler": "~0.27.0",
+    "@rollup/plugin-commonjs": "~21.0.2",
+    "@rollup/plugin-node-resolve": "~13.1.3",
+    "@types/jest": "~27.4.1",
+    "@types/jsdom": "~16.2.14",
+    "@typescript-eslint/eslint-plugin": "~5.15.0",
+    "@typescript-eslint/parser": "~5.15.0",
+    "eslint": "~8.11.0",
+    "jest": "~27.5.1",
+    "jest-environment-jsdom": "~27.5.1",
+    "jest-environment-jsdom-global": "~3.0.0",
+    "jest-standard-reporter": "~2.0.0",
+    "rollup": "~2.70.1",
+    "rollup-plugin-cleanup": "~3.2.1",
+    "rollup-plugin-license": "~2.6.1",
+    "rollup-plugin-terser": "~7.0.2",
+    "rollup-plugin-ts": "~2.0.5",
+    "ts-jest": "~27.1.3",
+    "typescript": "~4.6.2"
+  },
+  "peerDependencies": {
+    "@snowplow/browser-tracker": "~3.11.0"
+  }
+}
--- a/plugins/browser-plugin-web-vitals/src/index.ts
+++ b/plugins/browser-plugin-web-vitals/src/index.ts
@@ -1,0 +1,74 @@
+import { BrowserPlugin, BrowserTracker, dispatchToTrackersInCollection } from '@snowplow/browser-tracker-core';
+import { buildSelfDescribingEvent } from '@snowplow/tracker-core';
+import { WEB_VITALS_SCHEMA } from './schemata';
+import { attachWebVitalsPageListeners, createWebVitalsScript, webVitalsListener } from './utils';
+
+const _trackers: Record<string, BrowserTracker> = {};
+const WEB_VITALS_SOURCE = 'https://unpkg.com/web-vitals@3/dist/web-vitals.iife.js';
+let listenersAttached = false;
+
+interface WebVitalsPluginOptions {
+  loadWebVitalsScript?: boolean;
+  webVitalsSource?: string;
+}
+
+const defaultPluginOptions = {
+  loadWebVitalsScript: true,
+  webVitalsSource: WEB_VITALS_SOURCE,
+};
+
+/**
+ * Adds Web Vitals measurement events
+ *
+ * @param pluginOptions.loadWebVitalsScript - Should the plugin immediately load the Core Web Vitals measurement script from UNPKG CDN.
+ * @param pluginOptions.webVitalsSource - The URL endpoint the Web Vitals script should be loaded from. Defaults to the UNPKG CDN.
+ * @remarks
+ */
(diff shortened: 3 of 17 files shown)


plugins/browser-plugin-web-vitals/src/index.ts
```diff
@@ -0,0 +1,120 @@
...
+        webVitalsListener(webVitalsObject);
+      }
+
+      function sendValues() {
+        dispatchToTrackersInCollection(Object.keys(_trackers), _trackers, (t) => {
+          t.core.track(
+            buildSelfDescribingEvent({
+              event: {
+                schema: WEB_VITALS_SCHEMA,
+                data: webVitalsObject,
```
jethron, line 90: Should we double check here that `webVitalsObject` actually has some properties set worth reporting?

plugins/browser-plugin-web-vitals/src/index.ts
```diff
@@ -0,0 +1,120 @@
...
+      const isLatestVisibilityChangeSupported = supportedEntryTypes.indexOf('layout-shift') !== -1;
+
+      if (isLatestVisibilityChangeSupported) {
+        const onVisibilityChange = () => {
+          if (document.visibilityState === 'hidden') {
+            sendValues();
+            removeEventListener('visibilitychange', onVisibilityChange, true);
+          }
+        };
+        window.addEventListener('visibilitychange', onVisibilityChange, true);
```
jethron, line 112: Bit awkward that we are using `addEventListener` from `browser-tracker-core` above but then not here. Is it just for symmetry with `removeEventListener`? For my OCD if that's the case can we specify `window.removeEventListener()` above to clarify that relationship?

plugins/browser-plugin-web-vitals/src/index.ts
```diff
@@ -0,0 +1,120 @@
...
+
+      // Safari does not fire "visibilitychange" on the tab close
+      // So we have 2 options: loose Safari data, or loose LCP/CLS that depends on "visibilitychange" logic.
+      // Current solution: if LCP/CLS supported, use `onHidden` otherwise, use `pagehide` to fire the callback in the end.
+      //
+      // More details: https://github.com/treosh/web-vitals-reporter/issues/3
+      const supportedEntryTypes = (PerformanceObserver && PerformanceObserver.supportedEntryTypes) || [];
+      const isLatestVisibilityChangeSupported = supportedEntryTypes.indexOf('layout-shift') !== -1;
+
+      if (isLatestVisibilityChangeSupported) {
```
jethron, line 105: If there are multiple trackers, won't these listeners attach multiple times and call `sendValues()` repeatedly?

plugins/browser-plugin-web-vitals/src/types.ts
```diff
@@ -0,0 +1,29 @@
+export interface ReportCallback {
+  (metric: Metric): void;
+}
+
+/**
+ * @interface Metric
+ * Partial metric interface from https://github.com/GoogleChrome/web-vitals
+ */
+interface Metric {
```
jethron, line 9: Should we import this type from `web-vitals` so a change has the chance to break our build? Or is the build-dep too heavy to be worth it?

plugins/browser-plugin-web-vitals/src/index.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Adds Web Vitals measurement events
+ *
+ * @param pluginOptions.loadWebVitalsScript - Should the plugin immediately load the Core Web Vitals measurement script from UNPKG CDN.
+ * @param pluginOptions.webVitalsSource - The URL endpoint the Web Vitals script should be loaded from.
+ * @remarks
+ */
+export function WebVitalsPlugin(pluginOptions: WebVitalsPluginOptions = defaultPluginOptions): BrowserPlugin {
+  const webVitalsObject: Record<string, unknown> = {};
```
jethron, line 56: Should the type of this be more explicit?

plugins/browser-plugin-web-vitals/src/index.ts
```diff
@@ -0,0 +1,120 @@
...
+const _trackers: Record<string, BrowserTracker> = {};
+
+declare global {
+  interface Window {
+    webVitals: {
+      onCLS(cb: ReportCallback): void;
+      onFID(cb: ReportCallback): void;
+      onLCP(cb: ReportCallback): void;
+      onINP(cb: ReportCallback): void;
+      onTTFB(cb: ReportCallback): void;
```
jethron, line 16: Is the omission of `onFCP` deliberate?

plugins/browser-plugin-web-vitals/README.md
```diff
@@ -0,0 +1,54 @@
+# Snowplow Web Vitals Tracking
+
+[![npm version][npm-image]][npm-url]
+[![License][license-image]](LICENSE)
+
+Browser Plugin to be used with `@snowplow/browser-tracker`.
+
+The plugin adds the capability to track web performance metrics categorized as [Web Vitals](https://web.dev/vitals/) and collect them to your warehouse.
```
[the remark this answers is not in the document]
igneel64: Killing it 🔪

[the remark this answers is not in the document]
igneel64: Adding
> These metrics are tracked with an event based on the [web_vitals schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/web_vitals/jsonschema/).

plugins/browser-plugin-web-vitals/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Usage
+
+Initialize your tracker with the `WebVitalsPlugin`:
+
+```js
+import { newTracker } from '@snowplow/browser-tracker';
+import { WebVitalsPlugin } from '@snowplow/browser-plugin-web-vitals';
+
+newTracker('sp1', '{{collector}}', { plugins: [ WebVitalsPlugin() ] });
```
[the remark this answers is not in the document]
igneel64: Added:
```js
/* 
 * Available plugin options `WebVitalsPluginOptions`:
 * {
 *  loadWebVitalsScript: Should the plugin immediately load the Core Web Vitals measurement script from UNPKG CDN.
 *  webVitalsSource: The URL endpoint the Web Vitals script should be loaded from.
 * } 
 */
```

plugins/browser-plugin-web-vitals/src/index.ts
igneel64: Yes, at the time it was deliberate, but I am now thinking it will not hurt that much adding it. And there could be some use cases for reporting it and observing it over time.

plugins/browser-plugin-web-vitals/src/index.ts
```diff
@@ -0,0 +1,120 @@
...
+    return (arg) => {
+      webVitalsObject[metricSchemaName] = arg.value;
+      webVitalsObject.navigationType = arg.navigationType;
+    };
+  }
+  window.webVitals.onCLS(addWebVitalsMeasurement('cls'));
+  window.webVitals.onFID(addWebVitalsMeasurement('fid'));
+  window.webVitals.onLCP(addWebVitalsMeasurement('lcp'));
+  window.webVitals.onINP(addWebVitalsMeasurement('inp'));
+  window.webVitals.onTTFB(addWebVitalsMeasurement('ttfb'));
```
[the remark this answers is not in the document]
igneel64: Yeap will check and add a Log.

plugins/browser-plugin-web-vitals/src/index.ts
igneel64: Added

plugins/browser-plugin-web-vitals/src/index.ts
igneel64: Removed all the calls to the helper, is not needed currently based on the browser targets we have.

plugins/browser-plugin-web-vitals/src/index.ts
```diff
@@ -0,0 +1,120 @@
...
+import { ReportCallback } from './types';
+import { WEB_VITALS_SCHEMA } from './schemata';
+import { dispatchToTrackersInCollection } from '@snowplow/browser-tracker-core';
+import { LOG, buildSelfDescribingEvent } from '@snowplow/tracker-core';
+
+const _trackers: Record<string, BrowserTracker> = {};
+
+declare global {
+  interface Window {
+    webVitals: {
```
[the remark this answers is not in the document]
igneel64: Even if the `loadWebVitalsScript` is false, it does not mean that we do not have acces to the `window.webVitals` interface. It means that the client did not choose to load the script with our 'automatic way'. 
Some pages might inject it at some other point, maybe even asynchronously through other means.

In case the `window.webVitals` is not present, we consider that another issue, that is why we added the:
`LOG.warn('The window.webVitals API is currently unavailable. web_vitals events will not be collected.');`
Hope this makes sense

plugins/browser-plugin-web-vitals/src/index.ts
igneel64: Fixed every listener to be attached once, even with multiple trackers initialized.

plugins/browser-plugin-web-vitals/src/types.ts
igneel64: Added it and confirmed that since we are using only type imports, the final bundle is not affected.

plugins/browser-plugin-web-vitals/src/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+      webVitalsObject[metricSchemaName] = arg.value;
+      webVitalsObject.navigationType = arg.navigationType;
+    };
+  }
+  if (!window.webVitals) {
+    LOG.warn('The window.webVitals API is currently unavailable. web_vitals events will not be collected.');
+    return;
+  }
+
+  (window.webVitals as WebVitals).onCLS(addWebVitalsMeasurement('cls'));
```
igneel64, line 65: Since we added the `web-vitals` package, there is already a definition of `window.webVitals` which does not include the `onXXX` methods on it, as it is supposed to be used as a regular NPM package.

We cannot use the `declare` way as we did previously since there is a declaration conflict.

plugins/browser-plugin-web-vitals/src/index.ts
jethron: Slightly funkier, but probably not worth it:

```typescript
export type WebVitalsPayload = {
  navigationType: Metric["navigationType"];
} & {
  [f in Lowercase<Metric["name"]>]: Metric["value"];
};
```

plugins/browser-plugin-web-vitals/src/utils.ts
jethron: Ouch, didn't realise this. :(

Can we alias it to at least be a bit neater?

```typescript
const webVitals = window.webVitals as WebVitals;
webVitals.onFID(addWebVitalsMeasurement('fid'));
...
```

plugins/browser-plugin-web-vitals/src/utils.ts
```diff
@@ -0,0 +1,71 @@
...
+  //
+  // More details: https://github.com/treosh/web-vitals-reporter/issues/3
+  const supportedEntryTypes = (PerformanceObserver && PerformanceObserver.supportedEntryTypes) || [];
+  const isLatestVisibilityChangeSupported = supportedEntryTypes.indexOf('layout-shift') !== -1;
+
+  if (isLatestVisibilityChangeSupported) {
+    const onVisibilityChange = () => {
+      if (document.visibilityState === 'hidden') {
+        callback();
+        removeEventListener('visibilitychange', onVisibilityChange, true);
```
jethron, line 21: Nit: can we use `window.addEventListener` / `window.removeEventListener` or `addEventListener` / `removeEventListener` consistently?

jethron (approved):
Thanks! Looks better with multiple trackers now I think. Some minor quibbles but LGTM even without them.

plugins/browser-plugin-web-vitals/src/index.ts
```diff
@@ -0,0 +1,120 @@
...
+export function WebVitalsPlugin(pluginOptions: WebVitalsPluginOptions = defaultPluginOptions): BrowserPlugin {
+  const webVitalsObject: Record<string, unknown> = {};
+  const options = { ...defaultPluginOptions, ...pluginOptions };
+  let trackerId: string;
+  return {
+    activateBrowserPlugin: (tracker) => {
+      trackerId = tracker.id;
+      _trackers[trackerId] = tracker;
+
+      if (options.loadWebVitalsScript) {
```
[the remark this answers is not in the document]
jethron: Would we? Isn't the plugin config shared between all trackers, so it will only ever load from a single source? I think this would just be like:

```typescript
      if (options.loadWebVitalsScript && !webVitalsScript) {
        webVitalsScript = createWebVitalsScript(options.webVitalsSource);
      }
```
wouldn't it?

Maybe I'm missing something. :confused:

plugins/browser-plugin-web-vitals/src/index.ts
```diff
@@ -0,0 +1,74 @@
+import { BrowserPlugin, BrowserTracker, dispatchToTrackersInCollection } from '@snowplow/browser-tracker-core';
+import { buildSelfDescribingEvent } from '@snowplow/tracker-core';
+import { WEB_VITALS_SCHEMA } from './schemata';
+import { attachWebVitalsPageListeners, createWebVitalsScript, webVitalsListener } from './utils';
+
+const _trackers: Record<string, BrowserTracker> = {};
+const WEB_VITALS_SOURCE = 'https://unpkg.com/web-vitals@3/dist/web-vitals.iife.js';
```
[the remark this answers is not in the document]
igneel64: Yeap adding a section on the README to explain the reasoning as well.

plugins/browser-plugin-web-vitals/src/utils.ts
igneel64: Done

plugins/browser-plugin-web-vitals/src/utils.ts
igneel64: Done

plugins/browser-plugin-web-vitals/README.md
```diff
@@ -0,0 +1,62 @@
...
+import { newTracker } from '@snowplow/browser-tracker';
+import { WebVitalsPlugin } from '@snowplow/browser-plugin-web-vitals';
+
+newTracker('sp1', '{{collector}}', { plugins: [ WebVitalsPlugin(/* pluginOptions */) ] });
+
+/* 
+ * Available plugin options `WebVitalsPluginOptions`:
+ * {
+ *  loadWebVitalsScript: Should the plugin immediately load the Core Web Vitals measurement script from UNPKG CDN.
+ *  webVitalsSource: The URL endpoint the Web Vitals script should be loaded from.
```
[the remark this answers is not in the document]
igneel64: Changed slightly!

matus-tomlein (approved):
LGTM!

matus-tomlein (approved):
LGTM!

plugins/browser-plugin-web-vitals/package.json
```diff
@@ -0,0 +1,54 @@
...
+  ],
+  "scripts": {
+    "build": "rollup -c --silent --failAfterWarnings",
+    "test": "jest"
+  },
+  "dependencies": {
+    "@snowplow/browser-tracker-core": "workspace:*",
+    "@snowplow/tracker-core": "workspace:*",
+    "tslib": "^2.3.1",
+    "web-vitals": "~3.3.2"
```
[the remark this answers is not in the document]
igneel64: The bundle is not affected since we only use types from the library. The reason for adding this to deps is a practise from https://github.com/microsoft/types-publisher/issues/81#issuecomment-234051338

PR #1192, opened 2023-05-17
Add onSessionUpdate callback configuration option

### LINKED ISSUE #1191: Add a onSessionUpdate callback parameter
Add a `onSessionUpdate` callback option which would allow the user to supply a callback and be notified any time a new session is created on the tracker. Similar to the [mobile one](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/mobile-trackers/tracking-events/session-tracking/#session-callback).

```js
newTracker('sp1', '{{collector_url}}', { 
  appId: 'my-app-id', 
  plugins: [ ],
  onSessionUpdate: callbackWithSessionInfo available.
})
```

igneel64:
Add the `onSessionUpdate` callback configuration option. 

The callback will retrieve the first argument as the `ClientSession` object type:

⚠️  Exposed `ClientSession` type as well.

```ts
export interface ClientSession extends Record<string, unknown> {
  /**
   * An identifier for the user of the session (same as domain_userid)
   */
  userId: string;

  /**
   * An identifier for the session (same as domain_sessionid)
   */
  sessionId: string;

  /**
   * The index of the current session for this user (same as domain_sessionidx)
   */
  sessionIndex: number;

  /**
   * Index of the current event in the session
   */
  eventIndex: number;

  /**
   * The previous session identifier for this user
   */
  previousSessionId: string | null;

  /**
   * The mechanism that the session information has been stored on the device
   */
  storageMechanism: string;

  /**
   * Identifier of the first event for this session
   */
  firstEventId: string | null;

  /**
   * Date-time timestamp of when the first event in the session was tracked
   */
  firstEventTimestamp: string | null;
}

```

close #1191

_Note_
- This dependency injection can change if we think of a better way.

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/id_cookie.ts
+++ b/libraries/browser-tracker-core/src/tracker/id_cookie.ts
@@ -30,6 +30,7 @@
 
 import { PayloadBuilder } from '@snowplow/tracker-core';
 import { v4 as uuid } from 'uuid';
+import { ClientSession } from './types';
 
 /**
  * Indices of cookie values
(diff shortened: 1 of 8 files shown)


libraries/browser-tracker-core/src/tracker/types.ts
```diff
@@ -581,3 +586,48 @@ export interface BrowserTracker {
    */
   addPlugin: (configuration: BrowserPluginConfiguration) => void;
 }
+
+/**
+ * Schema for client client session context entity
+ */
+export interface ClientSession extends Record<string, unknown> {
```
jethron, line 593: This `extends Record<string, unknown>` thing is a really weird requirement. We should make `PayloadBuilder.addContextEntity()` generic at some point.

```diff
-  addContextEntity: (entity: SelfDescribingJson) => void;
+  addContextEntity: <T extends Record<keyof T, unknown>>(entity: SelfDescribingJson<T>) => void;
```

libraries/browser-tracker-core/src/tracker/id_cookie.ts
```diff
@@ -240,6 +211,9 @@ export function startNewIdCookieSession(idCookie: ParsedIdCookie, memorizedVisit
   const sessionId = uuid();
   idCookie[sessionIdIndex] = sessionId;
 
+  onSessionUpdateCallback &&
+    onSessionUpdateCallback(clientSessionFromIdCookie(idCookie, configStateStorageStrategy, configAnonymousTracking));
```
jethron, line 215: Nit: Aesthetically I think I'd prefer an if statement here?

jethron (approved):
LGTM!

libraries/browser-tracker-core/src/tracker/types.ts
igneel64: Yes, I have been eyeing that for many months now.

libraries/browser-tracker-core/src/tracker/id_cookie.ts
igneel64: Will change! 🤗

matus-tomlein (approved):

PR #1194, opened 2023-05-18
Add gitleaks pre-commit hook

### LINKED ISSUE #1193: Detect leaked keys pre-commit using `gitleaks`
To increase the security of all our contributors credentials, we can opt for using the [gitleaks](https://github.com/gitleaks/gitleaks) library. By adding a `gitleaks` check on our pre-commit hook, we can add an extra shift-left security check.

igneel64:
Adding a first implementation of [gitleaks](https://github.com/gitleaks/gitleaks) pre-commit hook.

close #1193

### DIFF
--- a/common/git-hooks/pre-commit
+++ b/common/git-hooks/pre-commit
@@ -7,3 +7,23 @@
 # are committed. The command is defined in common/config/rush/command-line.json
 # and uses the "rush-prettier" autoinstaller.
 node common/scripts/install-run-rush.js prettier || exit $?
+
+ERROR_COLOR='\033[0;31m'
+WARNING_COLOR='\033[0;33m'
+STOP_COLOR='\033[0m'
+
+if ! command -v gitleaks &> /dev/null
+then
+   echo "${ERROR_COLOR}\nGitleaks not found. To commit you safely, you need to install it per the instructions at https://github.com/gitleaks/gitleaks.${STOP_COLOR}"
+   exit 1
+elif [ "$SKIP" == "gitleaks" ]
+then
+   echo "${WARNING_COLOR}\nGitleaks skipped for this commit.${STOP_COLOR}"
+else
+   gitleaks protect -v --staged --report-path ./findings.json
+fi
+
+# if [ "$SKIP" != "gitleaks" ]
+# then
+#    gitleaks protect -v --staged --report-path ./findings.json
+# fi
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,6 @@ trackers/javascript-tracker/test/local/serve/
 # IDE files
 .idea/
 .DS_Store
+
+# gitleaks file
+findings.json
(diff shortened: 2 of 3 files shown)


matus-tomlein (commented):
For some reason it doesn't stop me from commiting a secret in a file even though running `gitleaks protect` separately does show the problem – see the video. Do you have any idea what could be the problem here?

Also I am not a fan of pre-commit hooks – they are easy to not run and can be quite annoying for the user. Wouldn't it be better to run the `gitleaks detect` in CI instead? I understand that would mean the secret is already committed but I think it'd be good enough and would run for all commits no matter if the developer decides to skip the precommit hooks.


https://github.com/snowplow/snowplow-javascript-tracker/assets/42166/5cee2833-483f-4c4d-81a2-b3b53db30df7

findings.json
```diff
@@ -0,0 +1 @@
+[]
```
matus-tomlein, line 1: Nit: is it necessary to have the findings file in git? Just makes it a bit more crowded in the root folder. Maybe we can add it to gitignore?

matus-tomlein (approved):
Works after running `rush install`!

LGTM!

findings.json
igneel64: Yeap, good call! Doing that. _(for now `gitleaks protect` does not work with a baseline file so it does not help so much to be commited anyway)_

common/git-hooks/pre-commit
```diff
@@ -6,4 +6,8 @@
 # Invoke the "rush prettier" custom command to reformat files whenever they
 # are committed. The command is defined in common/config/rush/command-line.json
 # and uses the "rush-prettier" autoinstaller.
-node common/scripts/install-run-rush.js prettier || exit $?
\ No newline at end of file
+node common/scripts/install-run-rush.js prettier || exit $?
+if [ "$SKIP_GITLEAKS" != 1 ]
```
jethron, line 10: Should we be checking for `SKIP=gitleaks` instead, per their readme?

common/git-hooks/pre-commit
```diff
@@ -6,4 +6,8 @@
 # Invoke the "rush prettier" custom command to reformat files whenever they
 # are committed. The command is defined in common/config/rush/command-line.json
 # and uses the "rush-prettier" autoinstaller.
-node common/scripts/install-run-rush.js prettier || exit $?
\ No newline at end of file
+node common/scripts/install-run-rush.js prettier || exit $?
+if [ "$SKIP_GITLEAKS" != 1 ] 
+then
+   gitleaks protect -v --staged --report-path ./findings.json
```
jethron, line 12: This just fails without gitleaks installed. Can we detect that and issue a warning or suggest to skip the check if not found?

jethron (approved):
Minor issues, LGTM!

common/git-hooks/pre-commit
igneel64: The readme here refers to the way it would be disabled if the user installed it using [precommit](https://pre-commit.com/). 
But since it opened a comment, I will change it for consistency. Thanks!

common/git-hooks/pre-commit
igneel64: Made it a bit fancier as well

PR #1196, opened 2023-05-22
Round dimension type values on payload

### LINKED ISSUE #1195: Round width and height browser reported values.
In some cases e.g. subpixel rendering browsers might report fractional values for width and height dimensions of the document and/or window.

This might create issues on the reporting of these values _(e.g. errors in the pipeline)_ without providing any significant meaning that we can extract insight from. To fix this, we can select to round up/down to the closest integer.

igneel64:
Round dimension type values.

Notes:
- Made `addPayloadPair` more lenient as `payloadBuilder.add`

close #1195

### DIFF
--- a/libraries/browser-tracker-core/src/helpers/browser_props.ts
+++ b/libraries/browser-tracker-core/src/helpers/browser_props.ts
@@ -1,8 +1,11 @@
+/* Separator used for dimension values e.g. widthxheight */
+const DIMENSION_SEPARATOR = 'x';
+
 export function getBrowserProperties() {
   return {
-    viewport: detectViewport(),
-    documentSize: detectDocumentSize(),
-    resolution: screen.width + 'x' + screen.height,
+    viewport: floorDimensionFields(detectViewport()),
+    documentSize: floorDimensionFields(detectDocumentSize()),
+    resolution: floorDimensionFields(detectScreenResolution()),
     colorDepth: screen.colorDepth,
     devicePixelRatio: window.devicePixelRatio,
     cookiesEnabled: window.navigator.cookieEnabled,
@@ -35,7 +38,7 @@ function detectViewport() {
   }
 
   if (width >= 0 && height >= 0) {
-    return width + 'x' + height;
+    return width + DIMENSION_SEPARATOR + height;
   } else {
     return null;
   }
@@ -55,5 +58,19 @@ function detectDocumentSize() {
     bodyHeight = be ? Math.max(be.offsetHeight, be.scrollHeight) : 0;
   var w = Math.max(de.clientWidth, de.offsetWidth, de.scrollWidth);
   var h = Math.max(de.clientHeight, de.offsetHeight, de.scrollHeight, bodyHeight);
-  return isNaN(w) || isNaN(h) ? '' : w + 'x' + h;
+  return isNaN(w) || isNaN(h) ? '' : w + DIMENSION_SEPARATOR + h;
+}
+
+function detectScreenResolution() {
+  return screen.width + DIMENSION_SEPARATOR + screen.height;
+}
+
+export function floorDimensionFields(field?: string | null) {
+  return (
+    field &&
+    field
+      .split(DIMENSION_SEPARATOR)
+      .map((dimension) => Math.floor(Number(dimension)))
+      .join(DIMENSION_SEPARATOR)
+  );
 }
--- a/libraries/tracker-core/src/core.ts
+++ b/libraries/tracker-core/src/core.ts
@@ -155,7 +155,7 @@ export interface TrackerCore {
    * @param key - Field name
    * @param value - Field value
    */
-  addPayloadPair: (key: string, value: string | number) => void;
+  addPayloadPair: (key: string, value: unknown) => void;
 
   /**
    * Get current base64 encoding state
@@ -429,7 +429,7 @@ export function trackerCore(configuration: CoreConfiguration = {}): TrackerCore 
      * @param key - Field name
      * @param value - Field value
      */
-    function addPayloadPair(key: string, value: string | number): void {
+    function addPayloadPair(key: string, value: unknown): void {
       payloadPairs[key] = value;
     }
 
--- a/common/changes/@snowplow/browser-tracker-core/feature-1195-round-width-x-height-values_2023-05-22-13-17.json
+++ b/common/changes/@snowplow/browser-tracker-core/feature-1195-round-width-x-height-values_2023-05-22-13-17.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
+      "comment": "Round dimension type values on payload",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/browser-tracker-core"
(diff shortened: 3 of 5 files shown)


libraries/browser-tracker-core/src/helpers/browser_props.ts
```diff
@@ -55,5 +58,18 @@ function detectDocumentSize() {
...
   var h = Math.max(de.clientHeight, de.offsetHeight, de.scrollHeight, bodyHeight);
-  return isNaN(w) || isNaN(h) ? '' : w + 'x' + h;
+  return isNaN(w) || isNaN(h) ? '' : w + DIMENSION_SEPARATOR + h;
+}
+
+function detectScreenResolution() {
+  return screen.width + DIMENSION_SEPARATOR + screen.height;
+}
+
+export function floorDimensionFields(field?: string | null) {
```
matus-tomlein, line 68: Nit: this seems a bit cumbersome to go split the string after previously joining it in the same code. Could we instead define a function that would round the values before joining them? It could be called from the `detectDocumentSize`, `detectViewport` and `detectScreenResolution` functions instead of concatenating the values directly. I mean something like:

```ts
function toDimension(x: number, y: number): string {
   return Math.floor(x) + 'x' + Math.floor(y);
}
```

libraries/tracker-core/src/core.ts
```diff
@@ -429,7 +429,7 @@ export function trackerCore(configuration: CoreConfiguration = {}): TrackerCore
      * @param key - Field name
      * @param value - Field value
      */
-    function addPayloadPair(key: string, value: string | number): void {
+    function addPayloadPair(key: string, value: unknown): void {
```
matus-tomlein, line 432: Was this added intentionally in this PR? Doesn't seem related to the dimension issue.

igneel64: Yeap, it came up as the typings complained in the `resolution` result from the dimension values function return type. This showcases this inconsistency between  `addPayloadPair` and `payloadBuilder.add`. Added this in this PR as a small fix/housekeeping.

WDYT ?

matus-tomlein: Ah, I see! That makes sense.

matus-tomlein (approved):
LGTM, just one nitpick with my preference to round the values before joining them into a string.

libraries/browser-tracker-core/src/helpers/browser_props.ts
igneel64: There were a couple of options on my mind, one being this one.

My thought process went like this:
- It seems 'wasted work' to join the string at some point and split it again.
- If this function is inside the other one's e.g. `detectDocumentSize`, `detectViewport` and `detectScreenResolution` , it will not need this _wasted_ join but it will not be clearly evident that you need to clean these types of values before sending them. By adding this function directly on the place that we send these properties makes it a bit more evident that if we are to add a new one, the same filter should apply.

Wdyt, is it worth it or more confusing ? 

I am all in for changing it as well

jethron: Yeah I think it'd be nicer if `detectDocumentSize`, `detectViewport`, and `detectScreenResolution` all just return `[number, number] | null` and `toDimension()` just took that and returned `string | null`. Best of both worlds.

libraries/browser-tracker-core/src/helpers/browser_props.ts
```diff
@@ -1,8 +1,11 @@
...
+const DIMENSION_SEPARATOR = 'x';
+
 export function getBrowserProperties() {
   return {
-    viewport: detectViewport(),
-    documentSize: detectDocumentSize(),
-    resolution: screen.width + 'x' + screen.height,
+    viewport: floorDimensionFields(detectViewport()),
+    documentSize: floorDimensionFields(detectDocumentSize()),
+    resolution: floorDimensionFields(detectScreenResolution()),
```
jethron, line 8: Have we confirmed when the fractional values actually occur (which useragent/circumstances)? It would be good to have that concrete before we throw away data... and before someone inevitably asks why it's wrong for that random user agent.

libraries/browser-tracker-core/src/helpers/browser_props.ts
```diff
@@ -55,5 +58,18 @@ function detectDocumentSize() {
...
+  return isNaN(w) || isNaN(h) ? '' : w + DIMENSION_SEPARATOR + h;
+}
+
+function detectScreenResolution() {
+  return screen.width + DIMENSION_SEPARATOR + screen.height;
+}
+
+export function floorDimensionFields(field?: string | null) {
+  return !field
+    ? field
```
jethron, line 70: Nit: Personally find `field && field.split()...` or flipping the conditionals a bit nicer here, starting with not for the ternary I find is a bit confusing.

jethron (approved):
No real problems, LGTM!

libraries/browser-tracker-core/src/helpers/browser_props.ts
igneel64: > throw away data
In what case will this occur ? Or you mean, not allowing the fractional part to get through ?

The change in behaviour from this PR is only that fractional values will be rounded down to the closest int for these attributes.

libraries/browser-tracker-core/src/helpers/browser_props.ts
igneel64: Yeap, missed that. Thanks and fixed!

libraries/browser-tracker-core/src/helpers/browser_props.ts
jethron: Yes, just that we're discarding that fractional info. Its existence means _something_ (bot?), I just don't know what, and we're throwing that data point away by rounding it. :stuck_out_tongue:

igneel64: Thought about it a bit, let me provide some of my thoughts:
- If the `innerWidth` or any other value is fractional probably means there is some user-agent bug or unexpected behaviour. It does not automatically signal bot traffic though.
- Does a condition like a fractional value on width/height make it a bad **event** that needs to be manually reviewed _without having a clear response if it is indeed bad_ ? That's where I am stuck in the logic.
- We cannot change the type of the parameter to accept fractional values as it is part of the atomic table and will probably be a breaking change.
- What we can is introduce a `browser_context` model change _(not sure we even have it modeled yet)_ where we allow for fractional values to be pushed in and analyzed normally without ending in the bad queue.

WDYT ?

jethron: To be clear, I have no problem with this change and agree with it - we just need to know when/why the fractional value occurs if possible, so it's explainable why the captured value is different from the real value.

Like if atomic.events had a maxLength on useragent of like 120 and we made the trackers .slice(0,120) the value so it was always good, we would probably want to document some of the known cases of it being over 120 characters to at least understand why the values are partial sometimes, right?

igneel64: Yeap, I will add a task for that so it will remain on our radar! I totally agree

Issue #1200, opened 2023-05-31
Add capability to track transaction errors with the snowplow-ecommerce plugin


igneel64:
**Is your feature request related to a problem? Please describe.**
Whenever an ecommerce transaction needs to be fullfilled across systems _,payment processors, internal systems_, there are many issues that might arise and also different resolution decisions. 

**Describe the solution you'd like**
Add the capability to track transaction errors and resolutions to the snowplow-ecommerce plugin.

**Additional context**
Corresponds to the [transaction_error](https://github.com/snowplow/iglu-central/pull/1297) schema.

PR #1210, opened 2023-06-19
Release/3.13.0

igneel64:
Release 3.13.0 containing:
- #1200 
- #1189


matus-tomlein (commented):
LGTM! But I think that you need to rebase this on top of master first, because although the `nextBump` here is `minor`, the `nextBump` in master is `patch` which I guess would result in this being a 3.12.2 after merging to master if we don't add a "Prepare for release" commit with the change in the `nextBump`.

matus-tomlein (approved):

PR #1212, opened 2023-06-22
Add Vimeo tracking plugin (close #1156)

### LINKED ISSUE #1156: Add Vimeo tracking plugin

greg-el:


plugins/browser-plugin-vimeo-tracking/src/mediaPluginBinding.ts
```diff
@@ -0,0 +1,215 @@
...
+  trackedPlayer: TrackedPlayer,
+  eventType: EventType,
+  eventData: EventData | unknown = {},
+  trackError: (error: VimeoError) => void
+): void {
+  getVimeoMediaAttributes(trackedPlayer)
+    .then((player) => {
+      updateMediaTracking({
+        id: trackedPlayer.config.id,
+        player,
```
matus-tomlein, line 46: Nit: this is fine to separate the `update` and `track` calls but wouldn't it be better to pass the `player` property directly in the `track` calls and avoid this extra update one?

plugins/browser-plugin-vimeo-tracking/src/mediaPluginBinding.ts
```diff
@@ -0,0 +1,215 @@
...
+      break;
+
+    case InternalVimeoEvent.INTERACTIVEHOTSPOTCLICKED:
+      trackMediaSelfDescribingEvent({
+        id,
+        event: {
+          schema: VimeoSchema.HOTSPOT_CLICK,
+          data: {},
+        },
+        context: [{ schema: VimeoSchema.INTERACTION, data: eventData as VimeoInteractiveHotspotClickedData }],
```
matus-tomlein, line 162: I am a bit scared by passing the values directly to the event or entity without filtering for only the known fields – in case the Vimeo APIs change and add a new property, it will result in validation errors.

plugins/browser-plugin-vimeo-tracking/src/mediaPluginBinding.ts
```diff
@@ -0,0 +1,215 @@
...
+          schema: VimeoSchema.CHAPTER_CHANGE,
+          data: eventData as ChapterData,
+        },
+      });
+      break;
+
+    case InternalVimeoEvent.LOADED:
+      // 'Loaded' can be fired by a new video being loaded into the same player
+      // so we should ensure player attributes are up-to-date
+      getVimeoMediaAttributes(trackedPlayer)
```
matus-tomlein, line 190: Nit: Is this necessary given that we already did the same thing in the `trackEvent` call?

plugins/browser-plugin-vimeo-tracking/src/types.ts
```diff
@@ -0,0 +1,115 @@
+import Player, { Error as VimeoError, QualityChangeEvent, TrackKind, VimeoCuePoint } from '@vimeo/player';
+import { InternalVimeoEvent, VimeoEvent } from './events';
+import { MediaEventType, MediaTrackingConfiguration } from '@snowplow/browser-plugin-media/src/types';
+
+/** The configuration object for the Vimeo tracking plugin */
+export interface VimeoTrackingConfiguration extends Omit<MediaTrackingConfiguration, 'captureEvents'> {
```
matus-tomlein, line 6: Omit is a cool type!

plugins/browser-plugin-vimeo-tracking/src/types.ts
```diff
@@ -0,0 +1,115 @@
...
+
+/** A {@link Player} and its associated {@link VimeoTrackingConfiguration | configuration} */
+export type TrackedPlayer = {
+  player: Player;
+  eventListeners: Partial<Record<InternalVimeoEvent, (eventData: EventData | unknown) => void>>;
+  config: InternalVimeoTrackingConfiguration;
+};
+
+/** Associated type for the Vimeo "meta" schema */
+export type VimeoPlayerMetadata = {
```
matus-tomlein, line 25: I didn't find where this type was used, is there an entity that is using it?

matus-tomlein (commented):
This looks great! Is there an example coming to the React app in `snowplow-javascript-tracker-examples`? Would be great to demo this and inspect the events.

plugins/browser-plugin-vimeo-tracking/src/types.ts
greg-el: I think I was going to use this as `SelfDescribingJson<VimeoPlayerMetadata>`, but then found out that wasn't possible. I've used it [here](https://github.com/snowplow/snowplow-javascript-tracker/pull/1212/commits/0d534e5ac54e50b7a30ce44f44125eda0b93ec7b#diff-a5acdcad1212365b743e3979aee311da7d3073cee339cd6d28a3043b1cb0003dR81) now.

plugins/browser-plugin-vimeo-tracking/src/mediaPluginBinding.ts
greg-el: Added explicit conversion functions rather than `as` casts.

plugins/browser-plugin-vimeo-tracking/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export { VimeoEvent } from './events';
+export { VimeoTrackingConfiguration } from './types';
+export { startVimeoTracking, endVimeoTracking, VimeoTrackingPlugin } from './initialization';
```
matus-tomlein, line 3: Should we reexport the `trackXX` APIs from the media plugin that are not auto-tracked by the Vimeo plugin? In particular, I think these would make sense:

* Ad events: trackMediaAdBreakEnd, trackMediaAdBreakStart, trackMediaAdClick, trackMediaAdComplete, trackMediaAdFirstQuartile, trackMediaAdMidpoint, trackMediaAdPause, trackMediaAdResume, trackMediaAdSkip, trackMediaAdStart, trackMediaAdThirdQuartile
* Custom events: trackMediaSelfDescribingEvent

greg-el: That seems sensible, prevents any confusion from importing from 2 separate sources.

matus-tomlein (approved):
LGTM, great work! 👍 

I think we should wait for the implementation of aggregating seek start and end events in the media plugin to prevent the flood of events when the user scrubs the video.

PR #1214, opened 2023-06-27
Allow events to be tracked on the onSessionUpdate callback (close #1213)

### LINKED ISSUE #1213: Allow for tracking events on the `onSessionUpdate` callback
Allow for the `onSessionUpdate` callback to track events.

**Additional context**
Currently the `onSessionUpdate` tracker callback cannot be used to track other events as in cases where the session is reset, it would not be able to detect a session identifier and create a loop.

igneel64:
Move the `onSessionUpdate` callback logic from the `startNewIdCookieSession` which just creates a new session id to the `startNewIdCookieSession` function to make sure persistence is there before trying to call any other events after a session update.

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/index.ts
+++ b/libraries/browser-tracker-core/src/tracker/index.ts
@@ -91,6 +91,7 @@ import {
   clientSessionFromIdCookie,
   incrementEventIndexInIdCookie,
   emptyIdCookie,
+  eventIndexFromIdCookie,
 } from './id_cookie';
 import { CLIENT_SESSION_SCHEMA, WEB_PAGE_SCHEMA, BROWSER_CONTEXT_SCHEMA } from './schemata';
 import { getBrowserProperties } from '../helpers/browser_props';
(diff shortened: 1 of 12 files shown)


matus-tomlein (commented):
I have just tried to use this in our React demo app and for some reason when I track an event without the `setTimeout`, the event is not tracked at all. If I add the `setTimeout`, the event gets tracked normally. I don't really understand why, do you have any idea? This is the code I use to create the tracker:

```js
let tracker = newTracker('ns1', 'http://localhost:9090', {
  onSessionUpdateCallback: function (session) {
    console.log('Session updated: ', session);
    trackStructEvent({
      category: "session",
      action: "update",
      label: "Session updated",
    });
  },
  plugins: [FormTrackingPlugin(), YouTubeTrackingPlugin()]
});
```

I get the `console.log` output but not the structured event.

igneel64:
@matus-tomlein All the cases seem to work as expected for me now.
Got really bogged down in undocumented mutations.

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -564,42 +571,54 @@ export function Tracker(
...
 
-    /*
-     * Sets or renews the session cookie
+    /**
+     * Sets or renews the session cookie.
+     * Responsible for calling the `onSessionUpdateCallback` callback.
      */
-    function setSessionCookie() {
+    function setSessionCookie(options: SessionCookieOptions = {}) {
+      const { onSessionUpdateCallback, configAnonymousTracking, clientSession } = options;
```
matus-tomlein, line 579: Nit 2: as we called, it is a bit confusing for me to have the `onSessionUpdateCallback` being passed here because this function is called even if the session does not change. The callback only needs to be called if `startNewIdCookieSession` is called. My preference would be to handle this outside this function. In effect it is the same thing, just it would be clearer to me personally.

matus-tomlein (approved):
I just had 2 nitpicks about how the arguments for the `setSessionCookie` are passed, but otherwise this LGTM, great job dealing with the confusing code!

PR #1216, opened 2023-06-29
Release/3.13.1

igneel64:
- Properly allow tracking events on the `onSessionUpdate` callback


matus-tomlein (approved):
LGTM!

(nit: the order of the commits is wrong, but I can live with that)

PR #1217, opened 2023-06-29
Add tracker.getQueueName method

### LINKED ISSUE #1215: Expose localstorage queue name
Currently there is no way to reliably get the localstorage queue name being used for a tracking session.

A tracker method which allows retrieving the queue name.
This can either be the current queue used or a function with an argument of the method the client wants to retrieve the queue name for. _GET and POST queues are different._

igneel64:
Add the `tracker.getQueueName` method.

This method will return the currently used queue name or if provided arguments will return the one that would be used with the provided `eventMethod`.

close #1215

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/out_queue.ts
+++ b/libraries/browser-tracker-core/src/tracker/out_queue.ts
@@ -33,6 +33,7 @@ import { SharedState } from '../state';
 import { localStorageAccessible } from '../detectors';
 import { LOG, Payload } from '@snowplow/tracker-core';
 import { PAYLOAD_DATA_SCHEMA } from './schemata';
+import { EventMethod } from './types';
 
 export interface OutQueue {
   enqueueRequest: (request: Payload, url: string) => void;
@@ -40,7 +41,11 @@ export interface OutQueue {
   setUseLocalStorage: (localStorage: boolean) => void;
   setAnonymousTracking: (anonymous: boolean) => void;
   setCollectorUrl: (url: string) => void;
-  setBufferSize: (bufferSize: number) => void;
+  setBufferSize: (newBufferSize: number) => void;
+  /**
+   * Returns the currently used queue name or if the `eventMethod` argument is provided, the queue name for the eventMethod.
+   */
+  getName: (eventMethod?: Exclude<EventMethod, 'beacon'>) => string;
 }
 
 /**
@@ -48,7 +53,7 @@ export interface OutQueue {
  * Instantiated once per tracker instance.
  *
  * @param id - The Snowplow function name (used to generate the localStorage key)
- * @param sharedSate - Stores reference to the outbound queue so it can unload the page when all queues are empty
+ * @param sharedState - Stores reference to the outbound queue so it can unload the page when all queues are empty
  * @param useLocalStorage - Whether to use localStorage at all
  * @param eventMethod - if null will use 'beacon' otherwise can be set to 'post', 'get', or 'beacon' to force.
  * @param postPath - The path where events are to be posted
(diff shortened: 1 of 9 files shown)


libraries/browser-tracker-core/src/tracker/types.ts
```diff
@@ -580,6 +580,11 @@ export interface BrowserTracker {
    */
   clearUserData: (configuration?: ClearUserDataConfiguration) => void;
 
+  /**
+   * Returns the currently used queue name or if the `eventMethod` argument is provided, the queue name for the eventMethod.
+   */
+  getQueueName: (eventMethod?: 'post' | 'get' | undefined) => string;
```
greg-el, line 586: nit: This can be `eventMethod?: 'post' | 'get'` if you are using `?`, right?

greg-el (changes requested):
Looks good! Just a couple of thoughts I've had on the types.

I think I'm not properly following the types you've used here. I can see both:
- `'post' | 'get' | undefined` 
- `Exclude<EventMethod, 'beacon'>` 

This seems to boil down to the same type, a potentially optional `'post' | 'get'`. Could it be clearer to create a type alias for `Exclude<EventMethod, 'beacon'>`, which then could be used with `?` as required?

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -67,7 +72,7 @@ export interface OutQueue {
  */
 export function OutQueueManager(
   id: string,
-  sharedSate: SharedState,
+  sharedState: SharedState,
```
jethron, line 75: :1st_place_medal: This bothered me every time I looked at this file.

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -137,16 +144,20 @@ export function OutQueueManager(
...
-    sharedSate.bufferFlushers.push(function (sync) {
+    sharedState.bufferFlushers.push(function (sync) {
       if (!executingQueue) {
         executeQueue(sync);
       }
     });
   }
 
+  function getQueueName(id: string, method: Exclude<EventMethod, 'beacon'>) {
+    return `snowplowOutQueue_${id}_${method === 'post' ? 'post2' : 'get'}`;
```
jethron, line 158: Since the type isn't checked at runtime and POST is the default these days, should we flip this and check for GET instead, defaulting to post?

So even though `getQueueName('beacon')` isn't actually correct use of the API, it'll still produce the correct/desired result more often? Probably a philosophical thing though, [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) vs [GIGO](https://en.wikipedia.org/wiki/Garbage_in,_garbage_out).

trackers/browser-tracker/test/tracker.test.ts
```diff
@@ -276,3 +277,24 @@ describe('Activity tracker behaviour', () => {
     expect(secondPageId).toBe(extractPageId(ppl));
   });
 });
+
+describe('API', () => {
+  it('getQueueName does return the queue name', () => {
```
jethron, line 282: Probably just nest this under another `describe('getQueueName', () => {})`?

Nit: s/does return/returns/

jethron (approved):
LGTM.

What's the intended use case here vs just a "clearQueue" method or similar? Does this encourage people to mess with the queue themselves? Or is this for like a plugin API?

libraries/browser-tracker-core/src/tracker/out_queue.ts
igneel64: Sure, good idea :)

libraries/browser-tracker-core/src/tracker/types.ts
igneel64: This slipped in the PR, should be `Exclude<EventMethod, 'beacon'>` as in the implementation.

greg-el (approved):

PR #1224, opened 2023-08-08
Allow disabling activity tracking actions (close #1161)

### LINKED ISSUE #1161: Add disableActivityTracking function

In single page apps certain portions of the app we may not want activity tracking.

The tracker should expose a `disableActivityTracking` method that stops activity tracking.

**Additional context**
For more info see the Discourse thread here: https://discourse.snowplow.io/t/dynamically-turn-off-activity-tracking/8200/5

igneel64:
Add `disableActivityTracking` and `disableActivityTrackingCallback` to disable the respective activity tracking and activity tracking callbacks from the tracker.

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/index.ts
+++ b/libraries/browser-tracker-core/src/tracker/index.ts
@@ -1048,20 +1048,20 @@ export function Tracker(
             //Clear page ping heartbeat on new page view
             window.clearInterval(config.activityInterval);
 
-            activityInterval(config, context, contextCallback);
+            scheduleActivityInterval(config, context, contextCallback);
           }
         }
       }
     }
 
-    function activityInterval(
+    function scheduleActivityInterval(
       config: ActivityConfig,
       context?: Array<SelfDescribingJson> | null,
       contextCallback?: (() => Array<SelfDescribingJson>) | null
     ) {
-      const executePagePing = (cb: ActivityCallback, c: Array<SelfDescribingJson>) => {
+      const executePagePing = (cb: ActivityCallback, context: Array<SelfDescribingJson>) => {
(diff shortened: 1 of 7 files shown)


libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -1137,6 +1137,17 @@ export function Tracker(
...
 
+    function disableActivityTrackingAction(actionKey: keyof ActivityConfigurations) {
+      const callbackConfiguration = activityTrackingConfig.configurations[actionKey];
+      if (callbackConfiguration?.configMinimumVisitLength === 0) {
+        window.clearTimeout(callbackConfiguration?.activityInterval);
+      } else {
+        window.clearInterval(callbackConfiguration?.activityInterval);
+      }
+
+      activityTrackingConfig.configurations[actionKey] = undefined;
```
matus-tomlein, line 1148: I'm wondering, should we also set `activityTrackingConfig.enabled` to false in case there are no more configurations left? From what I can tell it'd be more like a cosmetic change and not have much effect on anything (in which case do we really need the flag?), but wondering if you can see some reason to or considered it.

matus-tomlein (approved):
Looks great, this has been requested a lot!

jethron (approved):
LGTM

libraries/browser-tracker-core/src/tracker/index.ts
igneel64: The `enabled` flag seems to be mostly used for the activity tracking at the start and between pageviews. It does not have control over the interval/timeouts of the actual pinging.

It was the first thing I tried 😄

PR #1226, opened 2023-08-08
Update browser-plugin-ga-cookies to include options for Google Analytics 4 cookies

igneel64:
This update allows for the `browser-plugin-ga-cookies` to collect Google Analytics 4 cookie values.

Up to now the plugin did not accept any options and only collected a standard set of values for the schema `iglu:com.google.analytics/cookies/jsonschema/1-0-0`.

This PR adds the capability for the plugin to collect Google Analytics 4 cookies without breaking the previous implementation. With the options, you can specifiy which set of cookies you want to collect and since the [GA4 session cookie](https://support.google.com/analytics/answer/11397207?hl=en) name is based on a parameter called `container-id`, the client should input this as well to collect the correct value.

E.g.
```js
GaCookiesPlugin(); /* Current behaviour collecting only UA cookie values. */

GaCookiesPlugin({ ga4: true, measurementId: 'G-test' }); /* Collects UA and GA4 cookie values. */

GaCookiesPlugin({ ua: false, ga4: true, measurementId: 'G-test' }); /* Collects GA4 cookie values. */
```

_After collecting the GA4 session cookie, the customer will be able to extract information such as the session id and session index from the cookie._

New schema for GA4 cookies:
- https://github.com/snowplow/iglu-central/pull/1325

In the case of cookiePrefix/es N, N number of GA4 cookie contexts will be attached.


igneel64:
For JS tracker V4 we can change the defaults to GA4 cookies and have UA only as opt-in because it will be deprecated for all customers.

plugins/browser-plugin-ga-cookies/README.md
```diff
@@ -36,14 +36,25 @@ Initialize your tracker with the GaCookiesPlugin:
...
+) ] });
+
+/* 
+ * Available plugin options `GACookiesPluginOptions`:
+ * {
+ * ua: Send Universal Analytics specific cookie values. Defaults to true.
+ * ga4: Send Google Analytics 4 specific cookie values. Defaults to true
+ * ga4ContainerId: Container id to search the Google Analytics 4 session cookie. The cookie has the form of _ga_<container-id> where <container-id> is the data stream container id. Defaults to the empty string.
+ * } 
+ */
```
jethron, line 50: Can we update `trackers/javascript-tracker/src/configuration.ts` and `trackers/javascript-tracker/src/features.ts` to allow passing this through in the `newTracker` call too?

plugins/browser-plugin-ga-cookies/README.md
```diff
@@ -5,7 +5,7 @@
 
 Browser Plugin to be used with `@snowplow/browser-tracker`.
 
-Adds GA Cookies to your Snowplow tracking.
+Adds Universal Analytics and Google Analytics 4 cookies to your Snowplow tracking.
```
jethron, line 8: I guess technically Classic GA too for the old schema. :joy:

plugins/browser-plugin-ga-cookies/src/index.ts
```diff
@@ -28,27 +28,67 @@
...
+interface GACookiesPluginOptions {
+  ua?: boolean;
+  ga4: boolean;
+  ga4ContainerId: string;
+}
+
+const defaultPluginOptions = {
+  ua: true,
+  ga4: false,
+  ga4ContainerId: '',
```
jethron, line 45: Someone one day will want the session IDs for multiple trackers, should this be an array of stream IDs instead?

plugins/browser-plugin-ga-cookies/src/index.ts
```diff
@@ -28,27 +28,67 @@
...
+          if (value) {
+            uaCookiesContext.data[cookieType] = value;
+          }
+        });
+        contexts.push(uaCookiesContext);
+      }
+
+      if (ga4) {
+        if (!ga4ContainerId) {
+          LOG.warn('"ga4ContainerId" option not provided in browser-plugin-ga-cookies.');
```
jethron, line 76: Should we just grab everything in `_ga_.+` in this case instead?

plugins/browser-plugin-ga-cookies/src/types.ts
```diff
@@ -0,0 +1,14 @@
+export interface UACookies {
+  __utma?: string;
+  __utmb?: string;
+  __utmc?: string;
+  __utmv?: string;
+  __utmz?: string;
+  _ga?: string;
+  [key: string]: string | undefined;
```
jethron, line 8: Any idea why this `key` thing is defined?

igneel64: I think this was about making typescript happy in the loop of the index file _(not in the most TypeScripty way)_.

plugins/browser-plugin-ga-cookies/src/index.ts
igneel64: Hmm to allow this case, do you believe it would be fine to turn the `ga4ContainerId` value in the schema into an array of strings ?
 _(afaik gtag.js will use the same `_ga` cookie independently of the number of streams)_

Also I am thinking of adding an option to allow for collecting ga cookies using a different prefix https://developers.google.com/tag-platform/devguides/cookies#rename_cookies .

jethron: This is most likely going to be used as comparison to GA4 so you'll need to know which measurement ID (my bad, stream ID is a different concept) the session ID is supposed to match with.

So in the options an array is probably fine, but then sadly in the schema I think you'll need array of objects to capture them properly as key value pairs...

Might be able to include the measurement ID as a field and attach a separate entity for each one found, as an alternative? But you'll then have duplicate `_ga` info... :disappointed:

plugins/browser-plugin-ga-cookies/src/index.ts
igneel64: I did not add this feature for now, if you think it might be a useful option, feel free to reopen.

plugins/browser-plugin-ga-cookies/README.md
```diff
@@ -36,14 +36,26 @@ Initialize your tracker with the GaCookiesPlugin:
...
-newTracker('sp1', '{{collector}}', { plugins: [ GaCookiesPlugin() ] }); // Also stores reference at module level
+newTracker('sp1', '{{collector}}', { plugins: [ GaCookiesPlugin(
+    /* pluginOptions */
+) ] });
+
+/* 
+ * Available plugin options `GACookiesPluginOptions`:
+ * {
+ * ua: Send Universal Analytics specific cookie values. Defaults to true.
+ * ga4: Send Google Analytics 4 specific cookie values. Defaults to true
```
jethron, line 47: Defaults to `false` atm.

plugins/browser-plugin-ga-cookies/README.md
```diff
@@ -36,14 +36,26 @@ Initialize your tracker with the GaCookiesPlugin:
...
+ * ga4MeasurementId: Measurement id/ids to search the Google Analytics 4 session cookie. Can be a single measurement id as a string or an array of measurement id strings. The cookie has the form of _ga_<container-id> where <container-id> is the data stream container id.
```
jethron, line 48: Nit: Maybe `<prefix>_ga_<container-id>`?

plugins/browser-plugin-ga-cookies/src/index.ts
```diff
@@ -1,54 +1,85 @@
...
-      return [gaCookieData];
+        const ga4CookiesContext: SelfDescribingJson<GA4Cookies> = {
+          schema: GOOGLE_ANALYTICS_4_COOKIES_SCHEMA,
+          data: {
+            _ga: '',
+            session_cookies: [],
+          },
+        };
+
+        const measurementIdentifiers = Array.isArray(ga4MeasurementId) ? [...ga4MeasurementId] : [ga4MeasurementId];
```
jethron, line 63: Nit: Don't need to unpack/repack `ga4MeaurementId` here.

plugins/browser-plugin-ga-cookies/src/index.ts
```diff
@@ -1,54 +1,85 @@
...
+
+        [...measurementIdentifiers, GA_USER_COOKIE].forEach(function (cookieIdentifier) {
+          if (cookieIdentifier === GA_USER_COOKIE) {
+            ga4CookiesContext.data._ga = cookie(cookiePrefix + cookieIdentifier);
```
jethron, line 67: Nit: Just do this unconditionally outside of the loop.

plugins/browser-plugin-ga-cookies/src/index.ts
```diff
@@ -1,54 +1,85 @@
...
+          } else {
+            /* Measurement id/s */
+            const sessionCookieValue = cookie(
+              cookiePrefix + cookieIdentifier.replace(GA4_MEASUREMENT_ID_PREFIX, GA4_COOKIE_PREFIX)
+            );
+            ga4CookiesContext.data.session_cookies?.push({
```
jethron, line 73: Should we only `push()` it if we got a value? WDYT?

plugins/browser-plugin-ga-cookies/src/index.ts
```diff
@@ -1,54 +1,85 @@
...
+            uaCookiesContext.data[cookieType] = value;
+          }
+        });
+        contexts.push(uaCookiesContext);
+      }
+
+      if (ga4) {
+        if (!ga4MeasurementId) {
+          LOG.warn('"ga4MeasurementId" option not provided in browser-plugin-ga-cookies.');
+          return contexts;
```
jethron, line 53: So the API is now that if you want the old behaviour (client ID cookie but don't care about session IDs), you need to specify `ga4MeasurementId: []` to pass this check. Should we be more explicit that that's an option?

Also the default now for `{ua: false, ga4: true}` will do nothing, is that right?

plugins/browser-plugin-ga-cookies/src/index.ts
igneel64: Kind of reflexive to avoid mutations on the `ga4MeasurementId` array.

plugins/browser-plugin-ga-cookies/README.md
igneel64: Tried to keep it the similar to the official documentation but will add a not for the prefix, thanks!

plugins/browser-plugin-ga-cookies/src/index.ts
igneel64: Fixed, good call.

plugins/browser-plugin-ga-cookies/src/index.ts
igneel64: Updated to send the user cookie (_ga) when `{ ga4: true }` and not log the warning.
Should make sure the documentation page specifies that if you do not pass measurement ids, you would not be able to get session cookies.

jethron (approved):
LGTM

PR #1228, opened 2023-08-16
Add E2E testIdentifier randomness between test retries

igneel64:
Small fix to allow e2e test identifiers to add some randomness between test retries in the identifiers.

Currently tests that go into further retries will fail as the `testIdentifier` _(translating this to appId in the events)_ will be the same.


greg-el (approved):
LGTM!

matus-tomlein (approved):

PR #1233, opened 2023-08-28
Extend cross-domain linking with more user/session information

igneel64:
Extending the cross-domain linking capability with a new parameter format adding the information of [cross-navigation](https://github.com/snowplow/iglu-central/pull/1327/files) to a URLEncoded(B64(contents)) value which will be placed in the cross-domain linker param (`_sp`).

The tracker now accepts a `useExtendedCrossDomainLinker` parameter of type `ExtendedCrossDomainLinkerOptions`:
```js
{
/* Enables the new `_sp` format using the cross_navigation information */
 useExtendedCrossDomainLinker: true
}

{
/* Enables the new `_sp` format using the cross_navigation information plus optional configuration for `ExtendedCrossDomainLinkerAttributes` . */
 useExtendedCrossDomainLinker: {
  userId?: boolean;
  sessionId?: boolean;
  sourceId?: boolean;
  sourcePlatform?: boolean;
  /**
   * Allow for the collection of the link text when a cross-domain link is clicked. Can also accept a callback for customizing information collection.
   */
  reason?: boolean | ((evt: Event) => string);
}
}
```

The default behaviour is not changed.

Notes:
- Added `testing-library/dom` allowing us to test DOM interactions on unit type of tests.

### DIFF
--- a/libraries/browser-tracker-core/src/helpers/cross_domain.ts
+++ b/libraries/browser-tracker-core/src/helpers/cross_domain.ts
@@ -1,0 +1,72 @@
+import { ExtendedCrossDomainLinkerAttributes, Platform } from '../tracker/types';
+
+const DEFAULT_CROSS_DOMAIN_LINKER_PARAMS: ExtendedCrossDomainLinkerAttributes = {
+  sessionId: true,
+  sourceId: true,
+  sourcePlatform: false,
+  userId: false,
+  reason: false,
+};
+
+interface ExtendedCrossDomainLinkerValues {
+  domainUserId?: string;
+  /* Timestamp of the cross-domain link click. */
+  timestamp?: number;
+  /* Current user ID as set with setUserId(). */
+  userId?: string;
+  /* Visitor ID. */
+  sessionId?: string;
+  /* The app ID. */
+  sourceId?: string;
+  sourcePlatform?: Platform;
+  /* Link text of the cross-domain link. */
+  reason?: string;
+}
+
+export function createCrossDomainParameterValue(
+  isExtendedFormat: boolean,
+  attributeConfiguration: ExtendedCrossDomainLinkerAttributes | undefined,
+  attributeValues: ExtendedCrossDomainLinkerValues & {
+    /* As `reason` might be a callback, we also need to pass the event to calculate the reason value. */
+    event: Event;
+  }
+): string {
+  let crossDomainParameterValue;
+  const timestamp = new Date().getTime();
+  const config = { ...DEFAULT_CROSS_DOMAIN_LINKER_PARAMS, ...attributeConfiguration };
+  const { domainUserId, userId, sessionId, sourceId, sourcePlatform, event } = attributeValues;
+
+  const eventTarget = event.currentTarget as HTMLAnchorElement | HTMLAreaElement | null;
+  const reason = typeof config.reason === 'function' ? config.reason(event) : eventTarget?.textContent?.trim();
+
+  if (isExtendedFormat) {
+    /* Index is used by Enrich, so it should not be changed. */
+    crossDomainParameterValue = [
+      domainUserId,
+      timestamp,
+      config.sessionId && sessionId,
+      config.userId && urlSafeBase64Encode(userId || ''),
+      config.sourceId && urlSafeBase64Encode(sourceId || ''),
+      config.sourcePlatform && sourcePlatform,
+      config.reason && urlSafeBase64Encode(reason || ''),
+    ]
+      .map((attribute) => attribute || '')
+      .join('.')
+      // Remove trailing dots
+      .replace(/([.]*$)/, '');
+  } else {
+    crossDomainParameterValue = attributeValues.domainUserId + '.' + timestamp;
+  }
+
+  return crossDomainParameterValue;
+}
+
+/**
+ *
+ * The url-safe variation emits - and _ instead of + and / characters. Also removes = sign padding.
+ * @param {string} str The string to encode in a URL safe manner
+ * @return {string}
+ */
+export function urlSafeBase64Encode(str: string) {
+  return btoa(str).replace(/\+/g, '-').replace(/\//g, '_').replace(/\=+$/, '');
+}
--- a/libraries/browser-tracker-core/src/tracker/index.ts
+++ b/libraries/browser-tracker-core/src/tracker/index.ts
@@ -1,33 +1,3 @@
-/*
- * Copyright (c) 2022 Snowplow Analytics Ltd, 2010 Anthon Pang
- * All rights reserved.
- *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions are met:
- *
- * 1. Redistributions of source code must retain the above copyright notice, this
- *    list of conditions and the following disclaimer.
- *
- * 2. Redistributions in binary form must reproduce the above copyright notice,
- *    this list of conditions and the following disclaimer in the documentation
- *    and/or other materials provided with the distribution.
- *
- * 3. Neither the name of the copyright holder nor the names of its
- *    contributors may be used to endorse or promote products derived from
- *    this software without specific prior written permission.
- *
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
- * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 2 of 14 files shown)


libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -365,20 +350,40 @@ export function Tracker(
...
+          const extendedLinkData: ExtendedCrossDomainLinker = {
+            domainUserId,
+            userId: businessUserId || undefined,
+            sessionId: memorizedSessionId,
+            sourceId: configTrackerSiteId,
+            sourcePlatform: configPlatform,
+            timestamp,
+            reason: collectCrossDomainLinkText ? elt?.textContent?.trim() : undefined,
+          };
+          crossDomainParameterValue = encodeURIComponent(btoa(JSON.stringify(extendedLinkData)));
```
matus-tomlein, line 374: Question: I think we already discussed this but tbh I'm not 100% clear on what is the reason for using base64 encoding on the value. The entropy of the parameter will be the same, just that we won't see a UUID in there but a different set of characters – is there reason to believe that browsers will be less likely to remove that? I am just a bit worried that we are adding unnecessary complexity here.

matus-tomlein (commented):
In terms of the configuration of the feature, there are a couple of things on my mind:

1. Nit: Would "optional" be better than "extended" as a name for the extra properties? Extended feels a bit like enhanced ecommerce – what will we call the extended extended version? :D 
2. I think it would make sense to make the optional properties configurable – I can imagine some people might not want to send the `userId` or `appId` in the links but want the `sessionId`.

My suggestion would be to make the configuration something like this:

```ts
{
    crossDomainLinkerOptionalProps: {
         userId: true,
         sessionId: true,
         sourceId: true,
         sourcePlatform: true,
         reason: true
    }
}
```

We could also let users enable all optional props as you do now:

```ts
{
    crossDomainLinkerOptionalProps: true
}
```

libraries/browser-tracker-core/src/tracker/index.ts
igneel64: Excellent question, about the B64, the reason I added it was that I imagine internal apps, or in the future some more privacy-friendly browsers as well, might implement a 'strip email' filter from URL params.
Since `userId` will be a part of the new payload, and in some cases the user ID might be set as the email, I feel that this will give us some extra room.

WDYT ?

igneel64:
About the choise of `extended` in the naming I was thinking that this in v3 will add the 'new' way _(b64)_ with the  additional properties.
In v4 the old one would be removed and we would only keep this one.

If though we decide to keep the same formatting _(split with dots)_, optional will be a better naming.

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -366,20 +351,40 @@ export function Tracker(
...
+        const timestamp = new Date().getTime();
+        if (extended) {
+          const extendedLinkData: ExtendedCrossDomainLinker = {
+            domainUserId,
+            userId: businessUserId || undefined,
+            sessionId: memorizedSessionId,
+            sourceId: configTrackerSiteId,
+            sourcePlatform: configPlatform,
+            timestamp,
+            reason: collectCrossDomainLinkText ? elt?.textContent?.trim() : undefined,
```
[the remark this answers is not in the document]
igneel64: Yeap nice idea, adding!

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -366,20 +351,40 @@ export function Tracker(
...
+      const CROSS_DOMAIN_PARAMETER_NAME = '_sp';
+
+      return (evt) => {
+        const elt = evt.currentTarget as HTMLAnchorElement | HTMLAreaElement | null;
+        let crossDomainParameterValue;
+        const timestamp = new Date().getTime();
+        if (extended) {
+          const extendedLinkData: ExtendedCrossDomainLinker = {
+            domainUserId,
+            userId: businessUserId || undefined,
```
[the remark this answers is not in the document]
igneel64: Fixing that with optionals!

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -366,20 +351,40 @@ export function Tracker(
...
+            sessionId: memorizedSessionId,
+            sourceId: configTrackerSiteId,
+            sourcePlatform: configPlatform,
+            timestamp,
+            reason: collectCrossDomainLinkText ? elt?.textContent?.trim() : undefined,
+          };
+          crossDomainParameterValue = encodeURIComponent(btoa(JSON.stringify(extendedLinkData)));
```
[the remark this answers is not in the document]
igneel64: Taking a look at how enrich [does it](https://github.com/snowplow/enrich/blob/a3c7fd69961223c08f177c883b6d2f61892f33b3/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala#L53) and going to the docs for the package, I think we are fine doing the same replacement as the encoder [does](https://commons.apache.org/proper/commons-codec/archives/1.7/apidocs/org/apache/commons/codec/binary/Base64.html#encodeBase64URLSafe(byte[])).

Something like:
   `replace(/\+/g, '-').replace(/\//g, '_')`

Will be adding that.

libraries/browser-tracker-core/src/helpers/cross_domain.ts
```diff
@@ -0,0 +1,55 @@
...
+      domainUserId,
+      timestamp,
+      config.sessionId && sessionId,
+      config.userId && urlSafeBase64Encode(userId || ''),
+      config.sourceId && sourceId,
+      config.sourcePlatform && sourcePlatform,
+      config.reason && urlSafeBase64Encode(reason || ''),
+    ]
+      .map((attribute) => attribute || '')
+      .join('.');
```
matus-tomlein, line 39: Nit: this is just a visual thing but it would be nicer if the trailing dots were automatically removed at the end of the string. For instance, in case only `sessionId` is enabled in the extended config options, instead of:

```
duid.ts.sessionId....
```

the string could like like:

```
duid.ts.sessionId
```

matus-tomlein (approved):
Just one nitpick about the dots at the end of the string but otherwise LGTM

libraries/browser-tracker-core/src/helpers/cross_domain.ts
```diff
@@ -0,0 +1,55 @@
...
+  const reason = typeof config.reason === 'function' ? config.reason(event) : eventTarget?.textContent?.trim();
+
+  if (isExtendedFormat) {
+    /* Index is used by Enrich, so it should not be changed. */
+    crossDomainParameterValue = [
+      domainUserId,
+      timestamp,
+      config.sessionId && sessionId,
+      config.userId && urlSafeBase64Encode(userId || ''),
+      config.sourceId && sourceId,
```
jethron, line 34: I don't think there are any restrictions on the format of appId; using a domain (that will contain `.`) is not unheard of. You may need to b64 this, too.

igneel64: Will do, thanks

libraries/browser-tracker-core/src/tracker/types.ts
```diff
@@ -648,3 +667,18 @@ export interface ClientSession extends Record<string, unknown> {
    */
   firstEventTimestamp: string | null;
 }
+
+export interface ExtendedCrossDomainLinker {
+  domainUserId?: string;
+  /* Timestamp of the cross-domain link click. */
+  timestamp?: number;
```
greg-el, line 674: Is `timestamp` required for this type? It looks like the `timestamp` value will always come from [here](https://github.com/snowplow/snowplow-javascript-tracker/pull/1233/commits/e709c9958012e5088e305024beec6ed8a401c3d7#diff-e919d7558f1e654870709d4fbf8ccef9476b13af717811c0c34b5a269052af15R20)

libraries/browser-tracker-core/src/helpers/cross_domain.ts
igneel64: Done :)

matus-tomlein: Nice!

libraries/browser-tracker-core/src/tracker/types.ts
igneel64: This type is only used internally for the creation of the domain values. I moved it around a bit to make it more clear.

PR #1235, opened 2023-08-31
Allow for web vitals and performance navigation timing plugin in javascript tracker configuration

igneel64:
Allow for adding the Web Vitals and Performance Navigation Timing plugin on the JavaScript tracker build process.


matus-tomlein:
This is now part of #1345

PR #1242, opened 2023-09-29
Fix transform function for submit_form event (close #1241)

### LINKED ISSUE #1241: Fix transform function for submit_form event
**Describe the bug**

The transform function in [custom form tracking](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/javascript-trackers/javascript-tracker/javascript-tracker-v3/tracking-events/#custom-form-tracking) seems to have no effect to `submit_form` event, while it does work as expected for `focus_form` and `change_form` events.

**To Reproduce**

```javascript
var opts = {
    fields: {
        transform: function (value, elt) {
            if (elt.id === 'name') {
                return 'Redacted';
            } 
            return value;
        }
    }
};

snowplow('enableFormTracking', { options: opts });
```

**Expected behavior**

The expectation of the above snippet is that for field whose `id` is `'name'`, the value is equal to `'Redacted'` for all form tracking events (`focus_form`,`change_form`,`submit_form`).

adatzer:


plugins/browser-plugin-form-tracking/src/helpers.ts
```diff
@@ -252,17 +228,20 @@ function getInnerFormElements(trackingMarker: string, elt: HTMLFormElement) {
       if (child.type === 'submit') {
         return;
       }
-      var elementJson: ElementData = {
-        name: getElementIdentifier(child),
-        value: child.value,
-        nodeName: child.nodeName,
+      var elementJson: ElementDataWrapper = {
+        elementData: {
```
igneel64, line 232: @adatzer isn't this a breaking change since we changed the structure from 
`{ name, value, nodeName } to { elementData: { name, value, nodeName }, originalElement }` ?

adatzer: This type is not exported. Both `ElementData` and the transform function's arguments haven't changed.

plugins/browser-plugin-form-tracking/src/helpers.ts
```diff
@@ -317,15 +298,17 @@ function getFormSubmissionListener(
     var elt = e.target as HTMLFormElement;
     var innerElements = getInnerFormElements(trackingMarker, elt);
     innerElements.forEach(function (innerElement) {
-      innerElement.value = config.fieldTransform(innerElement.value, innerElement) ?? innerElement.value;
+      var eltData = innerElement.elementData;
+      eltData.value = config.fieldTransform(eltData.value, eltData, innerElement.originalElement) ?? eltData.value;
```
matus-tomlein, line 302: This has been the behaviour also before, but I am a bit confused why this part is different from the `change_form` event above? Here we coalesce `?? eltData.value` in case the `fieldTransform` returns a null value, but in in the change form we don't. Also, in the change form event, we exclude unchecked checkboxes but here we don't.

Am I missing something here? At least the coalesce behaviour should be the same – I think if the `fieldTransform` function returns `null`, we should keep it `null`. What do you think?

matus-tomlein: Oh okay, that makes sense

plugins/browser-plugin-form-tracking/src/helpers.ts
```diff
@@ -90,12 +60,18 @@ export interface ElementData extends Record<string, string | null | undefined> {
   type?: string;
 }
 
-export type transformFn = (x: string | null, elt: ElementData | TrackedHTMLElement) => string | null;
+export type transformFn = (
+  elementValue: string | null,
+  elementInfo: ElementData | TrackedHTMLElement,
+  elt: TrackedHTMLElement
```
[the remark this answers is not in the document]
igneel64: In my view it is clearer to allow for an 'expected' input _(even if in v3 it is a clunky third attribute)_ rather than trying to explain our, mistaken in my opinion, logic of how the `elt` attribute would work. 
I would much prefer adding the option of 'here is your element that you know how to work with and its properties are as expected' rather than changing an already weird logic _(the getElementIdentifier)_.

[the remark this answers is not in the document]
matus-tomlein: Okay, happy to go with that, but please make it clear in the docs what to expect when.

matus-tomlein (approved):
LGTM, great job debugging and fixing this!

Just a couple of questions:
* Could you refer to the issue in the PR title to keep the reference? so `Fix ... (close #1241)`
* Could we make it a bit more clear what this is fixing either in the issue title or in the issue description? I think this is useful for users to understand, if they were using the function before, what was wrong.
* Could you run `rush change` to fix the failing PR check?
* Also please update the docs to make the transform function API clear.

matus-tomlein:
Sorry for coming back to this later, but just realized that this is probably not a bug fix but a new feature since we are extending the API (adding a new attribute)? I am only saying this to note that it probably shouldn't go in a patch but a minor version.

PR #1257, opened 2023-10-20
Provide option not to retry on event send failure (close #1248)

### LINKED ISSUE #1248: Provide option not to retry on event send failure

greg-el:
This PR adds an option to the `newTracker` config, which allows the user to choose if they want to retry sending events in the case of a sending failure, e.g.:

```js
newTracker(..., {
    retryFailures: true,
});
```

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/out_queue.ts
+++ b/libraries/browser-tracker-core/src/tracker/out_queue.ts
@@ -64,6 +64,7 @@ export interface OutQueue {
  * @param retryStatusCodes – Failure HTTP response status codes from Collector for which sending events should be retried (they can override the `dontRetryStatusCodes`)
  * @param dontRetryStatusCodes – Failure HTTP response status codes from Collector for which sending events should not be retried
  * @param idService - Id service full URL. This URL will be added to the queue and will be called using a GET method.
+ * @param retryFailedRequests - Whether to retry failed requests - Takes precedent over `retryStatusCodes` and `dontRetryStatusCodes`
  * @returns object OutQueueManager instance
  */
 export function OutQueueManager(
@@ -83,7 +84,8 @@ export function OutQueueManager(
   withCredentials: boolean,
   retryStatusCodes: number[],
   dontRetryStatusCodes: number[],
-  idService?: string
+  idService?: string,
+  retryFailedRequests: boolean = true
 ): OutQueue {
   type PostEvent = {
     evt: Record<string, unknown>;
@@ -351,6 +353,10 @@ export function OutQueueManager(
       // Time out POST requests after connectionTimeout
       const xhrTimeout = setTimeout(function () {
         xhr.abort();
+
+        if (!retryFailedRequests) {
+          removeEventsFromQueue(numberToSend);
+        }
         executingQueue = false;
       }, connectionTimeout);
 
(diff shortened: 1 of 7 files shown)


libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -351,7 +352,14 @@ export function OutQueueManager(
...
+        if (retryFailures) {
+          LOG.warn(`Request failed, retrying.`);
+        } else {
+          LOG.error(`Request failed and retryFailures is false, will not retry.`);
+          removeEventsFromQueue(numberToSend);
+          executingQueue = false;
+        }
       }, connectionTimeout);
 
       const removeEventsFromQueue = (numberToSend: number): void => {
```
matus-tomlein, line 365: Can't comment in that function but should we check for `retryFailures` in `shouldRetryForStatusCode` function? I am not sure if that flag is taken into account if we get an error status code from the collector.

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -351,7 +352,14 @@ export function OutQueueManager(
...
       const xhrTimeout = setTimeout(function () {
         xhr.abort();
-        executingQueue = false;
+
+        if (retryFailures) {
+          LOG.warn(`Request failed, retrying.`);
+        } else {
+          LOG.error(`Request failed and retryFailures is false, will not retry.`);
+          removeEventsFromQueue(numberToSend);
+          executingQueue = false;
```
matus-tomlein, line 361: Might be missing something here, but should we set `executingQueue = false` also if `retryFailures` is true?

common/changes/@snowplow/browser-tracker/issue-1248-send-failure-retry-option_2023-10-20-16-06.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker",
+      "comment": "Retry control option",
```
matus-tomlein, line 5: Is this change item here intentional?

greg-el: Yeah, I had to make one since the API docs have changed. I've re-made the change file here with a more useful comment:
https://github.com/snowplow/snowplow-javascript-tracker/pull/1257/files#diff-69080c0b84c22e16c271e7e96183dcb2376f24014b73674c13101d66013f7f48R5

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -375,13 +385,19 @@ export function OutQueueManager(
...
-          } else {
-            if (!shouldRetryForStatusCode(xhr.status)) {
-              LOG.error(`Status ${xhr.status}, will not retry.`);
-              removeEventsFromQueue(numberToSend);
-            }
-            executingQueue = false;
+            return;
+          }
+
+          if (xhr.status === 0) {
```
matus-tomlein, line 391: Nit: can we use `else if` instead of the `return` and `if`? I think it's easier  to follow and makes it less likely we forget a `return` somewhere.

matus-tomlein, line 391: Why do we handle the status code 0 differently from other status codes? Shouldn't the `retryFailedRequests` also apply for 4xx and 5xx status codes (and override any custom retry code setting)? I would make it a part of `shouldRetryForStatusCode` function to check it – we do it similarly in the mobile trackers, [see here](https://github.com/snowplow/snowplow-ios-tracker/blob/master/Sources/Snowplow/Network/RequestResult.swift#L60-L63).

matus-tomlein (approved):
LGTM! 👍

PR #1258, opened 2023-10-20
Consider status codes <200 (close #1244)

### LINKED ISSUE #1244: Consider status codes <200
Only responses with codes >=200 are [considered by the tracker](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/libraries/browser-tracker-core/src/tracker/out_queue.ts#L374). This should be extended to consider status codes 0-199. Although it's probably unlikely we'll see a `1xx`, `0` is not-uncommonly used for fail states (cors error, timeout, connection abortion etc).

This would allow users to choose if they wish not to retry on codes 0-199 with the `dontRetryStatusCodes` option, which currently will do nothing for those values, and simply leave them in the `outQueue`.

[Looking](https://github.com/snowplow/snowplow-javascript-tracker/blame/master/libraries/browser-tracker-core/src/tracker/out_queue.ts#L374) [back](https://github.com/snowplow/snowplow-javascript-tracker/blame/250316f07a31e807d27ab6ca793ac928f5b5e98e/libraries/browser-tracker-core/src/tracker/out_queue.ts#L349) [through](https://github.com/snowplow/snowplow-javascript-tracker/blame/f5568539ac2b7b8f28b03c43430a276bcfbb282d/src/js/out_queue.js#L291) [history](https://github.com/snowplow/snowplow-javascript-tracker/blame/455bf7f1a143cdec648a06a073a719cde8c8a16d/src/js/out_queue.js#L189), I can't see any reason that below 200 was excluded specifically, it just seems to have "evolved" that way over time.

greg-el:
This PR allows status codes under 200 to be considered when passed into `shouldRetryForStatusCode`.

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/issue-1244-consider-status-codes-under-200_2023-10-20-16-20.json
+++ b/common/changes/@snowplow/browser-tracker-core/issue-1244-consider-status-codes-under-200_2023-10-20-16-20.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
+      "comment": "Consider status codes <200",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/browser-tracker-core"
+}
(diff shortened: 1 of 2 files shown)


matus-tomlein (approved):
LGTM!

PR #1263, opened 2023-10-27
Add support for callbacks on failed/successful event sending (close #…

greg-el:
This PR adds two optional callbacks to `newTracker`. These are `onRequestSuccess` and `onRequestFailure`. 
- `onRequestSuccess` fires when defined, and a response contains code 200-299 
-  `onRequestFailure` fires when defined, and a response contains any code that isn't retried, or on timeout

```ts
newTracker(..., {
    onRequestSuccess: (data: EventBatch) => foo(data),
    onRequestFailure: (data: RequestFailure) => bar(data),
});
```

Where `EventBatch` is:
```ts
export type PostBatch = Record<string, unknown>[];
export type GetBatch = string[];
export type EventBatch = GetBatch | PostBatch;
```

And `RequestFailure` is:
```ts
export type RequestFailure = {
  /** The batch of events that failed to send */
  events: EventBatch;
  /** The status code of the failed request */
  status?: number;
  /** The error message of the failed request */
  message?: string;
};
```
(`RequestFailure` could possibly be given a better name, any suggestions welcome)

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/out_queue.ts
+++ b/libraries/browser-tracker-core/src/tracker/out_queue.ts
@@ -1,38 +1,9 @@
-/*
- * Copyright (c) 2022 Snowplow Analytics Ltd, 2010 Anthon Pang
- * All rights reserved.
- *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions are met:
- *
- * 1. Redistributions of source code must retain the above copyright notice, this
- *    list of conditions and the following disclaimer.
- *
- * 2. Redistributions in binary form must reproduce the above copyright notice,
- *    this list of conditions and the following disclaimer in the documentation
- *    and/or other materials provided with the distribution.
- *
- * 3. Neither the name of the copyright holder nor the names of its
- *    contributors may be used to endorse or promote products derived from
- *    this software without specific prior written permission.
- *
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
- * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
- * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
- * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
- * FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
- * DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
- * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
- * CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
- * OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
- * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
- */
-
 import { attemptWriteLocalStorage, isString } from '../helpers';
 import { SharedState } from '../state';
 import { localStorageAccessible } from '../detectors';
 import { LOG, Payload } from '@snowplow/tracker-core';
 import { PAYLOAD_DATA_SCHEMA } from './schemata';
+import { EventBatch, RequestFailure } from './types';
 
 export interface OutQueue {
   enqueueRequest: (request: Payload, url: string) => void;
@@ -65,6 +36,8 @@ export interface OutQueue {
  * @param dontRetryStatusCodes – Failure HTTP response status codes from Collector for which sending events should not be retried
  * @param idService - Id service full URL. This URL will be added to the queue and will be called using a GET method.
  * @param retryFailedRequests - Whether to retry failed requests - Takes precedent over `retryStatusCodes` and `dontRetryStatusCodes`
+ * @param onRequestSuccess - Function called when a request succeeds
+ * @param onRequestFailure - Function called when a request does not succeed
  * @returns object OutQueueManager instance
  */
 export function OutQueueManager(
@@ -85,7 +58,9 @@ export function OutQueueManager(
   retryStatusCodes: number[],
   dontRetryStatusCodes: number[],
   idService?: string,
-  retryFailedRequests: boolean = true
+  retryFailedRequests: boolean = true,
+  onRequestSuccess?: (data: EventBatch) => void,
+  onRequestFailure?: (data: RequestFailure) => void
 ): OutQueue {
   type PostEvent = {
     evt: Record<string, unknown>;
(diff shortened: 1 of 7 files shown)


libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -230,14 +210,68 @@ export function OutQueueManager(
...
    */
   function sendPostRequestWithoutQueueing(body: PostEvent, configCollectorUrl: string) {
     const xhr = initializeXMLHttpRequest(configCollectorUrl, true, false);
-    xhr.send(encloseInPayloadDataEnvelope(attachStmToEvent([body.evt])));
+    const batch = attachStmToEvent([body.evt]);
+    currentBatch = batch as PostBatch;
+
+    // The timeout and onreadystatechange events are not set for oversized requests
+    // so we must set them here to listen for the response
+    setXhrCallbacks(xhr, 0);
```
matus-tomlein, line 235: This changes a bit the behaviour from how we handled oversize requests before:

1. adds a timeout – this is good in theory but in practice it doesn't make much difference because oversize requests are not blocking other requests from being executed (`executingQueue` remains false while they are executed).
2. sets `executingQueue` to `false` after the request ends – I think we don't want to do this unless we set `executingQueue` to true also for oversize requests. This can have the side effect that in case there is another parallel request being made (which may happen), `executingQueue` is set to false prematurely.

I think we have two options:

1. handle oversize requests just like any other requests – block other requests from being made while they are executed (set `executingQueue` to true) and add timeout for them.
2. keep the behaviour as is and don't set `executingQueue` to `true` for oversize request, but also don't set it to `false` after they are done (also don't add timeout for them).

I think I prefer option 2 just for the reason that change is not related with this PR and if we want to change how oversize requests are done, we should do it separately.

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -85,13 +58,20 @@ export function OutQueueManager(
...
 ): OutQueue {
   type PostEvent = {
     evt: Record<string, unknown>;
     bytes: number;
   };
 
+  // A reference to the current batch of events being sent
+  // This is copied into the onRequestSuccess and onRequestFailure callbacks when called,
+  // and then cleared after the callbacks are run
+  let currentBatch: EventBatch = [];
```
matus-tomlein, line 73: I don't really like that this is a "global" variable within the out queue manager. Could we instead just pass the current batch to the `setXhrCallbacks` function so that it can use it in the callbacks?

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -420,7 +422,9 @@ export function OutQueueManager(
           if (beaconStatus === true) {
             onPostSuccess(numberToSend);
```
matus-tomlein, line 423: Maybe in this case we should just call the onRequestSuccess as we won't get any more information from the beacon requests?

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -376,21 +393,6 @@ export function OutQueueManager(
         executeQueue();
       };
```
matus-tomlein, line 394: This `onPostSuccess` callback is now used only in one case – maybe we don't need it anymore? We can directly add the two function calls to the place where needed.

libraries/browser-tracker-core/src/tracker/types.ts
```diff
@@ -661,3 +631,33 @@ export interface ClientSession extends Record<string, unknown> {
...
+ * A collection of POST events which are sent to the collector.
+ * This will be a collection of JSON objects.
+ */
+export type PostBatch = Record<string, unknown>[];
+
+/**
+ * A collection of events which are sent to the collector.
+ * This can either be a collection of query strings or JSON objects.
+ */
+export type EventBatch = GetBatch | PostBatch;
```
matus-tomlein, line 651: To be honest, I am a bit skeptical about the usefulness of this information in the success and failure callbacks – what can users do with it? But maybe I'm just being a grumpy old man and probably it's better than nothing :)

greg-el: After discussion on Slack - the event batches will be staying.

libraries/browser-tracker-core/src/tracker/out_queue.ts
greg-el: This is now passed into setXhrCallbacks, and directly into `onreadystateupdate` for oversized payloads.

libraries/browser-tracker-core/src/tracker/out_queue.ts
greg-el: I've seperated out the `onreadystateupdate` callback for oversized payloads into its own thing which should be much more understandable now. All that happens now on completion is the firing of the callback for success/failure, if defined.

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -230,14 +205,72 @@ export function OutQueueManager(
     return typeof queue[0] === 'object' && 'evt' in queue[0];
   };
 
+  // Runs `onRequestFailure` callback if defined
+  const tryOnRequestFailure = (status: number, message: string, events: EventBatch, willRetry: boolean) => {
+    onRequestFailure?.({ events, status, message, willRetry });
```
matus-tomlein, line 210: Nit: is this function really worth having? Seems like it's just forwarding the values to the callback.

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -230,14 +205,72 @@ export function OutQueueManager(
...
+        } else {
+          tryOnRequestFailure(xhr.status, xhr.statusText, batch, false);
+        }
+      }
+    };
+
+    xhr.send(encloseInPayloadDataEnvelope(batch));
+  }
+
+  const removeEventsFromQueue = (numberToSend: number): void => {
```
matus-tomlein, line 235: Nit: Maybe use the `function ... () { .. }` syntax to match with the other functions here?

libraries/browser-tracker-core/src/tracker/out_queue.ts
```diff
@@ -418,9 +411,13 @@ export function OutQueueManager(
           // When beaconStatus is true, we can't _guarantee_ that it was successful (beacon queues asynchronously)
           // but the browser has taken it out of our hands, so we want to flush the queue assuming it will do its job
           if (beaconStatus === true) {
-            onPostSuccess(numberToSend);
+            removeEventsFromQueue(numberToSend);
+            executeQueue();
+            onRequestSuccess?.(batch);
```
matus-tomlein, line 416: I think it'd be good to be consistent about the order of execution of these actions – probably the callback should go before `executeQueue`. Above it's called even before removing events from queue.

matus-tomlein (approved):
LGTM, nice work! Just a few more nitpicks.

libraries/browser-tracker-core/src/tracker/out_queue.ts
greg-el: Oh yeah, that was from when I was trying to squeeze it under 15kb, that can go now 😄

libraries/browser-tracker-core/src/tracker/out_queue.ts
greg-el: Good point, this was moved from [here](https://github.com/snowplow/snowplow-javascript-tracker/pull/1263/files/f2b574fca59a1424e1fb361ef08b2a4fa51e3aea#diff-a57c80855fac59c2b4a6266d84fed6d821cfba1991cbfecc49924a515c54e942L363), since I didn't want to define even more functions in `exeuteQueue` and needed it in scope for `setXhrCallbacks`. Now it's defined outside a function we can use `function ... ()` 😄

PR #1265, opened 2023-11-11
Drop invalid client hints (close #1264)

### LINKED ISSUE #1264: Drop invalid client hints
[The client hints schema](https://github.com/snowplow/iglu-central/blob/master/schemas/org.ietf/http_client_hints/jsonschema/1-0-0) used follows exactly the [User-Agent Client Hints interface specification](https://wicg.github.io/ua-client-hints/#interface). 

The JS tracker, in order to attach the client hints entity, [currently checks whether `userAgentData` is there](https://github.com/snowplow/snowplow-javascript-tracker/blob/1df5e9f48411720a6449a8882665e490a58cbdbf/plugins/browser-plugin-client-hints/src/index.ts#L79), i.e. whether the API is implemented. However this means that user-agents who have implemented the API, **but** have not done so according to the spec, create invalid events, which however represent valid user behaviour.

**Expected behavior**
 
The JS tracker checks, at least briefly, the compliance of the implemented client-hints API before attaching the context entity to the event. If it finds that the user-agent is not compliant, it does not attach the client-hints context, i.e. an inaccurate implementation of the user-agent client hints interface has the same outcome as no implementation of it.

Example of invalid user-agent client hints:

```
{
    "brands": [
        {
            "brand": "Opera GX",
            "version": 98
        }
    ]
}
```

adatzer:


plugins/browser-plugin-client-hints/src/index.ts
```diff
@@ -63,35 +33,63 @@ let uaClientHints: HttpClientHints;
...
+ * https://wicg.github.io/ua-client-hints/#interface
+ */
+function validClientHints(hints: HttpClientHints): boolean {
+  if (!hints || typeof hints.isMobile !== 'boolean' || !Array.isArray(hints.brands)) {
+    return false;
+  }
+
+  if (
+    hints.brands.some(
+      (brand) => typeof brand.brand !== 'string' || typeof brand.version !== 'string' || Object.keys(brand).length !== 2
```
matus-tomlein, line 54: Nit: regarding the last condition that there have to be exactly 2 fields – wouldn't it be better to not take the `brands` array directly from the browser but map it and construct it on our side? This could make sure that in case the browser returns more than two properties, we can still work with the data if the attributes we need are present?

I mean something like:
```ts
const brands = hints.brands.map(b => {
   const { brand, version } = b;
   return { brand, version };
});
```

plugins/browser-plugin-client-hints/src/index.ts
```diff
@@ -63,35 +33,63 @@ let uaClientHints: HttpClientHints;
 function forceArray<T>(array: T[] | Record<string, T>): T[] {
   if (Array.isArray(array)) return array;
 
+  // Object.keys coerces to object
+  // null and undefined throw TypeError
```
matus-tomlein, line 37: I don't really understand this comment, could you expand on what is it for?

matus-tomlein (approved):
LGTM! Great that you added tests to the plugin as well!

plugins/browser-plugin-client-hints/src/index.ts
adatzer: Nice idea, i'll change so!

plugins/browser-plugin-client-hints/src/index.ts
adatzer: It is a note for callers of this function, since there is no type check here.

matus-tomlein (approved):

PR #1269, opened 2023-11-24
Add OTB Button Click Tracking (close #1267)

### LINKED ISSUE #1267: Add OTB Button Click Tracking

greg-el:
This PR adds a plugin for automatic button click tracking, similar to the existing link click tracking.

It provides a single method, `enableButtonClickTracking`. Which accepts a couple of options:

### `enableButtonClickTracking.filter`

This allows the user to provide either an allowlist, denylist, or a custom callback, that will only track a click if the element that has been clicked on passes the filter, for example of a function filter that will accept the following button:

```html
<button id="my-cool-button">Click Here</button>
```

```TS
filter: (elem: HtmlElement) => elem.id === "my-cool-button")
```

### `enableButtonClickTracking.context`

Allows attaching a [dynamic entity](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/mobile-trackers/custom-tracking-using-schemas/global-context/#dynamic-entities) to the button click events.

note: The integration tests won't pass on CI, as the button click schema isn't on Iglu Central yet (see https://github.com/snowplow/iglu-central/compare/master...release/r150)

### DIFF
--- a/plugins/browser-plugin-button-click-tracking/src/api.ts
+++ b/plugins/browser-plugin-button-click-tracking/src/api.ts
@@ -1,0 +1,108 @@
+import { BrowserPlugin, BrowserTracker, dispatchToTrackersInCollection } from '@snowplow/browser-tracker-core';
+import { DynamicContext, CommonEventProperties, resolveDynamicContext } from '@snowplow/tracker-core';
+
+import { createEventFromButton, filterFunctionFromFilter } from './util';
+import { buildButtonClick } from './core';
+import { ButtonClickEvent, ButtonClickTrackingConfiguration, FilterFunction } from './types';
+
+const _trackers: Record<string, BrowserTracker> = {};
+
+// The event listeners added to the document, for each tracker
+// This allows them to be removed with `removeEventListener` if button click tracking is disabled
+const _listeners: Record<string, (event: MouseEvent) => void> = {};
+
+/**
+ * Button click tracking
+ *
+ * Will automatically tracking button clicks once enabled with 'enableButtonClickTracking'
+ * or you can manually track button clicks with 'trackButtonClick'
+ */
+export function ButtonClickTrackingPlugin(): BrowserPlugin {
+  return {
+    activateBrowserPlugin: (tracker: BrowserTracker) => {
+      _trackers[tracker.id] = tracker;
+    },
+  };
+}
+
+/**
+ * Manually log a click
+ *
+ * @param event - The event information
+ * @param trackers - The tracker identifiers which the event will be sent to
+ */
+export function trackButtonClick(
+  event: ButtonClickEvent & CommonEventProperties,
+  trackers: Array<string> = Object.keys(_trackers)
+) {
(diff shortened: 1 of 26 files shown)


plugins/browser-plugin-button-click-tracking/src/api.ts
```diff
@@ -0,0 +1,109 @@
...
+  trackers: Array<string> = Object.keys(_trackers)
+) {
+  dispatchToTrackersInCollection(trackers, _trackers, (t) => {
+    t.core.track(buildButtonClick(event), event.context, event.timestamp);
+  });
+}
+
+/**
+ *
+ * Enable automatic button click tracking
```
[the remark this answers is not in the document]
matus-tomlein: Also please mention the `sp-button-label` way to set the button label somewhere!

plugins/browser-plugin-button-click-tracking/src/api.ts
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @param configuration  - The configuration for automatic button click tracking
+ * @param trackers - The tracker identifiers which the event will be sent to
+ */
+export function enableButtonClickTracking(
```
matus-tomlein, line 53: What it be too much of a hassle to add `disableButtonClickTracking`?
With SPAs people are now more keen to enable and disable tracker features only for a certain period of the page lifecycle. We have had to add a disable activity tracking feature recently for instance.

plugins/browser-plugin-button-click-tracking/src/util.ts
```diff
@@ -0,0 +1,65 @@
+import { CommonEventProperties } from '@snowplow/tracker-core';
+import { ButtonClickEvent } from './core';
+
+export type AllowList = { allowlist: string[] };
```
matus-tomlein, line 4: Could we add API docs here to explain these types? Also make it clear that the allowlist is a list of classes.

matus-tomlein (commented):
Very nice implementation, great work! 👍

plugins/browser-plugin-button-click-tracking/src/api.ts
greg-el: Added!

plugins/browser-plugin-button-click-tracking/src/api.ts
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Disable automatic click tracking for all `<button>` and `<input type="button">` elements on the page
+ *
+ * Can be re-enabled with {@link enableButtonClickTracking}
+ */
+export function disableButtonClickTracking() {
+  for (const trackerId in _trackers) {
+    document.removeEventListener('click', _config[trackerId].listener);
+  }
```
matus-tomlein, line 84: Can we also clear the `_config`? Just in case someone tries calling the method twice so that it doesn't try to remove the listener twice.

matus-tomlein (approved):
LGTM! 👍

PR #1273, opened 2023-12-12
Update web-vitals to accept custom entities

jethron:


igneel64 (approved):
Nice work.
I am thinking that maybe we would need that, or similar capability, for all plugins moving forward so the `context` parameter would be available in all plugins.

Even if we just change it for this requested case, we can implement it later on in a more universal way still as a minor change.

just ccing @matus-tomlein if he has a better idea for an API

matus-tomlein (approved):
LGTM!

Agree with Peter that it'd be nice to have a unified way for all plugins to attach the dynamic context property, but they do have slightly different initialization flows, not sure how we can improve this in a non-breaking way (improvements on the level of global context are the only thing that comes to my mind). Unifying the plugin initialization in v4 as we discussed would definitely be nice though!

PR #1274, opened 2023-12-12
Do not set the previous session ID reference in cookies if anonymous tracking is enabled (close #1268)

### LINKED ISSUE #1268: Do not set the previous session ID reference in cookies if anonymous tracking is enabled
As [reported here](https://discourse.snowplow.io/t/is-sp-id-cookie-a-valid-alternative-to-the-client-session-context/9362/4), in case anonymous tracking with session tracking is enabled, the tracker still stores the reference to the previous session ID in the `_sp_id` cookie. While this is not directly tracked in the event, the information is present in the collector request header and could be used by the cookie enrichment to, in a way, reconstruct the user ID by stitching together the sessions.

In case anonymous tracking is enabled, the tracker should not set the previous session ID reference in cookies.

matus-tomlein:
Issue #1268 

Addresses the issue where the previous session ID reference was updated in the cookies even in case anonymous tracking (with session tracking was enabled).

This was a problem, because the cookie value is sent as an HTTP header along with requests. It allowed one to use a cookie enrichment to inspect the value and use the previous session ID reference to basically reconstruct the user identifier (join together their previous sessions).

To solve the issue, I updated the function that serializes the cookie value to accept a flag whether anonymous tracking is enabled. In case it is enabled, we skip serializing the domain user ID and the previous session ID.

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/issue-1268-previous_session_id_anonymous_tracking_2023-12-12-13-35.json
+++ b/common/changes/@snowplow/browser-tracker-core/issue-1268-previous_session_id_anonymous_tracking_2023-12-12-13-35.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
+      "comment": "Do not set the previous session ID reference in cookies if anonymous tracking is enabled (#1268)",
+      "type": "none"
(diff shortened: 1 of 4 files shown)


igneel64 (approved):

greg-el (approved):
LGTM!

PR #1281, opened 2024-01-10
Support clicks on child elements in button click tracking (close #1280)

### LINKED ISSUE #1280: Support clicks on child elements in button click tracking
In case a button has child elements (e.g., a span or svg with an icon), button click tracking does not recognize clicks on these elements.

We should update the button click tracking to also recognize clicks on child elements of buttons.

matus-tomlein:
Issue #1280

This reopens the PR #1279 from @jethron  that fixes the issue where clicks on buttons with child elements are not recognized.


greg-el (approved):
LGTM!

Issue #1259, opened 2023-10-24
getDomainUserId void return type 


jordmccord:
**Describe the bug**
The return type of `getDomainUserId` is `void` when it should be `string | null` (presuming you can't set the id as a number)

**To Reproduce**
Try and check if the `getDomainUserId()` is set in an if statement, you get the following type error `An expression of type 'void' cannot be tested for truthiness.`. 

**Expected behavior**
You should be able to check if the value is set and use it as a string type if it set.

**Screenshots**
![Screenshot 2023-10-24 at 10 19 22](https://github.com/snowplow/snowplow-javascript-tracker/assets/7591840/a37e0068-2808-47ac-b514-f529548d9334)

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
N/A

**Additional context**
getUserId, getCookieName also have the same issue

PR #1284, opened 2024-01-25
BrowserTracker: fix method type definitions

jethron:
Per the [TypeScript docs](https://www.typescriptlang.org/docs/handbook/2/functions.html#return-type-void), functions that return values are type-compatible with functions declared as having void return type; however if you're consuming these type definitions this becomes annoying as you are required to cast `as unknown as string` at the call sites to get the correct types actually returned from these methods. Because the types are compatible tsc never reported any issue with these methods.

### DIFF
--- a/api-docs/docs/browser-tracker/browser-tracker.api.md
+++ b/api-docs/docs/browser-tracker/browser-tracker.api.md
@@ -75,7 +75,6 @@ export interface BrowserTracker {
     getCookieName: (basename: string) => string;
     getDomainSessionIndex: () => number;
     getDomainUserId: () => string;
-    // Warning: (ae-forgotten-export) The symbol "ParsedIdCookie" needs to be exported by the entry point index.module.d.ts
     getDomainUserInfo: () => ParsedIdCookie;
     getPageViewId: () => string;
     getTabId: () => string | null;
@@ -267,6 +266,21 @@ export interface PageViewEvent {
     contextCallback?: (() => Array<SelfDescribingJson>) | null;
     title?: string | null;
 }
+
+// @public
+export type ParsedIdCookie = [
+cookieDisabled: string,
+domainUserId: string,
+cookieCreateTs: number,
+visitCount: number,
+nowTs: number,
+lastVisitTs: number | undefined,
+sessionId: string,
+previousSessionId: string,
+firstEventId: string,
+firstEventTs: number | undefined,
+eventIndex: number
+];
 
 // @public (undocumented)
 export type Platform = "web" | "mob" | "pc" | "srv" | "app" | "tv" | "cnsl" | "iot";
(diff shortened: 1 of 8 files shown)


igneel64 (approved):
💯

libraries/browser-tracker-core/src/tracker/types.ts
```diff
@@ -1,3 +1,4 @@
+import { ParsedIdCookie } from './id_cookie';
```
matus-tomlein, line 1: Should we move the `ParsedIdCookie` type here to avoid the circular import of types between `types.ts` and `id_cookie.ts`?

api-docs/docs/browser-tracker/browser-tracker.api.md
```diff
@@ -72,13 +72,14 @@ export interface BrowserTracker {
...
     enableAnonymousTracking: (configuration?: EnableAnonymousTrackingConfiguration) => void;
     flushBuffer: (configuration?: FlushBufferConfiguration) => void;
-    getCookieName: (basename: string) => void;
-    getDomainSessionIndex: () => void;
-    getDomainUserId: () => void;
-    getDomainUserInfo: () => void;
+    getCookieName: (basename: string) => string;
+    getDomainSessionIndex: () => number;
+    getDomainUserId: () => string;
+    // Warning: (ae-forgotten-export) The symbol "ParsedIdCookie" needs to be exported by the entry point index.module.d.ts
```
matus-tomlein, line 78: Shouldn't we export the `ParsedIdCookie` as part of this PR?

matus-tomlein (commented):
This is great!

My concern is whether this should wait for the v4 release or can go out in v3? I think this might be a breaking change and can produce type errors for users who previously worked around the `void` return type by type casting the return values. What do you think?

Also please reference issue #1259 in a close parentheses in the commit message when merging.

api-docs/docs/browser-tracker/browser-tracker.api.md
jethron: Good call! I'd left it because that warning was already there for other unexported types, and it involved a change to the tracker itself rather than just the `-core` library. I have instances where the type is useful though, so have exported it from both.

libraries/browser-tracker-core/src/tracker/types.ts
jethron: Good catch! Also migrated it to a labelled tuple which is hopefully a bit easier to work with.

jethron:
> My concern is whether this should wait for the v4 release or can go out in v3? I think this might be a breaking change and can produce type errors for users who previously worked around the `void` return type by type casting the return values. What do you think?

I'm not sure this would really happen: `void` isn't castable to any other type AFAIK, so users would have likely had to use either: `as unknown as T`, `as any`, or `as undefined` (though I don't think this would be useful, so unlikely); casting straight from `void` to anything more specific would have caused errors by itself. So any cast, even if it was incorrect, would have been erased by `unknown`/`any` and will avoid an error even if that cast isn't compatible with these changes.

So I think it's safe for V3, but can be persuaded otherwise if you have an example where this is problematic?

matus-tomlein (approved):
That's a good point! I didn't realize that one would need to cast the types to `unknown` first anyway and thus reset the type, but that does make sense. In that case I think we should be good for a v3 release, can't think of any scenario where this would be breaking.

PR #1285, opened 2024-01-25
Always access `navigator` via `window`

jethron:
This one is pretty niche, sorry. :sweat_smile:  I'm implementing the browser-tracker in a Shopify web-pixel's [strict sandbox](https://shopify.dev/docs/apps/marketing/pixels#strict-sandbox-web-pixel-app-extension).

In this environment, the typical globals (`window`, `document`, `navigator`, `screen`, etc.) are not defined, but copies of their more essential data are passed through to the sandboxed environment in the non-global "context" they provide via their APIs. Using this data is enough to manually recreate the globals before asynchronously importing the browser-tracker module and start sending events - with the exception of `navigator`. For some reason in the sandbox Shopify essentially do:

```javascript
Object.defineProperty(globalThis, "navigator", {value: undefined, configurable: false, writable: false});
```
Which means I can't define `navigator` in the global scope (I _can_ define `window`, and `window.navigator`, though).
Because the tracker tries to access properties of the global `navigator` on creation it instantly triggers a script error and does not function. By making the tracker consistently access `window.navigator` instead, I can avoid this error and use the tracker as intended, since I can ensure `window.navigator` has the expected values the tracker is looking for.

This change just makes the tracker consistently access `window.navigator` (as was already the case in some places) instead of `navigator` directly.
Alternatively, we could make sure `navigator` is defined before accessing it, but in this case that would result in not collecting that data, which is less preferable.

This also fixes a typescript nag about `window.navigator.sendBeacon` always being truthy.

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/navigator-access_2024-01-25-03-04.json
+++ b/common/changes/@snowplow/browser-tracker-core/navigator-access_2024-01-25-03-04.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
+      "comment": "Consistently access navigator via window.navigator",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/browser-tracker-core"
+}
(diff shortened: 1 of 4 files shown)


igneel64 (approved):

matus-tomlein (approved):
LGTM!

Issue #1291, opened 2024-03-05
Export snowplow-ecommerce types for wrapper creation


igneel64:
**Is your feature request related to a problem? Please describe.**
We are trying to create wrappers for the snowplow ecommerce browser plugin and would like to expose typings from the package to create these wrappers in TypeScript.

PR #1293, opened 2024-03-07
Release/3.22.0

igneel64:
- #1284 
- #1291

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/bt-typedefs_2024-01-30-01-31.json
+++ b/common/changes/@snowplow/browser-tracker-core/bt-typedefs_2024-01-30-01-31.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
+      "comment": "",
+      "type": "none"
+    }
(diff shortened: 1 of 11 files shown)


common/changes/@snowplow/browser-tracker-core/bt-typedefs_2024-01-30-01-31.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
+      "comment": "",
```
matus-tomlein, line 5: Is it a problem that there is no comment in these changes?

matus-tomlein (approved):
LGTM!

common/changes/@snowplow/browser-tracker-core/bt-typedefs_2024-01-30-01-31.json
igneel64: This will be used in the generated changelog _(which I think we don't use anyway)_
I don't think the empty entry will make a difference _(think :P )_

PR #1294, opened 2024-03-07
Cache browser properties and use the ResizeObserver to update when changed (close #1295)

### LINKED ISSUE #1295: Function `getBrowserProperties` is too expensive
**Describe the bug**
My company is currently working to improve INP (Interaction to Next Paint) metrics all over our e-commerce platform. During our analysis, we see `track()` calls in snowplow that just consume too much time executing. 

**To Reproduce**
Create a performance profile of an interaction where tracking is involved with a CPU throttling of 4x - 6x. Zoom into the interaction to see the callstack, as example like this:
before
-> The `track()` function can eat up to 30% of the entire interaction task! This means, the user has to wait for the `track()` call to be finished until he sees the result he wanted in the browser.

**Expected behavior**
The `track()` function should be as fast as possible, that no one has to think about it's impact regarding performance.

 - OS: MacOS Sonoma 14.3.1
 - Browser: Chrome 
 - Version: 119.0.6045.105

**Additional context**
I have found the reason why the `track()` function is expensive. It is related with the internal `beforeTrack()` method which calls `getBrowserProperties()`. While getting these browser properties, very expensive browser calls are involved like `document.body.clientHeight`or similar. I'm currently working on a change that would fix that by involving a `ResizeObserver` approach, moving the expensive calls to another place in the call stack where it doesn't affect the `track()` call anymore.

I would love to hear from you guys what you think about my approach and if this is feasible. Check out my PR to see a "before" and "after" comparison of the measured flame graph:  https://github.com/snowplow/snowplow-javascript-tracker/pull/1294

rvetere:
Hey everyone!

I've been diving into how we can make our track() function smoother and faster, and I've found something interesting. When I checked out how it's affecting our website's performance, especially with the new [INP metrics](https://web.dev/articles/inp) that Google is rolling out, I noticed that track() calls were taking up a whopping 30% of the time during some interactions. This slows down the experience for our users, which is something we definitely want to avoid.

The problem is caused by `getBrowserProperties` where expensive browser calls are involved like e.g. "document.body.clientWidth". Synchronously grab these values in every `beforeTrack()` is a performance impact we do not want. 

The viewport or device resolution is unlikely to change between every single track call. I think this could be greatly improved by involving a `ResizeObserver`. The observer would listen for changes, when we actually have to grab/update these browser properties - and the `getBrowserProperties()` function itself would only use the last cached values.

Let me shortly show my analysis about the call stack. 

First, we look at a flame-graph as it is today:
before
-> Here we see, the `track()` call eats up to 30% of the entire input handler with a CPU throttling of 6x (on my M1 Max 64GB)

And the same again with the new `ResizeObserver` approach:
after
-> The execution time of the `track()` call comes down to ~6% of the entire input handler, still with 6x CPU throttling. 

My tests show that i still get the exact same data for the properties `viewport`, `documentSize` and `resolution`.

Would love to get your feedback, we really want to improve the library itself and not just create a patch for the package here at Digitec Galaxus AG.

Todos:

- [x] Only return cached browser properties in `getBrowserProperties()` - to make it a cheap and fast call
- [x] Detect if browser can use `ResizeObserver`, fallback to old approach if not available
- [x] If ResizeObserver is available, add an instance to update cached browser properties on observed changes
- [ ] Write tests (🚨 Doesn't make sense right now as this would involve an entire e2e stack)
- [x] Cleanup method to destroy ResizeObserver instance again (🚨 as there is no `destroy` method in `Tracker` class, we cannot implement this)
- [ ] Consider a debounce for updating the cache as it can happen very often in short time (🚨 I don't want to defer the update of the properties even further - by doing this, we increase the chance to be "out-of-sync" with the browser properties at the moment of `track()` function)

Fixes #1295

### DIFF
--- a/libraries/browser-tracker-core/src/helpers/browser_props.ts
+++ b/libraries/browser-tracker-core/src/helpers/browser_props.ts
@@ -1,8 +1,17 @@
+let cachedProperties: any = null;
+
 /* Separator used for dimension values e.g. widthxheight */
 const DIMENSION_SEPARATOR = 'x';
 
 export function getBrowserProperties() {
-  return {
+  if (!cachedProperties) {
+    updateBrowserProperties();
+  }
+  return cachedProperties;
+}
+
+function updateBrowserProperties() {
+  cachedProperties = {
     viewport: floorDimensionFields(detectViewport()),
     documentSize: floorDimensionFields(detectDocumentSize()),
     resolution: floorDimensionFields(detectScreenResolution()),
@@ -17,6 +26,20 @@ export function getBrowserProperties() {
     hardwareConcurrency: (window.navigator as any).hardwareConcurrency,
   };
 }
+
+// Initialize the ResizeObserver
+const resizeObserver = new ResizeObserver((entries) => {
+  for (let entry of entries) {
+    // Check if the observed entry is for document's body or root element
+    if (entry.target === document.body || entry.target === document.documentElement) {
+      updateBrowserProperties(); // Update the cache if there's a change
+    }
+  }
+});
+
+// Start observing the document's body and root element for size changes
+resizeObserver.observe(document.body);
+resizeObserver.observe(document.documentElement);
 
 /**
  * Gets the current viewport.
--- a/libraries/browser-tracker-core/src/helpers/browser_props.ts
+++ b/libraries/browser-tracker-core/src/helpers/browser_props.ts
@@ -1,17 +1,49 @@
-let cachedProperties: any = null;
+interface BrowserProperties {
+  viewport?: string | null;
+  documentSize?: string | null;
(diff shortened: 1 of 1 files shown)


snowplowcla:

rvetere:
I signed it!

snowplowcla:
Confirmed! @rvetere has signed the Contributor License Agreement. Thanks so much.

matus-tomlein (approved):
Thanks again for the contribution @rvetere! This is a huge improvement and, based on my testing so far, seems to work just as well as the previous implementation!

We will do some more testing in CI on all supported browsers. I'll merge this as is and we will take on any subsequent work necessary. We plan to release this in the next minor release of the tracker!

PR #1296, opened 2024-03-12
Fix snowplow ecommerce typename conflicts

igneel64:
Mistakenly it was not detected that the Snowplow ecommerce plugin exposes already two types `Promotion` and `Transaction` from the universal analytics ecommerce wrapper functionality. This overrides the definitions from other types exposed at 3.22.0.

To mitigate this we can rename the types `Promotion` and `Transaction` from the Snowplow Ecommerce package to `SPPromotion` and `SPTransaction`.

This would be bumped on a patch version.


matus-tomlein (approved):
LGTM!

Btw, I am waffling whether `EcommercePromotion` (and transaction) would be better than `SPPromotion` but not sure..

igneel64:
@matus-tomlein Shared the same thought. Reason I use the SP prefix is because other solutions (GA4 and UA) can be regarded as `Ecommerce`.

PR #1297, opened 2024-03-12
Release/3.22.1

igneel64:
- #1296


matus-tomlein (approved):
LGTM!

Seems like that time of year when we have to fix the integration tests for the media plugins again... But the failures seem nothing related to the changes here, so I think it's fine to continue with the release.

PR #1299, opened 2024-03-20
Add Event Specification option map to Snowplow Media Plugin

igneel64:
Proposal for adding a Snowplow Media Plugin option for event specification.

E.g.
```ts
      SnowplowMediaPlugin({ eventSpecificationIds: { 'play_event': 'test' } }),
```

If the proposal seems okay, we can move this away from Draft.

### DIFF
--- a/plugins/browser-plugin-media/src/api.ts
+++ b/plugins/browser-plugin-media/src/api.ts
@@ -21,23 +21,29 @@ import {
   MediaTrackErrorArguments,
   MediaTrackSelfDescribingEventArguments,
   EventWithContext,
+  MediaPluginOptions,
 } from './types';
+import { setEventSpecificationContext } from './utils';
 
 export { MediaAdBreakType as MediaPlayerAdBreakType };
 
 const _trackers: Record<string, BrowserTracker> = {};
 const _context: Record<string, SelfDescribingJson[]> = {};
+const defaultPluginOptions = {};
 
 /**
  * Adds media tracking
  */
-export function SnowplowMediaPlugin(): BrowserPlugin {
+export function SnowplowMediaPlugin(pluginOptions: MediaPluginOptions = defaultPluginOptions): BrowserPlugin {
   let trackerId: string;
   return {
     activateBrowserPlugin: (tracker) => {
       trackerId = tracker.id;
       _trackers[trackerId] = tracker;
       _context[trackerId] = [];
+    },
+    beforeTrack(payloadBuilder) {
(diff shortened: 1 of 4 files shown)


plugins/browser-plugin-media/src/api.ts
```diff
@@ -21,24 +21,30 @@ import {
...
 
 const _trackers: Record<string, BrowserTracker> = {};
 const _context: Record<string, SelfDescribingJson[]> = {};
+const defaultPluginOptions = {};
 
 /**
  * Adds media tracking
  */
-export function SnowplowMediaPlugin(): BrowserPlugin {
+export function SnowplowMediaPlugin(pluginOptions: MediaPluginOptions = defaultPluginOptions): BrowserPlugin {
```
matus-tomlein, line 37: Why not have this be a part of the `MediaTrackingConfiguration` object that is passed to the `startMediaTracking` function? Unless there is a reason we need to have it here, I think it might be better to have only one place to configure the plugin?

plugins/browser-plugin-media/src/api.ts
```diff
@@ -21,24 +21,30 @@ import {
...
   let trackerId: string;
   return {
     activateBrowserPlugin: (tracker) => {
       trackerId = tracker.id;
       _trackers[trackerId] = tracker;
       _context[trackerId] = [];
     },
+    beforeTrack(payloadBuilder) {
```
matus-tomlein, line 45: This is fine but seems a bit of a roundabout way for adding the information. All events are constructed in the `track()` function in this file – look for the `dispatchToTrackersInCollection()` calls at the end of this file (line 711). Wouldn't it be better to add the information there?

matus-tomlein (commented):
I think this makes sense! Just had a few comments about considering a different structure for the configuration and how the entities are attached, but this seems fine as well.

One more thing I wanted to suggest – does this need to be a part of the media plugin? Looking at what we need to do here, there is nothing media-specific here. It just adds a event specification ID depending on some definition of events.

What I'm trying to say is, could we instead have a more general plugin that could be used to add the event specification ID to automatically tracked events from any plugin (e.g., performance, consent, ...)? The plugin configuration could be autogenerated from Snowtype/Console and could contain ES IDs for different scenarios (e.g., event A with entity B and C). That would seem to me more flexible and less disruptive to existing code.

matus-tomlein:
Can we close this in favour of #1300?

igneel64:
Yeap done!

PR #1300, opened 2024-03-25
Add the Event Specifications plugin

igneel64:


plugins/browser-plugin-event-specifications/src/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  const options = { ...defaultPluginOptions, ...pluginOptions };
+  const enabledIntegrations = Object.keys(integrations).filter(
+    (integration) => options[integration as AvailablePlugins]
+  );
+  let trackerId: string;
+  return {
+    activateBrowserPlugin: (tracker) => {
+      trackerId = tracker.id;
+      _trackers[trackerId] = tracker;
+      return;
```
matus-tomlein, line 25: Nit: extra return

matus-tomlein (commented):
I like this, seems like a nice solution!

Just one question, did you consider having a typed interface for the `EventSpecificationsPluginOptions` constructor parameter? Currently users can pass `Record<string, string>` for the specific plugin events, which opens room for typos (e.g., `play_evnt`) and doesn't show the users what are the available options. It is more flexible for us this way, so it's a trade-off, just wanted to hear your thoughts on that.

igneel64:
> Just one question, did you consider having a typed interface for the EventSpecificationsPluginOptions constructor parameter? Currently users can pass Record<string, string> for the specific plugin events, which opens room for typos (e.g., play_evnt) and doesn't show the users what are the available options. It is more flexible for us this way, so it's a trade-off, just wanted to hear your thoughts on that.

Good call! 
The thinking is that this plugin input will never be typed by the user. It will be retrieved directly from the console like the following:
image

This is also mentioned in the README and will be added in the documentation page as well.
> This configuration should be retrieved directly from your Data Product in the [Snowplow BDP Console](https://console.snowplowanalytics.com).

plugins/browser-plugin-event-specifications/src/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    },
+    beforeTrack(payloadBuilder) {
+      enabledIntegrations.forEach((integrationKey) => {
+        const matchedEvent = integrations[integrationKey as AvailablePlugins].detectMatchingEvent(payloadBuilder);
+        if (!matchedEvent) {
+          return;
+        }
+        const matchingPluginEventMap = Object.values(options).find(
```
matus-tomlein, line 33: Why do we consider all integrations here? For instance if the user configured the plugin like this:

```ts
EventSpecificationPlugin({
  OtherPlugin: { play_event: "wrong" },
  SnowplowMediaPlugin: { play_event: "wrong" },
})
```

And the current event was matched against the media plugin, wouldn't this code result in the `wrong` ES ID being taken?

This can't happen now since we only have one integration, but if there is such a problem, it would be good to solve it before we add more integrations.

matus-tomlein (approved):
That makes sense! If it's going to be primarily used by copying from Console, then it should be fine this way.

Except for the one potential problem that I commented, this LGTM!

plugins/browser-plugin-event-specifications/src/index.ts
igneel64: Yeap, fixed:

```ts
const matchingPluginEventMap = options[integrationKey as AvailablePlugins];
const matchingEventSpecificationId = matchingPluginEventMap && matchingPluginEventMap[matchedEvent];
if (!matchingEventSpecificationId) {
    return;
}
```

Plus both integrations and options can only have `AvailablePlugins` as keys

matus-tomlein: looks good!

Issue #1125, opened 2022-10-25
Option to reset pageViewId.


VIKIVIKA:
**Describe the bug**

If we create multiple tracker's using newTracker, and call window.snowplow("trackPageView") for the first time,

Snowplow creates similar pageViewIDs for all the pageViews,

where as for further request's if window.snowplow("trackPageView") is called, 
we get different pageViewIds for different tracker's available.

**To Reproduce**
Consider the Following,

```
<script>
	;(function(p,l,o,w,i,n,g){if(!p[i]){p.GlobalSnowplowNamespace=p.GlobalSnowplowNamespace||[]; p.GlobalSnowplowNamespace.push(i);p[i]=function(){(p[i].q=p[i].q||[]).push(arguments) };p[i].q=p[i].q||[];n=l.createElement(o);g=l.getElementsByTagName(o)[0];n.async=1; n.src=w;g.parentNode.insertBefore(n,g)}}(window,document,"script","https://cdnjs.cloudflare.com/ajax/libs/snowplow/2.18.2/sp.min.js","snowplow"));
		
		snowplow('newTracker', 'tracker1', 'collector.tracker1.com', initOptions);
		snowplow('newTracker', 'tracker2, 'collector.tracker1.com', initOptions);
</script>

After this first step if we run `snowplow('trackPageView');`, **tracker1** and **tracker2** will have same pageViewID's.

but after the first trackPageView, if we run again the same **snowplow('trackPageView');**,  **tracker1** and **tracker2** will have different pageViewID's respectively.

Above example is an use case for a single page application, where in we are calling trackPageView on page load as well as when a user navigates to different route without reloading the page.

```	

**Expected behavior**

Should have same page view id's if we call trackPageView in case of multiple tracker's or 

an option should be exposed like how we have for **preservePageViewId**, similarly we can use **resetPageViewId** in combination with **preservePageViewId**, so pageView id will be same across and can be reset by the user whenever requried.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

igneel64:
Hello @VIKIVIKA , thanks for reporting this, we will be taking a look and update you soon :) 

In the meantime a codesandbox with the reproduction would be really helpful. Cheers!

VIKIVIKA:
@igneel64 here is the sandbox with reproducible code,

https://codesandbox.io/s/withered-cache-xjio6q?file=/src/App.js

[...]

PR #1308, opened 2024-05-02
Add an option to generate the page view ID according to changes in the page URL to account for events tracked before page views in SPAs (close #1307 and #1125)

### LINKED ISSUE #1307: Add an option to generate the page view ID according to changes in the page URL to account for events tracked before page views in SPAs
In SPAs, as the user is navigating through pages, it may happen that certain events belonging to a page are tracked before the page view event. For example, auto-tracked events relating to the experiment selection in A/B testing may be tracked before the actual page view.
We would want these events to share the same page view ID as the page view event that is tracked for the same page URL even though that is tracked later.

Currently the page view ID is generated at the time of the page view event, so all events tracked before it have a different page view ID (this is not the case for the very first page view event, but is for the second and subsequent events).

Generate the page view ID according to the current URL of the page. In case the URL changes, assign all events coming from the new URL the same page view ID.

matus-tomlein:
Issue #1307

## Problem

In SPAs, as the user is navigating through pages, it may happen that certain events belonging to a page are tracked before the page view event. For example, auto-tracked events relating to the experiment selection in A/B testing may be tracked before the actual page view.
We would want these events to share the same page view ID as the page view event that is tracked for the same page URL even though that is tracked later.

Currently the page view ID is generated at the time of the page view event, so all events tracked before it have a different page view ID (this is not the case for the very first page view event, but is for the second and subsequent events).

## Solution

This PR adds an option called `preservePageViewIdForUrl` to the tracker configuration which has the following settings:

* `false` (default) – keep the current behaviour (i.e., new page view ID is generated when new page view event is tracked)
*  `true` or `'full'` – the page view ID will be kept the same for all page views with that exact URL (even for events tracked before the page view event).
* `'pathname'` – the page view ID will be kept the same for all page views with the same pathname (search params or fragment may change).
* `'pathnameAndSearch'` – the page view ID will be kept the same for all page views with the same pathname and search params (fragment may change).

Thus, if this option is configured to something else than `false`, events that happen on the second page but precede the second page view will have the same page view ID as the second page view. To illustrate, it will work as follows:

1. User navigates to page `http://example.com/a`
2. Custom event A is tracked with page view ID 1
3. Page view event is tracked with page view ID 1
4. Custom event B is tracked with page view ID 1
5. User clicks on a link that takes them to `http://example.com/b`
6. Custom event C is tracked with page view ID 2 (this is what this PR enables, previously it would have page view ID 1)
7. Page view event is tracked with page view ID 2
8. Custom event D is tracked with page view ID 2

This PR also fixes issue #1125 which reported the problem that if two trackers are initialized on the same page, the second page view event will have different page view IDs for each of them (the first will have the same). If this new option is configured, the second page view event will have the same page view ID on both trackers.

cc @davidher-mann and @j7i – we discussed this problem on our call, would be great to get your feedback if this solution fixes the issue for you!

### DIFF
--- a/.bundlemonrc.json
+++ b/.bundlemonrc.json
@@ -7,7 +7,7 @@
     },
     {
       "path": "./trackers/browser-tracker/dist/index.umd.min.js",
-      "maxSize": "15.5kb",
+      "maxSize": "16kb",
       "maxPercentIncrease": 10
     },
     {
@@ -22,7 +22,7 @@
     },
     {
       "path": "./libraries/browser-tracker-core/dist/index.module.js",
-      "maxSize": "27kb",
+      "maxSize": "28kb",
       "maxPercentIncrease": 10
     },
     {
--- a/api-docs/docs/browser-tracker/browser-tracker.api.md
+++ b/api-docs/docs/browser-tracker/browser-tracker.api.md
@@ -83,6 +83,7 @@ export interface BrowserTracker {
     namespace: string;
     newSession: () => void;
     preservePageViewId: () => void;
+    preservePageViewIdForUrl: (preserve: PreservePageViewIdForUrl) => void;
     setBufferSize: (newBufferSize: number) => void;
     setCollectorUrl: (collectorUrl: string) => void;
     setCookiePath: (path: string) => void;
@@ -291,6 +292,9 @@ export type PostBatch = Record<string, unknown>[];
 // @public
 export function preservePageViewId(trackers?: Array<string>): void;
 
+// @public (undocumented)
+export type PreservePageViewIdForUrl = boolean | "full" | "pathname" | "pathnameAndSearch";
+
 // @public
 export function removeGlobalContexts(contexts: Array<ConditionalContextProvider | ContextPrimitive>, trackers?: Array<string>): void;
 
@@ -417,6 +421,7 @@ export type TrackerConfiguration = {
     retryFailedRequests?: boolean;
     onRequestSuccess?: (data: EventBatch) => void;
     onRequestFailure?: (data: RequestFailure) => void;
+    preservePageViewIdForUrl?: PreservePageViewIdForUrl;
 };
 
 // @public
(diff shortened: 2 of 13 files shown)


davidher-mann:
That was quick Matus - thanks!

j7i:
Thank you @matus-tomlein - I will be able to look into it on Monday.

j7i:
This solution should solve our current problem, and I also appreciate the options provided to further specify an appropriate strategy such as `pathname`. Thank you!

However, we are currently aiming for a deferred mode where we delay the execution of tracking events (should go live this week). In this case, we would actually fire tracking events on a new page, even though they may originate from a previous page. We do this to shorten presentation delays on interactions.

With the current solution this would result in this scenario (when running in deferred mode; page view events are excluded from deferral):

1. User navigates to page http://example.com/a (generate page view ID 1)
2. Custom event A is scheduled
3. Custom event A is tracked with page view ID 1
4. Page view event is tracked with page view ID 1
5. Custom event B is scheduled
6. --- User clicks on a link that takes them to http://example.com/b (generate page view ID 2)
7. Custom event B is tracked with page view ID 2 (should be page view ID 1 👀)
8. Custom event C is scheduled
9. Page view event is tracked with page view ID 2
10. Custom event C is tracked with page view ID 2

We will consider to checkout this strategy in combination with performance improvements regarding the cookie strategy / expensive cookie reads/writes. As we may can get rid off the deferral mode as soon as tracking calls are less expensive.

j7i:
As an addition to my previous comment:
If there is some sort of override available in combination with a global context getter, this could work with a deferred approach. But I think this would be another topic to further discuss / explore.

```typescript
/** This code is for demonstration purposes only */

// should also return the web page entity among the other global ones
import { getGlobalContext } from "@snowplow/browser-tracker";

const globalContext = getGlobalContext(); // could receive a tracker id as an argument
const timestamp = Date.now();
const event = { ... };
const customContext = [ ... ]; 

requestIdleCallback(() => 
  trackSelfDescribingEvent({
    event,
    context,
    globalContext,
    timestamp,
  });
);

```

Jack-Keene (approved):
LGTM!

PR #1312, opened 2024-05-29
browser-plugin-media-tracking: adjust fileExtension extraction

jethron:
The browser-plugin-media-tracking plugin builds its own payload for the [media_element](https://github.com/snowplow/iglu-central/tree/master/schemas/org.whatwg/media_element/jsonschema/1-0-0) entity, including the `fileExtension` property.

The current logic optimistically splits the current source URI on `.` and takes the last resulting substring.

In cases where the source URI contains query parameters or other trailing data after the file extension, the resulting value for `fileExtension` can be overly long, going over the `maxLength: 255` the schema defines.

This change adjusts the behaviour to only extract values following a dot that also excludes URI metacharacters like `?` that would denote query parameters. It also now prefers parsing the URI path value explicitly over the other parts like query parameters.

### DIFF
--- a/common/changes/@snowplow/browser-plugin-media-tracking/PE-6330-file_ext_fix_2024-05-29-01-32.json
+++ b/common/changes/@snowplow/browser-plugin-media-tracking/PE-6330-file_ext_fix_2024-05-29-01-32.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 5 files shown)


greg-el (approved):
Fix looks good to me!

Could you change the target branch to `release/3.23.1`?

PR #1314, opened 2024-06-04
Release 3.23.1

greg-el:
Bug fixes:

- #1312 
- #1313


greg-el:
Ah yeah, would probably be better if it didn't do that. I've merged it in.

PR #1316, opened 2024-06-06
Export MediaEventType from media plugin (Close #1315)

### LINKED ISSUE #1315: Export `MediaEventType` from media plugin
**Describe the bug**

 The MediaEventType type is not exported from the plugin although it is used in the public API of the plugin. 

**Expected behavior**
We should export it.

Jack-Keene:


plugins/browser-plugin-media/src/api.ts
```diff
@@ -396,7 +396,9 @@ export function trackMediaAdSkip(
   trackers: Array<string> = Object.keys(_trackers)
 ) {
   let { percentProgress } = args;
-  if (percentProgress !== undefined) { percentProgress = Math.floor(percentProgress); }
```
greg-el, line 399: Can you remove these styling changes from the PR?

greg-el (approved):
LGTM

plugins/browser-plugin-media/src/api.ts
Jack-Keene: It does this on the Pre-commit CI now, so not really!

PR #1320, opened 2024-06-17
Handle Infinity and NaN durations in Html Media Tracking (close #1319)

### LINKED ISSUE #1319: Handle durations of `NaN` and `+Infinity` in HTML Media Plugin instead of defaulting to 0

greg-el:
This PR handles `NaN` and `Infinity` being returned from video/audio `duration`.

Previously this defaulted to 0 if a falsy value was returned. As the [the schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/media_player/jsonschema/1-0-0#L18) allows `null` for duration,  we now return `null` if duration is `NaN` or `Infinity` (which [are](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/duration) the only other possible return types).


matus-tomlein (approved):
LGTM!

PR #1324, opened 2024-07-04
Fix API docs CI action and update it's dependencies

matus-tomlein:
This PR updates the API docs action to fix two things:

1. Update it's dependencies to avoid vulnerabilities.
2. Fix the CI action which had trouble with one specific which was defined inline in the code. I extracted a new interface for it to define it explicitly.


Jack-Keene (approved):
LGTM!

PR #1325, opened 2024-07-05
link-click-tracking: event delegation

jethron:
There are several API changes included:

- Global event listeners are now used instead of per-element listeners
- `refreshLinkClickTracking` is now deprecated and does nothing as it should no longer be required (but still exists for API compatibility)
- There is now a `disableLinkClickTracking` method to balance `enableLinkClickTracking`
- `trackLinkClick` can now optionally be passed a link element directly, rather than requiring manual construction of the event payload

See PE-4138

### DIFF
--- a/plugins/browser-plugin-link-click-tracking/src/index.ts
+++ b/plugins/browser-plugin-link-click-tracking/src/index.ts
@@ -166,17 +166,17 @@ export function refreshLinkClickTracking(_trackers: Array<string> = []) {
 /**
  * Manually log a click.
  *
- * @param event The event information
+ * @param event The event or element information
  * @param trackers The tracker identifiers which the event will be sent to
  */
 export function trackLinkClick(
-  event: (LinkClickEvent | { element: TrackableElement }) & CommonEventProperties,
+  event: (LinkClickEvent | { element: TrackableElement; trackContent?: boolean }) & CommonEventProperties,
   trackers: Array<string> = Object.keys(_trackers)
 ) {
   dispatchToTrackersInCollection(trackers, _trackers, (t) => {
     let payload: PayloadBuilder | undefined;
     if ('element' in event) {
-      const includeContent = _configuration[t.id]?.linkTrackingContent ?? false;
+      const includeContent = event.trackContent ?? _configuration[t.id]?.linkTrackingContent ?? false;
       payload = processClick(event.element, includeContent);
     } else {
       payload = buildLinkClick(event);
--- a/plugins/browser-plugin-link-click-tracking/src/index.ts
+++ b/plugins/browser-plugin-link-click-tracking/src/index.ts
@@ -145,9 +145,9 @@ export function disableLinkClickTracking(trackers: Array<string> = Object.keys(_
   trackers.forEach((id) => {
     if (_trackers[id] && _listeners[id]) {
       // remove all possible cases where the handler may have been attached
-      window.removeEventListener('click', _listeners[id]);
-      window.removeEventListener('mouseup', _listeners[id]);
-      window.removeEventListener('mousedown', _listeners[id]);
+      window.removeEventListener('click', _listeners[id], true);
+      window.removeEventListener('mouseup', _listeners[id], true);
+      window.removeEventListener('mousedown', _listeners[id], true);
     }
   });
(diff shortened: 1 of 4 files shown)


plugins/browser-plugin-link-click-tracking/src/index.ts
```diff
@@ -83,178 +99,200 @@ export interface LinkClickTrackingConfiguration {
...
-    sourceElement = parentElement;
+  // Ignore script pseudo-protocol links
+  if (!ELEMENT_PROTOCOL_FILTER.test(sourceHref)) {
+    elementId = anchorElement.id;
+    elementClasses = getCssClasses(anchorElement);
+    elementTarget = anchorElement.target;
+    elementContent = includeContent ? anchorElement.innerHTML : undefined;
+
+    // decodeUrl %xx
+    // NOTE(jethron): Why do we do this?
```
matus-tomlein, line 211: I think it would be a good idea to drop the `unescape` or replace it with `decodeURI`. Based [on the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape), they don't recommend using it.

I assume what we want to do here is just to make the URI look nicer so that a URL like:
`https://snowplow.io/?x=a%20b%20c`
will turn into:
`https://snowplow.io/?x=a b c`

I am not convinced it's a good idea, because we don't seem to do it for the URL property in page views, which creates an inconsistency in how the URL is represented.

I think that as part of the breaking changes of the v4, we can also drop this. What do you think?

plugins/browser-plugin-link-click-tracking/test/events.test.ts
```diff
@@ -36,17 +36,15 @@ const getUEEvents = F.compose(F.filter(F.compose(F.eq('ue'), F.get('evt.e'))));
...
+    from: (a: any, n: number = 0) =>
+      F.nth(
+        n,
+        F.compose(F.filter(F.compose(F.eq(schema), F.get('schema'))), F.flatten, extractEventProperties, getUEEvents)(a)
+      ),
   };
 };
 
-describe('AdTrackingPlugin', () => {
+describe('LinkClickTrackingPlugin', () => {
```
matus-tomlein, line 47: The link click tracking plugin has one of the lowest unit test coverage in the tracker (around 18% last time we checked it). This is a bit increasing the scope of the PR, but would you consider adding a few unit tests to cover more of the event listening? Considering how flaky the integration tests are, we wanted to move more of the testing to the unit tests. Peter has done some unit tests in the cross domain plugin that use jest to mock links and also test link clicks ([see here](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/libraries/browser-tracker-core/test/tracker/cross_domain.test.ts#L53)). Maybe we could do something similar here if you have the capacity left?

matus-tomlein:
I have just rebased the `release/4.0.0` branch on top of master since it was quite a bit behind which resulted in a few conflicts here. Sorry about that, hope it won't be too much trouble to rebase!

plugins/browser-plugin-link-click-tracking/src/index.ts
jethron: Yeah I agree; it's a lossy transform and would be hard to parse after the fact for any non-human-legibility use case. I'll remove this.

plugins/browser-plugin-link-click-tracking/test/events.test.ts
matus-tomlein: Very nice tests, thanks! I think @igneel64 used [istanbul.js](https://istanbul.js.org) to calculate the code coverage, there is [a report here](https://www.notion.so/keep-in-the-snow/Trackers-Integration-Testing-overview-Web-focused-cae853ad6ff2452da7275479ec99976f).

matus-tomlein (approved):
LGTM! Very nice work and seems to be working well!

matus-tomlein:
A sidenote: when you merge the PR, can you please use "squash and merge" or rename the commit to follow the style that our other commits do: https://github.com/snowplow/snowplow-javascript-tracker/commit/2d5b8afc873ee5dcf5b02837a96fd0066e48254e

PR #1328, opened 2024-07-16
Truncate stack when application_error stack trace exceeds 8kb (close #1327)

### LINKED ISSUE #1327: Application_error stack trace exceeds 8kb
**Describe the bug**
Sometimes the stack trace in the trackError event exceeds 8kb, causing a failed event.
 
**Expected behavior**
The error should not cause a failed event - the stack trace should be truncated.

Jack-Keene:
This PR fixes an issue where the stack trace exceeds 8k - causing a failed event on the error tracking. 

It uses the 'Blob' API to check the size of the stack trace, and if it exceeds 8kb truncates it using the `TextDecoder` API.

### DIFF
--- a/plugins/browser-plugin-error-tracking/src/index.ts
+++ b/plugins/browser-plugin-error-tracking/src/index.ts
@@ -36,6 +36,7 @@ import {
   dispatchToTrackersInCollection,
 } from '@snowplow/browser-tracker-core';
 import { buildSelfDescribingEvent, CommonEventProperties, SelfDescribingJson } from '@snowplow/tracker-core';
+import { truncateStackTrace } from './util';
 
 let _trackers: Record<string, BrowserTracker> = {};
 
@@ -74,7 +75,7 @@ export function trackError(
   trackers: Array<string> = Object.keys(_trackers)
 ) {
   const { message, filename, lineno, colno, error, context, timestamp } = event,
-    stack = error && error.stack ? error.stack : null;
+    stack = error && truncateStackTrace(error.stack);
 
   dispatchToTrackersInCollection(trackers, _trackers, (t) => {
     t.core.track(
--- a/plugins/browser-plugin-error-tracking/src/index.ts
+++ b/plugins/browser-plugin-error-tracking/src/index.ts
@@ -36,7 +36,7 @@ import {
   dispatchToTrackersInCollection,
 } from '@snowplow/browser-tracker-core';
 import { buildSelfDescribingEvent, CommonEventProperties, SelfDescribingJson } from '@snowplow/tracker-core';
-import { truncateStackTrace } from './util';
+import { truncateString } from './util';
 
 let _trackers: Record<string, BrowserTracker> = {};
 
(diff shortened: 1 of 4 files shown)


plugins/browser-plugin-error-tracking/src/util.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Truncate stack trace if it's size is greater than 8kb
+ *
+ * @param stackTrace - The stack trace to truncate
+ * @returns The trucated stack trace
+ */
+export function truncateStackTrace(stackTrace?: string): string | undefined {
+  const byteSize = (str: string) => new Blob([str]).size;
```
matus-tomlein, line 8: I think this is complicating it too much 😅 The `stackTrace` property in the schema is a string with max length 8192 characters. Converting it to bytes and setting the limit in bytes is a bit unintuitive.

You could just take the prefix of the string – take the first 8192 characters if the string has more than that. See how we do it in [the mobile trackers](https://github.com/snowplow/snowplow-ios-tracker/blob/master/Sources/Snowplow/Events/TrackerError.swift#L67). Btw, there are also limits on the message and exception name that would be good to check as well.

plugins/browser-plugin-error-tracking/src/index.ts
```diff
@@ -74,7 +75,7 @@ export function trackError(
   trackers: Array<string> = Object.keys(_trackers)
 ) {
   const { message, filename, lineno, colno, error, context, timestamp } = event,
-    stack = error && error.stack ? error.stack : null;
+    stack = error && truncateStackTrace(error.stack);
```
matus-tomlein, line 78: Can you please add a unit test to make sure we don't regress on the behaviour in the future?

plugins/browser-plugin-error-tracking/src/util.ts
Jack-Keene: Ah ok, I thought it was actually limited in bytes, so this was taking into account multibyte characters!

matus-tomlein (approved):
LGTM! 👍

plugins/browser-plugin-error-tracking/src/util.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Truncate string if it's longer than 8192 chars
+ *
+ * @param string - The stack trace to truncate
+ * @param maxLength - The maximum length of the truncated string. Default = 8192
+ * @returns The truncated string
+ */
+export function truncateString(string?: string, maxLength: number = 8192): string | undefined {
+  if (string && string.length > 8192) {
```
jethron, line 9: Should this be testing against `maxLength`?

plugins/browser-plugin-error-tracking/test/events.test.ts
```diff
@@ -126,4 +142,16 @@ describe('AdTrackingPlugin', () => {
...
     });
   });
+
+  it('trackError replaces undefined messages with placeholder', () => {
+    expect(
+      extractUeEvent('iglu:com.snowplowanalytics.snowplow/application_error/jsonschema/1-0-1').from(state.outQueues[3])
+    ).toMatchObject({
+      schema: 'iglu:com.snowplowanalytics.snowplow/application_error/jsonschema/1-0-1',
+      data: {
+        message: 'y'.repeat(2048),
```
jethron, line 152: I'm kind of surprised this doesn't fail with the above implementation?

Jack-Keene: i think that was just by chance as the test string was greater than 8192, and was then truncated to the correct length!

PR #1329, opened 2024-07-17
Add event delegation for form tracking

jethron:
- Global event listeners are now used instead of per-element listeners
- Event listeners now use the capture phase instead of bubble
- Forms added after the initial `enableFormTracking` call should automatically be tracked without needing additional calls
- Field value transform functions can now return `null` for all events
- Add unit tests for form tracking
- The new `targets` parameter allows listening for forms in only a subset of the document tree, or across multiple same-origin documents (e.g. iframes)

PE-5793 PE-6528 #to-review

### DIFF
--- a/plugins/browser-plugin-form-tracking/src/helpers.ts
+++ b/plugins/browser-plugin-form-tracking/src/helpers.ts
@@ -1,16 +1,16 @@
 import {
-  getCssClasses,
   addEventListener,
-  BrowserTracker,
-  FilterCriterion,
+  getCssClasses,
   getFilterByClass,
   getFilterByName,
+  type BrowserTracker,
+  type FilterCriterion,
 } from '@snowplow/browser-tracker-core';
 import {
-  resolveDynamicContext,
-  DynamicContext,
   buildFormFocusOrChange,
   buildFormSubmission,
+  resolveDynamicContext,
+  type DynamicContext,
 } from '@snowplow/tracker-core';
 
 /** The form tracking configuration */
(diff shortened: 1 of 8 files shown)


plugins/browser-plugin-form-tracking/src/helpers.ts
```diff
@@ -78,237 +78,351 @@ interface FormConfiguration {
...
+ * @param forms Explicitly configured list of form elements to track, if any
+ * @returns List of EventTargets to add the listener to
+ */
+function getTargetList(configTargets: EventTarget[] | undefined, forms: FormConfiguration['forms']) {
+  // we attach to document rather than window because the window focus event occurs more often than we require
+  const targets = configTargets ?? [document];
+
+  if (forms) {
+    Array.prototype.forEach.call(forms, (form: HTMLFormElement) => {
+      targets.push(form.ownerDocument.documentElement);
```
matus-tomlein, line 133: I am probably missing something, but is it possible that if two forms were passed here that come from the same document, we will subscribe to the form events from that document twice (will track duplicate form event)?

jethron: We'll call addEventListener multiple times, yes, but because we use an identical listener function instance they will naturally dedupe and only fire once.

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

> If the function or object is already in the list of event listeners for this target, the function or object is not added a second time.

Though we can probably do a quick uniqueness pass first if you think it'd be safer?

matus-tomlein: Ahh, wasn't aware that it dedupes by the listener function! That sounds good!

matus-tomlein (approved):
Very nice implementation and appreciate the added unit tests and disable function! 👍

Issue #1304, opened 2024-04-02
Properly allow for type SelfDescribingJSON to be added on `addGlobalContexts`


igneel64:
**Describe the bug**
Currently the `addGlobalContexts` function does not allow for typed `SelfDescribingJSON` schemas to be added.

**To Reproduce**

https://github.com/snowplow/snowplow-javascript-tracker/assets/15251081/4b4ed24f-5bdd-4184-8871-1f658b17453c

**Expected behavior**
The `addGlobalContexts` function allows for proper `SelfDescribingJSON` objects to be passed to it.

PR #1330, opened 2024-07-17
Fix SelfDescribingJson type to allow optional keys in type parameter (close #1203 and #1304)

### LINKED ISSUE #1203: Improve `SelfDescribingJson` type
The `SelfDescribingJson` type in [core](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/libraries/tracker-core/src/core.ts#L48) takes a type parameter, but will not accept a type with optional properties:

```typescript
type Foo = {
  bar: string;
  baz?: string;
};

const sdj: SelfDescribingJson<Foo> = {
  schema: 'iglu:com.snowplow/foo/jsonschema/1-0-0',
  data: {
    bar: 'bar',
    // note baz not provided
  },
};
```

will fail with:

```
Type 'Foo' does not satisfy the constraint 'Record<keyof Foo, unknown>'.   Property 'baz' is optional in type 'Foo' but required in type 'Record<keyof Foo, unknown>'.ts(2344)
```

Schemas with optional properties are very common, and providing a type to accomodate this would improve type safety wherever `SelfDescribingJson` is currently used.

jethron:
`Record<keyof T,` has the unfortunate side-effect of erasing the `optional` modifier for any properties in `T`, essentially mandating `Required<T>` instead of just `T`.

Explicitly using an index type removes this erasure while still being largely compatible with the rest of the interface expressed by `Record`.
An index type accepting only strings is preferable over plain `{}` because that allows arrays for `data`, which we don't support AFAIK outside of special cases like `payload_data` and `contexts`.

Should fix #1203 and #1304.

### DIFF
--- a/api-docs/docs/browser-tracker/browser-tracker.api.md
+++ b/api-docs/docs/browser-tracker/browser-tracker.api.md
@@ -143,7 +143,9 @@ export interface ClientSession extends Record<string, unknown> {
 }
 
 // @public
-export interface CommonEventProperties<T = Record<string, unknown>> {
+export interface CommonEventProperties<T extends {
+    [_: string]: unknown;
+} = Record<string, unknown>> {
     context?: Array<SelfDescribingJson<T>> | null;
     // Warning: (ae-forgotten-export) The symbol "Timestamp" needs to be exported by the entry point index.module.d.ts
     timestamp?: Timestamp | null;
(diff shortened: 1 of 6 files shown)


matus-tomlein (approved):
Would like to get feedback from @igneel64 on this but LGTM!

matus-tomlein:
Sorry @jethron, you might need to rebase this branch as I rebased the release branch on top of master.

matus-tomlein:
@jethron can you please merge this into the release branch? I think we can move forward with this.

PR #1331, opened 2024-07-17
Add support for named global context

jethron:
This is a proposal for "named global context"; an update to the `addGlobalContexts`/`removeGlobalContexts` APIs to allow naming each primitive/generator.

Once named, you can remove them by referencing just the name, but you can also upsert the value by adding via the same name again. This allows avoiding gymnastics required by the current API, where to update a primitive you have to have the original value around in order to remove it in order to add it again without creating duplicates.

### DIFF
--- a/api-docs/docs/browser-tracker/browser-tracker.api.md
+++ b/api-docs/docs/browser-tracker/browser-tracker.api.md
@@ -29,7 +29,7 @@ export interface ActivityTrackingConfigurationCallback {
 }
 
 // @public
-export function addGlobalContexts(contexts: Array<ConditionalContextProvider | ContextPrimitive>, trackers?: Array<string>): void;
+export function addGlobalContexts(contexts: Array<ConditionalContextProvider | ContextPrimitive> | Record<string, ConditionalContextProvider | ContextPrimitive>, trackers?: Array<string>): void;
 
 // @public
 export function addPlugin(configuration: BrowserPluginConfiguration, trackers?: Array<string>): void;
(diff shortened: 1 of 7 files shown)


matus-tomlein (commented):
I really like the approach and that it is done in a non-breaking way!

I am a bit unsure whether we should keep providing so many ways to configure global context or rather should deprecate the non-named approach and just guide users to name their global context. It makes it more confusing for users if the API supports so many variations. On the other hand, global context is a commonly used feature and making the change non-breaking should speed up the adoption of the v4, so this solution is very tempting.

I must be doing something stupid, but for some reason I don't see the global context entities configured using the named approach in events tracked from a demo React app. I do see the ones configured with the unnamed approach though. This is the way I am adding the context:

```
  // this one doesn't show up in events
  addGlobalContexts({
    c1: {
      schema: "iglu:com.acme_company/page/jsonschema/1-0-0",
      data: {
        title: "Page 1",
      },
    },
  });

  // this one gets tracked in events
  addGlobalContexts([
    {
      schema: "iglu:com.acme_company/page/jsonschema/1-0-0",
      data: {
        title: "Page 2",
      },
    },
  ]);
```

Do you see anything obvious that I am doing wrong here? I'll give it another shot tomorrow.

jethron:
Nope, that looks fine to me. I can't replicate this issue, it works as expected for both JS tracker and Browser tracker when I build it. :confused: Did you do a full rebuild to get the changes in tracker-core etc?

matus-tomlein (approved):
Sorry, this was my confusion, for some reason I thought that global context was in the `browser-tracker-core`, not in the `tracker-core`.

It all works very nicely and is a solid improvement! I am still a bit unsure about having multiple options to do the same thing in the API, but I think we can make the trade-off towards making it easier to upgrade to v4 for users and do a better job in the docs to make sure that people use the named approach.

Issue #1332, opened 2024-07-17
Tracking a page view with a custom title sets the title for future page views as well


matus-tomlein:
**Describe the bug**
This bug was introduced in [this commit](https://github.com/snowplow/snowplow-javascript-tracker/commit/9b7e4f6cc7c5089b706413cf79d4846e5e7bb0e4) that tried to fix the issue that the custom page title was not considered when tracking page views. However, the result of the change is that when one tracks a page view with a custom title, and another page view without title, the second page view will get the title of the first page view.

**To Reproduce**
Track two page views after each other, the first one with a title, the second without.

**Expected behavior**
The second page view should not get the title of the first page view.

PR #1333, opened 2024-07-18
Fix integration tests

matus-tomlein:
Makes the tests less flaky by:

1. Updates the webdriver dependencies
2. In `integration.test.ts` accounts for the bug #1332 which causes the second page view to have the same custom title as the first one if not explicitly specified. Currently just comments the check out, but will fix the bug in a follow up PR.
3. In `sessionStorage.test.ts` makes the page reload between checks of the expired session cookie as Chrome is flaky about purging the cookie if the page does not reload.
4. In `cookies.test.ts` skips a check for a cookie being dropped as this is flaky on Chrome.
5. In `media.test.ts` removes checks for the `duration` as not all events have this property if the video is not fully loaded yet.
6. Moves the media tests to a separate folder and disables the Vimeo integration tests on CI (still work locallly) as Vimeo blocks Sauce Labs


trackers/javascript-tracker/test/functional/cookies.test.ts
```diff
@@ -2,23 +2,24 @@ describe('Tracker created domain cookies', () => {
   it('contain the expected cookie names', async () => {
     await browser.url('/cookies.html');
     await browser.waitUntil(async () => (await $('#init').getText()) === 'true', {
-      timeout: 5000,
+      timeout: 10000,
       timeoutMsg: 'expected init after 5s',
-      interval: 250,
+      interval: 5000,
```
jethron, line 7: Not sure why we need to increase the interval this much?

jethron (approved):
LGTM!

trackers/javascript-tracker/test/functional/cookies.test.ts
matus-tomlein: Good catch, that was a relict of my experiments, have now removed.

Issue #1317, opened 2024-06-13
setDocumentTitle does not override the title for trackPageView event


marcin-j:
**Describe the bug**
setDocumentTitle does not override the title for trackPageView event

**To Reproduce**
1. Create tracker
2. Override page title with some custom title`setDocumentTitle("foo bar");`
3. Track page view `trackPageView();`

**Expected behavior**
Page view event should be sent with "foo bar" title, but instead, it's sent with regular document.title.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOS Sonoma
 - Browser Chrome
 - Version 125.0.6422.144 (Official Build) (arm64)

PR #1334, opened 2024-07-19
Fix tracking a page view with a custom title sets the title for future page views as well (close #1332)

### LINKED ISSUE #1332: Tracking a page view with a custom title sets the title for future page views as well

matus-tomlein:
#1332

Fixes an issue introduced in 3.24.0 where setting a title in the `trackPageView` call resulted in the title being the same also in the following page views unless overriden again.

The change now enforces the title to be reset on the next page view event.

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/issue-1332-page_title_fix_2024-07-19-13-02.json
+++ b/common/changes/@snowplow/browser-tracker-core/issue-1332-page_title_fix_2024-07-19-13-02.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
+      "comment": "Fix tracking a page view with a custom title sets the title for future page views as well (#1332)",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/browser-tracker-core"
+}
--- a/libraries/browser-tracker-core/src/tracker/index.ts
+++ b/libraries/browser-tracker-core/src/tracker/index.ts
@@ -210,6 +210,10 @@ export function Tracker(
       lastDocumentTitle = document.title,
       // Custom title
       lastConfigTitle: string | null | undefined,
+      // Indicates that the lastConfigTitle was set from a trackPageView call
+      // Custom title configured this way has a shorter lifespan than when set using setDocumentTitle.
+      // It only lasts until the next trackPageView call.
+      lastConfigTitleFromTrackPageView: boolean = false,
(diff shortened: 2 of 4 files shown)


jethron (commented):
I think this is actually the intended behaviour (as weird as it is); the same occurs for the associated `setCustomUrl` and `setReferrerUrl` methods, so fixing _just title_ might actually be more confusing than leaving it as-is? I think the intent is that if you use one of these methods once, you're basically taking ownership of the values and doomed to call it every time for the life of the tracker.

It's probably a good breaking change for v4 rather than a fix for v3; `trackPageView` should take overrides for all of `title`, `url`, and `referrer` (more like `buildPageView`), and only persist it until the next call to `trackPageView` -- and we just remove these methods entirely IMO. (Though there's the case where you want to force these values to track an event without a pageview first, I guess... Maybe these belong on `CommonEventProperties`, but that's a slippery slope. Filters could help here I guess? Track a pageview, filter it out, then track the event you want with the custom values?)

matus-tomlein:
Thanks @jethron, but I'm not sure I'm following:

1. I didn't change the behaviour of `setDocumentTitle` – that still persists across an unlimited number of page views (until it is called again). So don't understand how is it different from `setCustomUrl` or `setReferrerUrl`? Unless I am missing something, only the page title can be overriden in the `logPageView` method, so only in that one we need to handle the case where we can override the title in two ways (from the `setDocumentTitle` and `trackPageView`).
2. It's questionable if this is a breaking change because this was the behaviour before 3.24.0 but we broke it when fixing issue #1317. Also our integration tests expect this behaviour (we just didn't notice they were failing as everything was failing...).
3. It's a good idea to add overrides for the `url` and `referrer` to the `trackPageView` in v4. Not sure if we can totally drop the global setter functions, but agree that it's better to pass the values when tracking page view rather than set them globally.

jethron (approved):
My bad @matus-tomlein. I totally misread this! :sweat_smile: I'd thought it was using `setDocumentTitle` and that was what we're reversing; totally opposite.

LGTM!

Issue #1306, opened 2024-04-16
ResiezObserver loop completed with undelivered notifications


sykes10:
**Describe the bug**
After upgrading @snowplow/browser-tracker from version 3.22.1 to version 3.23.0 we have been getting thousands of Bugsnag with the message `ResiezObserver loop completed with undelivered notifications`.

This message is reported for all browsers and OS using our website.

We don't have any instances of ResizeObserver in our codebase that might be causing this.

**To Reproduce**
Run the latest version of @snowplow/browser-tracker and capture the errors with a tool like Sentry or Bugsnag. We don't know how to trigger it, but Bugsnag reports this error for each session.

**Expected behaviour**
No errors should be reported. 

**Observations**
Checking the codebase I see the ResizeObserver being registered but not unregistered at https://github.com/snowplow/snowplow-javascript-tracker/blob/1a073e4ef57cce3d5307ec2ed7a9f1e3a022ff9f/libraries/browser-tracker-core/src/helpers/browser_props.ts#L34.

matus-tomlein:
Hi @sykes10, thank you for reporting this problem!

I have tried to reproduce it using [a demo React app](https://github.com/snowplow-incubator/snowplow-javascript-tracker-examples/tree/master/react) with the JS tracker version 3.23.0 and installed BugSnag, but wasn't able to produce the errors that you are getting.

Could you please help us investigate this deeper? Searching for the problem, I see that it is often happening due to users having some browser extensions installed. Do you see the problem happening for all users or certain browsers/setups? Are you able to reproduce this locally?

sykes10:
Hi @matus-tomlein, Thanks for your quick reply, first I can confirm that downgrading the package to 3.22.1 stops all the Bugsnags.

To give you more context our application is a Nuxt app version 3.8.2 with vue 3.3.13.

I also checked the demo you linked and I see that the Snowplow version you have in the package.json is 3.14.0 but the error is introduced in 3.23.0. Could you double-check that it was tested with the right version?

Also here's a list of some browsers that are experiencing this.
![image](https://github.com/snowplow/snowplow-javascript-tracker/assets/11991645/e379422a-fb06-42a9-94a1-b6dd7ffbdd29)

miike:
@sykes10 Just following up on this one - did you manage to find a way to recreate this?

sykes10:
I've been busy these last days but I will try to create a reproducible environment this week.

AngusMorton:
[...]

PR #1335, opened 2024-07-23
Avoid undelivered notifications error

AngusMorton:
Maybe closes: https://github.com/snowplow/snowplow-javascript-tracker/issues/1306

I have yet to validate that this change fixes anything. I plan to test this and see if it reduces the error rate. I will comment once I've tested this in production and confirmed that it resolves the errors.

Based on the [ResizeObserver documentation](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver), I suspect that calling `readBrowserProperties()` forces a synchronous layout, which then enqueues a new resize observer notification if the body size changes when the layout is re-calculated.


snowplowcla:

libraries/browser-tracker-core/src/helpers/browser_props.ts
```diff
@@ -18,26 +18,44 @@ function useResizeObserver(): boolean {
...
+        }
+      });
     }
   });
   resizeObserver.observe(document.body);
   resizeObserver.observe(document.documentElement);
 }
 
+function scheduleBrowserPropertiesCallback(func: () => void) {
+  if (requestAnimationFrame) {
```
jethron, line 52: This needs to check with `typeof` or against `window` or it will still throw `is not defined` (though I'm not sure we offically still support anything that doesn't support rAF anyway).

jethron (commented):
LGTM after comments if LGT @matus-tomlein. Look forward to the results of your testing!

We should probably make `getBrowserProperties` only do the expensive stuff conditionally as well; 1/3 callers don't even use the expensive `viewport`/`documentSize` values and it's when the tracker first loads so probably a bad time to be forcing reflow.

libraries/browser-tracker-core/src/helpers/browser_props.ts
AngusMorton: > I'm not sure we offically still support anything that doesn't support rAF anyway

Assuming the [browserslistrc](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/.browserslistrc) is up to date, it looks like IE 9 is in the supported list of browsers, which doesn't support rAF.

AngusMorton:
> We should probably make getBrowserProperties only do the expensive stuff conditionally as well; 1/3 callers don't even use the expensive viewport/documentSize values and it's when the tracker first loads so probably a bad time to be forcing reflow.

I'm happy to make that change if you're open to it, I can do it in a separate PR.

AngusMorton:
I signed it!

snowplowcla:
Confirmed! @AngusMorton has signed the Contributor License Agreement. Thanks so much.

libraries/browser-tracker-core/src/helpers/browser_props.ts
```diff
@@ -18,20 +18,26 @@ function useResizeObserver(): boolean {
...
-  const resizeObserver = new ResizeObserver((entries) => {
-    for (let entry of entries) {
-      if (entry.target === document.body || entry.target === document.documentElement) {
+  const resizeObserver = new ResizeObserver(() => {
+    if (!readBrowserPropertiesTask) {
+      // The browser property lookup causes a forced synchronous layout when offsets/sizes are
+      // queried. It's possible that the forced synchronous layout causes the ResizeObserver
+      // to be fired again, leading to an infinite loop and the "ResizeObserver loop completed
+      // with undelivered notifications" error.
+      readBrowserPropertiesTask = requestAnimationFrame(() => {
```
matus-tomlein, line 37: Unfortunately the `requestAnimationFrame` API is not supported on all browsers that the JS tracker v3 supports. Could we check if the API is available in the browser and avoid using it if not?

matus-tomlein (commented):
Except for the browser compatibility, this looks good to me too!

But would be good to get some validation if this actually helps. I have unfortunately been unable to reproduce the issue thus far, so can't test it out. If you have a way to validate it, please do and let us know!

libraries/browser-tracker-core/src/helpers/browser_props.ts
```diff
@@ -18,20 +18,26 @@ function useResizeObserver(): boolean {
...
-  if(!document || !document.body || !document.documentElement) {
+  if (!document || !document.body || !document.documentElement) {
     return;
   }
   resizeObserverInitialized = true;
 
-  const resizeObserver = new ResizeObserver((entries) => {
-    for (let entry of entries) {
-      if (entry.target === document.body || entry.target === document.documentElement) {
+  const resizeObserver = new ResizeObserver(() => {
```
devbrains-com, line 31: To add a fallback for oder browsers I suggest doing this:

```
    var requestAnimationFrameIsSupported = !!window.requestAnimationFrame;
    var resizeObserver = new ResizeObserver(() => {
        if(!requestAnimationFrameIsSupported) {
            cachedProperties = readBrowserProperties();
        }
        else if (!readBrowserPropertiesTask) {
            readBrowserPropertiesTask = requestAnimationFrame(() => {
            readBrowserPropertiesTask = null;
            cachedProperties = readBrowserProperties();
        });
    }
    });
```

devbrains-com: Or Alternative:

```
    const requestAnimationFrameIsSupported = !!window.requestAnimationFrame;
    const resizeObserver = new ResizeObserver(() => {
        if(!requestAnimationFrameIsSupported) {
            setTimeout(() => {
                cachedProperties = readBrowserProperties();
            }, 0);
        }
        else if (!readBrowserPropertiesTask) {
            readBrowserPropertiesTask = requestAnimationFrame(() => {
            readBrowserPropertiesTask = null;
            cachedProperties = readBrowserProperties();
        });
    }
    });
```

matus-tomlein (approved):
Thanks again for the contribution! We'll work on the browser compatibility change separately.

PR #1336, opened 2024-07-23
Release/3.24.2

### LINKED ISSUE #1327: Application_error stack trace exceeds 8kb

### LINKED ISSUE #1332: Tracking a page view with a custom title sets the title for future page views as well

matus-tomlein:
This release fixes a bug that caused the page title passed in `trackPageView` to be applied to the following page views (unless a new title was specified). It also prevents failed events with the `application_error` schema by truncating the stack trace not to exceed the limit in the schema.

**Bug fixes**
* Fix tracking a page view with a custom title sets the title for future page views as well (close #1332)
* Truncate stack when `application_error` stack trace exceeds 8kb (close #1327)

**Under the hood**
* Fix integration tests (#1333)
* Fix API docs CI action and update it's dependencies (#1324)


Jack-Keene (approved):
LGTM!

PR #1340, opened 2024-09-03
Make cookie writes async by default to improve tracker performance

matus-tomlein:
This PR makes cookie writes async by default on the browser tracker in order to improve performance of the tracker call (to block the app for a shorter period of time when an event is tracked).

It adds a wrapper around the calls to read and write cookies that caches the last written value and uses `setTimeout` with 10ms to schedule the cookie write (it also debounces if there are multiple cookie writes in short sequences).

There is also a new option to keep the previous behaviour and write cookies synchronously with the track calls – the `synchronousCookieWrite` option. This may be useful in tests and it also has the benefit of checking that the cookies were correctly set.

I added an e2e test that compares the performance of the different storage strategies and the sync option, results below.

image

Numbers 2024-09-03 13 53 01

### DIFF
--- a/libraries/browser-tracker-core/src/tracker/cookie_storage.ts
+++ b/libraries/browser-tracker-core/src/tracker/cookie_storage.ts
@@ -70,7 +70,8 @@ function newCookie(name: string): Cookie {
   let lastSetValueArgs: Parameters<typeof setValue> | undefined;
   let cacheExpireAt: Date | undefined;
   let flushed = true;
-  const flushTimeout = 10;
+  const flushTimeout = 10; // milliseconds
+  const maxCacheTtl = 0.1; // seconds
 
   function getValue(): string {
     // Note: we can't cache the cookie value as we don't know the expiration date
@@ -92,9 +93,7 @@ function newCookie(name: string): Cookie {
       }, flushTimeout);
     }
 
-    if (ttl !== undefined) {
-      cacheExpireAt = new Date(Date.now() + ttl * 1000);
-    }
+    cacheExpireAt = new Date(Date.now() + Math.min(maxCacheTtl, ttl ?? maxCacheTtl) * 1000);
     return true;
   }
 
--- a/libraries/browser-tracker-core/src/tracker/cookie_storage.ts
+++ b/libraries/browser-tracker-core/src/tracker/cookie_storage.ts
@@ -67,35 +67,33 @@ interface Cookie {
 
 function newCookie(name: string): Cookie {
   let flushTimer: ReturnType<typeof setTimeout> | undefined;
-  let expireCookieTimer: ReturnType<typeof setTimeout> | undefined;
   let lastSetValueArgs: Parameters<typeof setValue> | undefined;
+  let cacheExpireAt: Date | undefined;
+  let flushed = true;
   const flushTimeout = 10;
 
   function getValue(): string {
     // Note: we can't cache the cookie value as we don't know the expiration date
-    return lastSetValueArgs?.[0] ?? cookie(name);
+    if (lastSetValueArgs && (!cacheExpireAt || cacheExpireAt > new Date())) {
+      return lastSetValueArgs[0] ?? cookie(name);
+    }
+    return cookie(name);
   }
 
   function setValue(value?: string, ttl?: number, path?: string, domain?: string, samesite?: string, secure?: boolean): boolean {
     lastSetValueArgs = [value, ttl, path, domain, samesite, secure];
+    flushed = false;
 
     // throttle setting the cookie
     if (flushTimer === undefined) {
       flushTimer = setTimeout(() => {
         flushTimer = undefined;
-        flush()
+        flush();
       }, flushTimeout);
     }
 
-    // set expiration timer
-    if (expireCookieTimer !== undefined) {
-      clearTimeout(expireCookieTimer);
-      expireCookieTimer = undefined;
-    }
     if (ttl !== undefined) {
-      expireCookieTimer = setTimeout(() => {
-        lastSetValueArgs = undefined;
-      }, ttl * 1000);
+      cacheExpireAt = new Date(Date.now() + ttl * 1000);
     }
     return true;
   }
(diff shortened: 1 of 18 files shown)


libraries/browser-tracker-core/src/tracker/cookie_storage.ts
```diff
@@ -0,0 +1,174 @@
...
+      clearTimeout(setCookieTimer);
+    }
+    setCookieTimer = setTimeout(() => {
+      setCookieTimer = undefined;
+      cookie(name, value, ttl, path, domain, samesite, secure);
+
+      if (ttl !== undefined) {
+        scheduleClearCache(ttl);
+      }
+    }, debounceTimeout);
```
igneel64, line 90: I am just thinking... Why aren't we using the requestAnimationFrame and setTimeout trick here?
```
requestAnimationFrame(() => {
    setTimeout(() => {
       cookie()...
    }, 0);
  });
```
Since the time is random it might again hit a blocking point when rendering would be critical and INP metrics might still be an issue. `requestAnimationFrame` is a bit safer on that aspect.

libraries/browser-tracker-core/src/tracker/cookie_storage.ts
```diff
@@ -0,0 +1,174 @@
...
+export const asyncCookieStorage = newCookieStorage();
+
+/**
+ * Cookie storage instance with synchronous cookie writes
+ */
+export const syncCookieStorage: CookieStorage = {
+    getCookie: cookie,
+    setCookie: (name, value, ttl, path, domain, samesite, secure) => {
+      cookie(name, value, ttl, path, domain, samesite, secure);
+      return document.cookie.indexOf(`${name}=`) !== -1 ? true : false;
```
jethron, line 170: Nit: `<bool expr> ? true : false` is just `<bool expr>`.

libraries/browser-tracker-core/src/tracker/cookie_storage.ts
jethron: I think the race risk is already pretty high with this change, adding indeterminate amounts of time before updating the value seems dangerous to values like `eventIndex`.

libraries/browser-tracker-core/src/tracker/cookie_storage.ts
```diff
@@ -0,0 +1,174 @@
...
+    // Note: we can't cache the cookie value as we don't know the expiration date
+    return cachedValue ?? cookie(name);
+  }
+
+  function setValue(value?: string, ttl?: number, path?: string, domain?: string, samesite?: string, secure?: boolean): boolean {
+    cachedValue = value;
+
+    // debounce setting the cookie
+    if (setCookieTimer !== undefined) {
+      clearTimeout(setCookieTimer);
```
jethron, line 81: Don't we want to throttle rather than debounce, to update to the latest value as quickly as possible for other pages/tabs that will need to read it?

jethron (commented):
Does this work ok with things like link/form tracking that navigate away from the page? I feel like a whole new task to update the cookie might break some of the data consistency we have in place in those cases, particularly stuff in `client_session`.

matus-tomlein:
Thanks both for the reviews!

@jethron I have added a `flushPendingCookies()` function to the tracker that forces all cookies to be written synchronously. I have added a call for this function in the link click tracking plugin and in the form tracking for form submit events. It's not a great solution (don't really like to add complexity and configuration options), but I think it can help if we identify the calls where cookies need to be written synchronously and force flush the cookies there.

It's a trade-off between performance and correctness that we have to make. In my opinion, the performance gains make it worth it in this case. Also I don't see another way to speed up the cookie writes without making them async, but happy to make changes if you have some other idea that would be safer!

jethron:
Sorry for the delay on this @matus-tomlein!

I did some testing and did find a race, but then when I compared it to v3 the behaviour there had a similar problem anyway, so I guess it doesn't really count as a regression -- it's just a long-standing bug. So I don't think that should block this.

The only other difference I really noted was that event "bursts" no longer automatically batch; the performance improvement kind of means every event gets sent individually now where in some cases in v3 because of the contention on the cookies those events would all be bundled together in a single payload, even if your batch size was 1 (the default). This isn't a big deal IMO, but good to note that the behaviour will be different now.

e.g. in below test v4 generates 20-30 requests but v3 generates more like 10-15 for similar numbers of events.

---

To test I had the same page open in two on-screen browser windows (important, just different tabs would throttle intervals) and ran the following snippet at roughly the same time (it tries to give some leeway with an initial timeout) in each via browser console:

```javascript
let start = Date.now();
setTimeout(function(){
  let i = setInterval(function(){
    if ((Date.now() - start) > 3000) clearInterval(i);
    spActivity('trackPageView')}, 100)
}, (+new Date()) % 1000);
```
(`spActivity` because I was using `trackers/javascript-tracker/test/pages/activity-callback.html` - I had to modify it to enable the `session` context  - but any page should work as well with normal `snowplow`).

Most events are OK, but when there's a race like this enforces, eventually you get collisions, where between the two pages events with different event IDs/payloads will generate with the same `eventIndex` value (which is extracted from the cookie) and this then means all the following events are technically wrong as well.

v3: `eventIndex` 933 generated twice
image
image

v4: `eventIndex` 975 generated twice
image
image

jethron (approved):
Races in generated data, but similar problems already exist in v3 so not a dealbreak.

Slight changes to batch behaviour but acceptable for the performance improvement IMO.

Otherwise LGTM, guess we'll find out if there's anything else we missed. :)

jethron:
FML, that was testing `release/4.0.0`, not your actual branch. This branch does actually seem a bit worse, the above seems to generate duplicate events way more often, even without much of a race involved. :/

matus-tomlein:
Thanks @jethron for testing that! That's a good test case since it's quite possible that users will open the same website in multiple windows.

I think the problem is not so much in the async cookie writes but in the fact that we now cache the cookie values. Once a cookie value is cached, we don't check for the actual cookie value which may be modified by another window.

I have now made a change to add a maximum lifetime of the cached cookie value to 100 ms. I have run the same test as you have and while I still get conflicts on the `eventIndex`, the events now can recover after a conflict which was not the case before – before once the caches got out of sync, they would never get back in sync.

We can lower the max cache lifetime further, hard to say what should be the setting... Happy to go with a different value if you prefer.

PR #1341, opened 2024-09-06
Fix message when a resource triggers errors

jethron:
The `error` event fires on the `window` both when an unhandled JavaScript exception is thrown _or_ a resource fails to load. Rather than track a message presuming the former, attempt to detect the latter and specify the failed resource.

See also:
- https://developer.mozilla.org/en-US/docs/Web/API/Window/error_event
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/error_event

See also:
BCPF-1456

### DIFF
--- a/plugins/browser-plugin-error-tracking/src/index.ts
+++ b/plugins/browser-plugin-error-tracking/src/index.ts
@@ -85,7 +85,7 @@ export function trackError(
           schema: 'iglu:com.snowplowanalytics.snowplow/application_error/jsonschema/1-0-1',
           data: {
             programmingLanguage: 'JAVASCRIPT',
-            message: truncatedMessage ?? "JS Exception. Browser doesn't support ErrorEvent API",
+            message: truncatedMessage ?? 'trackError called without required message',
             stackTrace: stack,
             lineNumber: lineno,
             lineColumn: colno,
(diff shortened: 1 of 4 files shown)


plugins/browser-plugin-error-tracking/src/index.ts
```diff
@@ -104,9 +104,9 @@ export function trackError(
  */
 export interface ErrorTrackingConfiguration {
   /** A callback which allows on certain errors to be tracked */
```
matus-tomlein, line 106: Seems like there is a word missing in this sentence.

plugins/browser-plugin-error-tracking/src/index.ts
```diff
@@ -147,15 +147,35 @@ function sendError(
...
+        lineno: errorEvent.lineno,
+        colno: errorEvent.colno,
+        error: errorEvent.error,
+        context,
+      },
+      trackers
+    );
+  } else if (errorEvent.target && 'tagName' in errorEvent.target) {
+    const element: any = errorEvent.target;
+    trackError(
```
matus-tomlein, line 164: Might have missed this, but can we cover this branch (and perhaps also the one below) with a unit test as well?

matus-tomlein: Just realized that the `sendError` function is not exported. Leaving the comment here if you can trigger the event listener in an easy way and test the whole handling, otherwise feel free to skip this.

matus-tomlein (approved):
LGTM! 👍

Issue #798, opened 2020-03-03
Fix silent failure if OptimizelyX is enabled but optimizely.get doesn't exist


colmsnowplow:
We added a [check](https://github.com/snowplow/snowplow-javascript-tracker/blob/7203e5fa2db5c97aa5cf1359deb578a12eb033fe/src/js/tracker.js#L1064) on optimizely.get to prevent the tracker from throwing if the function doesn't exist in `getOptimizelyXData()`.

However, in the scenario where it doesn't exist, this function returns undefined, and then [we attempt to call `getActiveExperimentIds()` and `getVariationMap()`](https://github.com/snowplow/snowplow-javascript-tracker/blob/7203e5fa2db5c97aa5cf1359deb578a12eb033fe/src/js/tracker.js#L1102-L1103) against the output.

This appears to cause the tracker to silently fail without sending data. Likely rectified by adding a further check, or moving the existing check to `getOptimizelyXSummary`.

PR #1343, opened 2024-09-17
Handle optimizely.get not yet existing

jethron:
Fixes a regression from 24928acd08164d42f95e3c1de88f63d6cdef1b67 that undoes the fix from #798.

If `window.optimizely` exists, but `window.optimizely.get` does not yet exist, we never populate `state`.
The fix for #798 added checks for dereferencing methods directly on the result, but provided no default for `getActiveExperimentIds()` meaning `experiment_ids` would also be `undefined` if `state` was.

Originally the code used `lodash/map` to then iterate over `experiment_ids`, that was then updated to directly be `experiment_ids.map`; with no default this tries to dereference `undefined` to call the method, whereas `lodash/map` accepts `undefined` as a value and returns an empty array.

This changes the plugin to:
- bail out early if `state` isn't correctly populated
- change some types to remove some cases of `any` that were obscuring the error
- note that the global types are optional and may be `undefined`
- removes the code for extracting subproperties, as it's not used and obscures the return type of `getOptimizelyXData`

### DIFF
--- a/plugins/browser-plugin-optimizely-x/src/index.ts
+++ b/plugins/browser-plugin-optimizely-x/src/index.ts
@@ -33,6 +33,14 @@ import { OptimizelyxSummary } from './contexts';
 
 declare global {
   interface Window {
+    /*
+    The following property is technically optional since the Optimizely SDK
+    loads asynchronously; the inner `get` is also optional for the older SDK
+    and during brief periods where the SDK is only partially loaded.
+    As Window interface type conflicts can break builds, we leave them as
+    non-optional for backwards compatibility, but changes to this plugin should
+    be aware that this interface may not be fully available. See #1343
+    */
     optimizely: {
       get: (property: string) => { [key: string]: any };
     };
@@ -47,16 +55,12 @@ export function OptimizelyXPlugin(): BrowserPlugin {
    * Check that *both* optimizely and optimizely.get exist
    *
    * @param property - optimizely data property
-   * @param snd - optional nested property
    */
-  function getOptimizelyXData(property: string, snd?: string) {
+  function getOptimizelyXData(property: string) {
     const windowOptimizelyAlias = window.optimizely;
-    let data;
+    let data: Record<string, any> | undefined;
     if (windowOptimizelyAlias && typeof windowOptimizelyAlias.get === 'function') {
       data = windowOptimizelyAlias.get(property);
-      if (typeof snd !== 'undefined' && data !== undefined) {
-        data = data[snd];
-      }
     }
     return data;
   }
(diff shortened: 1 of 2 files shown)


plugins/browser-plugin-optimizely-x/src/index.ts
```diff
@@ -33,8 +33,8 @@ import { OptimizelyxSummary } from './contexts';
 
 declare global {
   interface Window {
-    optimizely: {
-      get: (property: string) => { [key: string]: any };
+    optimizely?: {
```
matus-tomlein, line 36: Just to be sure – do you see any potential of this being a breaking change? When I ran `rush build`, it first gave me this error:

```
==[ @snowplow/browser-plugin-optimizely-x ]======================[ 13 of 34 ]==                   
[!] (plugin Typescript) TS2687: All declarations of 'optimizely' must have identical modifiers.   
src/index.ts (36:5)                                                                               
                                                                                                  
36     optimizely?: {                                                                             
       ~~~~~~~~~~                                                                                 
                                                                                                  
                                                                                                  
Returned error code: 1                                                                            
"@snowplow/browser-plugin-optimizely-x" failed to build.  
```

After running `rush clean`, it worked. This made me think that maybe in case users have defined their own global Window type with optimizely, it could be breaking for their build to change the type here? Maybe this is not an issue at all or it's super unlikely to be an issue, so I don't think it needs to be a blocker, but I thought I'd check.

matus-tomlein (approved):
LGTM!

plugins/browser-plugin-optimizely-x/src/index.ts
matus-tomlein: Hmm, I'm just thinking maybe the error was a conflict in the type definition with the older optimizely plugin? It [defines the optimizely type here](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/plugins/browser-plugin-optimizely/src/index.ts#L44).

jethron: I also had this issue, but the "conflicting definition" was the same plugin's `dist` output, so I figured it was some kind of bug in our tsconfig include paths or something, but you could be right.

The Optimizely SDK itself doesn't seem to declare any type at the global level so I'm not sure what the best approach would be. I don't mind backing out the type change and just keeping the defensive code?

matus-tomlein: Yeah, I think you're right that the conflict was with the `dist` output of the same plugin, I just ran into it again and removing the `dist` folder and building again solved it.

But it seems reasonable to keep the defensive code but undo the type change if we're not 100% sure about what the type change might do... Most likely won't be a problem, but I'm not super confident about it.

PR #1344, opened 2024-09-24
Migrate html5 plugin to new media plugin

greg-el:
This PR refactors the tracking for the HTML media plugin to use the Snowplow Media Tracking plugin. I've also copied the `DynamicContext` changes to the Media plugin present in the YouTube PR to send the HTML entities.

API Changes:
- The `id` arg in the config passed by the user has been renamed `video` to be consistent with other media trackers, and avoid any confusion with `id` in the Media plugin
- It now accepts either a string (HTML id) or a `HTMLMediaElement`.


plugins/browser-plugin-media-tracking/src/player.ts
```diff
@@ -0,0 +1,174 @@
...
+type VideoMediaPlayerUpdate = Pick<MediaPlayerUpdate, 'fullscreen' | 'pictureInPicture'>;
+
+function updatePlayer(el: HTMLMediaElement): MediaPlayerUpdate {
+  const common: CommonMediaPlayerUpdate = {
+    currentTime: el.currentTime || 0,
+    duration: getDuration(el),
+    ended: el.ended,
+    livestream: el.duration === Infinity,
+    loop: el.loop,
+    mediaType: el.tagName === 'VIDEO' ? MediaType.Video : MediaType.Audio,
```
matus-tomlein, line 35: Why not use the `isHtmlVideoElement` function from the `helpers.ts` file?

plugins/browser-plugin-media-tracking/src/player.ts
```diff
@@ -0,0 +1,174 @@
...
+    });
+  }
+
+  video.addEventListener('timeupdate', () => {
+    updateMediaTracking({ id, player: updatePlayer(video) });
+  });
+
+  let isWaiting = false;
+
+  for (const event of config.captureEvents ?? []) {
```
matus-tomlein, line 89: So by default we don't capture any events? Shouldn't we have some default list of events to capture?

matus-tomlein: It might be better to subscribe to all the events regardless of the `captureEvents` setting and let the underlying media package filter them using the `captureEvents` property. This will let the package update the media session entity correctly (based on the play, pause, volume events), but it won't track events that the user doesn't want (by default it will track all events).

plugins/browser-plugin-media-tracking/src/player.ts
```diff
@@ -0,0 +1,174 @@
...
+  if (el instanceof HTMLVideoElement) {
+    context.push(() => buildHTMLVideoElementEntity(el));
+  }
+
+  return context;
+}
+
+export function setUpListeners(config: ElementConfig) {
+  const { video } = config;
+  const id = uuidv4();
```
matus-tomlein, line 65: We should let users provide their own `id` (the media session id) similar like we do in the Vimeo plugin. Not sure we can make it generate a uuid by default because then the user is unable to stop the media tracking?

plugins/browser-plugin-media-tracking/src/api.ts
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * **Ping Configuration (Optional):**
+ * - `config.pings.pingInterval` (`number`): The interval (in seconds) for sending ping events. Default is 30(s).
+ * - `config.pings.maxPausedPings` (`number`): The maximum number of ping events sent while playback is paused. Default is 1.
+ *
+ * **Other Options (Optional):**
+ * - `config.updatePageActivityWhilePlaying` (`boolean`): Whether to update page activity while media is playing. Enabled by default.
+ * - `config.filterOutRepeatedEvents` {@link FilterOutRepeatedEvents}: Whether to suppress consecutive identical events. Default is false.
+ */
+export function enableMediaTracking(config: Config) {
```
matus-tomlein, line 53: I think we will need a `disableMediaTracking` function as well – that one should unsubscribe the listeners and call `endMediaTracking` on the underlying media package.

plugins/browser-plugin-media-tracking/src/api.ts
```diff
@@ -0,0 +1,70 @@
...
+  };
+}
+
+/**
+ * Enables media tracking on an HTML `<video>` or `<audio>` element.
+ *
+ * @param {Config} config - Configuration options for media tracking.
+ *
+ * **Required:**
+ * - `config.video` (string | HTMLMediaElement): The ID of the media element or the element itself.
```
matus-tomlein, line 37: Here we should also accept the media session id.

plugins/browser-plugin-media-tracking/src/api.ts
matus-tomlein: Come to think of it, do you think we should rename the functions to `startMediaTracking` and `endMediaTracking` to reflect the Vimeo and media plugins?

If we keep it `enableMediaTracking` and add the `id` property (for the media session id), this will be rather tricky for users to upgrade because we will be changing the meaning of the `id` property – previously it identified the video element, now the media session. So I'd prefer if we change the function completely to make users' apps break rather than fail silently.

plugins/browser-plugin-media-tracking/src/player.ts
greg-el: Sounds good, `captureEvents` is already passed into the `startMediaTracking` config, so I've just removed the switch so all events listeners are added.

plugins/browser-plugin-media-tracking/src/api.ts
greg-el: That sounds like a sensible idea to me.

plugins/browser-plugin-media-tracking/src/api.ts
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * **Ping Configuration (Optional):**
+ * - `config.pings.pingInterval` (`number`): The interval (in seconds) for sending ping events. Default is 30(s).
+ * - `config.pings.maxPausedPings` (`number`): The maximum number of ping events sent while playback is paused. Default is 1.
+ *
+ * **Other Options (Optional):**
+ * - `config.updatePageActivityWhilePlaying` (`boolean`): Whether to update page activity while media is playing. Enabled by default.
+ * - `config.filterOutRepeatedEvents` {@link FilterOutRepeatedEvents}: Whether to suppress consecutive identical events. Default is false.
+ */
+export function startMediaTracking(config: Config) {
```
matus-tomlein, line 55: Sorry I may have misled you – I think now we have a clash between the `startMediaTracking` function in this plugin and the one in the media plugin. I think that is the reason why the integration tests don't work for me – when I look into the compiled `sp.js` function, I can see that it is the one from the other media plugin.

Can we rename this function to `startHtml5MediaTracking` (or something similar) and also rename the `endMediaTracking` one?

matus-tomlein: You can also rebase on top of the `release/4.0.0` branch which disables the youtube integration tests that are failing here as welll.

trackers/javascript-tracker/test/pages/media/tracking.html
```diff
@@ -33,12 +33,9 @@
       });
 
       window.snowplow('enableMediaTracking', {
```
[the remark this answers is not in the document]
matus-tomlein: Just noting that this should be fixed for all the media pages in the integration tests.

matus-tomlein (commented):
Once we fix the integration tests and validate that things work, I am happy to approve this! The code looks great!

plugins/browser-plugin-media-tracking/src/helperFunctions.ts
```diff
@@ -114,12 +49,41 @@ export function dataUrlHandler(url: string): string {
...
+      language: textTrackList[i].language,
+      kind: textTrackList[i].kind,
+      mode: textTrackList[i].mode,
+    };
+  });
+}
+
+export function setConfigDefaults(config: Config): Config {
+  const defaults = {
+    captureEvents: HTMLEventTypes,
```
matus-tomlein, line 80: Do we need to set the defaults? Does it change anything compared to not specifying `captureEvents` if not passed by user? If we don't specify `captureEvents`, the tracker will track all events tracked by the media-tracking plugin.

If there is a reason to specify the default `captureEvents`, please add ping event type to them. That is tracked by default in the media plugin but because we override this here, it is not tracked anymore.

matus-tomlein (approved):
LGTM, nice work! 👍 

Just one more comment about the default capture events – either remove the defaults or add ping events there.

plugins/browser-plugin-media-tracking/src/helperFunctions.ts
greg-el: Yep we can get rid of that now. I've added ping and percentprogress to `HTML5MediaEventType` also.

PR #1345, opened 2024-09-25
Remove browser-features, optimizely, ecommerce and consent plugins, deprecate enhanced ecommerce and performance timing plugins and add web vitals by default

matus-tomlein:
This PR revises the plugins in the v4 release. The goals are:

1. Make it more straightforward for customers to choose the correct plugins (so they don't have to choose from 3 ecommerce plugins and steer towards the ones supported by our dbt packages)
2. Remove old plugins that build on deprecated APIs (e.g., optimizely)

Below is the list of changes. This is still open for discussion if you have a different opinion!

**Removed plugins**
* browser-features
    * has already been deprecated in v3
* optimizely
    * builds on deprecated optimizely APIs
* ecommerce
    * we don't support it in our dbt packages, superseded by snowplow-ecommerce
* consent
   * superseded by enhanced-consent, not sure if there is value in just deprecating?

**Deprecated plugins**
* enhanced ecommerce
   * I thought that since this one reflect's the GA Enhanced Ecommerce, it's worth keeping it, but happy to remove as well
* performance timing
   * was considering removing this but it is currently the most downloaded plugin on NPM and although most of the APIs it builds on are deprecated, not all

**Added default plugins in sp.js**
* web vitals (needs configuration to be enabled)

**Removed default plugins in sp.js**
* optimizely-x
* timezone

**Other changes**
* Add an option to enable performance navigation timing


igneel64 (approved):
🚀

trackers/javascript-tracker/src/features.ts
```diff
@@ -197,5 +147,15 @@ export function Plugins(configuration: JavaScriptTrackerConfiguration) {
     activatedPlugins.push([EventSpecificationsPlugin(), apiMethods]);
   }
 
+  if (plugins.performanceNavigationTiming) {
+    const { PerformanceNavigationTimingPlugin, ...apiMethods } = PerformanceNavigationTiming;
+    activatedPlugins.push([PerformanceNavigationTimingPlugin(), apiMethods]);
+  }
+
+  if (plugins.webVitals) {
```
jethron, line 155: I think this should also be guarded by a `configuration.contexts` flag to be disabled by default since the default behaviour is to load an external script. I think that should be expressly opt-in behaviour for the SDK.

trackers/javascript-tracker/tracker.config.ts
```diff
@@ -1,30 +1,28 @@
...
 export const errorTracking = true;
 export const formTracking = true;
 export const gaCookies = true;
-export const geolocation = true;
 export const linkClickTracking = true;
-export const performanceTiming = true;
+export const performanceNavigationTiming = true;
 export const siteTracking = true;
 export const snowplowEcommerceTracking = true;
 export const timezone = true;
```
jethron, line 11: I'd probably argue for excluding `timezone` after #1338.

jethron (commented):
I would prefer `consent` and `ecommerce` stick around as deprecated while they're still "not deprecated" and first-class (or the only built-in option) in other platform's trackers. But not a hill I'd die on.

trackers/javascript-tracker/src/features.ts
matus-tomlein: Good point! Have added a configuration for the web vitals plugin.

jethron (approved):
LGTM!

PR #1347, opened 2024-10-04
Fix SelfDescribingJson type regression from #1330

jethron:
Fixes a regression from the change in #1330 where in trying to prohibit arrays as SDJ `data` values, I accidentally mandated that types have a string index property defined. :sweat_smile: 

Instead, allow any object, and just explicitly bar arrays as values. This moves the type error from the type signature to the value assignment, but I don't see a better way for TypeScript to define "accepts objects but not arrays" in type parameter constraints while still having a non-inferred default type for that parameter.


matus-tomlein (approved):
LGTM! Could you add a test that tracks a self-describing event using a type generated by Snowtype so that we can catch any regressions in the future?

For instance you can use something like this:
```ts
interface EventSpecification {
  id: string
  name: string
  data_product_id: string
  data_product_name: string
}
```

jethron:
> Could you add a test that tracks a self-describing event (and maybe also an entity?) using a type generated by Snowtype so that we can catch any regressions in the future?

Great idea! That found more issues, so I ended up needing to change way more stuff than I'd hoped. :/

This is probably a good integration test idea, to generate some known data product with the latest version of snowtype to check for regressions. Needs adding a secret to the repo though so I can't do that myself.

matus-tomlein (approved):
Thanks for adding the Snowtype generated code to tests!

It does seem more complicated but as far as I can tell it seems to work!

PR #1348, opened 2024-10-04
Use capture-phase listeners for button click tracking

jethron:
In v4 we switched the Form Tracking (#1329) and Link Click Tracking (#1325)  plugins to using capture-phase event listeners instead of bubble-phase.

This makes the same change for the Button Click Tracking plugin to make it consistent.

This has caused issues for some clients historically, e.g. BCPF-1421 where pages stopping propagation of event bubbling prevented buttons being tracked.


matus-tomlein (approved):
LGTM!

PR #1349, opened 2024-10-04
Switch to GA4 as default for gaCookies plugin

jethron:
This changes the default behaviour of the GA Cookies plugin to look for GA4 cookies instead of Universal Analytics cookies.

This uses the `iglu:com.google.ga4/cookies/jsonschema/1-0-0` schema instead of `iglu:com.google.analytics/cookies/jsonschema/1-0-0`, so it is a breaking change.


matus-tomlein (approved):
LGTM!

PR #1350, opened 2024-10-04
Update discoverRootDomain default behaviour

jethron:
In v3, `discoverRootDomain` was disabled by default, and enabling it meant ignoring the `cookieDomain` configuration. If neither were set, cookies would have no `Domain` attribute attached and would bind to the current domain; cookies with no domain are unique in that they do not get shared to subdomains.

In 257ddb31198e48add33c4a39acdb0fdc55d9b248 we updated the `discoverRootDomain` setting to default to enabled, and changed how it interacted with the `cookieDomain` setting so that `discoverRootDomain` is essentially always enabled, but gets ignored if `cookieDomain` is also specified. This however made it impossible to use the 'no domain' behaviour available in v3, because disabling `discoverRootDomain` would have no effect without specifying `cookieDomain` as well.

This can be slightly confusing, so instead, this change allows explicitly disabling `discoverRootDomain` even if no `cookieDomain` is specified, allowing the same behaviour as all configurations available in v3.

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/discover-root-default_2024-10-04-03-51.json
+++ b/common/changes/@snowplow/browser-tracker-core/discover-root-default_2024-10-04-03-51.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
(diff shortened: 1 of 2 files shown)


matus-tomlein (approved):
LGTM!

PR #1351, opened 2024-10-04
Support events on buttons/links/forms in ShadowDOMs

jethron:
When DOM events on elements within Custom Components/ShadowRoots trigger against the `document`, the `target` is set to the Custom Element/shadow host rather than the actual element that caused the event.

This change adds support to the from/button/link tracking plugins to detect when this occurs, and use the first element from the Event's [`composedPath`](https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath) instead of the `target`. This enables them to see the actual even target and fire correctly in this scenario.

For the Form tracker, only `focus` events are considered `composed`, so `change` and `submit` events don't actually get reported outside the shadow DOM. As a workaround, when the `focus` events occur, if enabled the change/submit event listeners are bound directly to the form (or field if it has no associated form) rather than just the `document`.

### DIFF
--- a/common/changes/@snowplow/browser-plugin-button-click-tracking/shadow-clicks_2024-10-04-05-19.json
+++ b/common/changes/@snowplow/browser-plugin-button-click-tracking/shadow-clicks_2024-10-04-05-19.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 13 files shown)


matus-tomlein (approved):
LGTM! 👍

PR #1352, opened 2024-10-04
Disable b64 for SDJ data on POST requests from browsers by default

jethron:
In POST requests this usually results in a smaller payload size, and is easier to debug. The node tracker shares this setting across multiple emitters that may not all use POST so we can't make this assumption there, because it is considerably less efficient when used for GET requests.


matus-tomlein (approved):
Yay! 🥳 

Should we update the API docs in the `TrackerConfiguration` type to state the default is false for POST requests?

PR #1354, opened 2024-10-15
Make base64 encoding an optional parameter in Node newTracker call

matus-tomlein:
This PR makes the `encodeBase64` parameter in the `newTracker` call of the Node tracker optional. It keeps the default true setting.


jethron (approved):
LGTM!

Issue #1355, opened 2024-10-16
Tracker Version v4: Pending Fetch API Requests Accumulating in OutQueue on Chrome


j7i:
### Description
We are currently conducting tests on the new tracker version v4 and have encountered an issue where requests become stuck in a pending state.

Initially, the OutQueue functions correctly, effectively sending out events. However, as more events are triggered, they begin to accumulate in the OutQueue (local storage) and fail to be sent.

These accumulated events are only dispatched upon refreshing the page, given that they are recorded in the local storage via the OutQueue.

This problem appears with the latest version of the Chrome browser. We have not observed the same issue in either Firefox or Safari, where the system operates as expected.

--

We're addressing this issue here to keep everyone who might be experiencing the same problem updated on potential solutions and causes. Moreover, we warmly invite anyone facing similar issues to share their insights and findings. If there's a more effective platform to report our progress, please don't hesitate to let us know.

### Steps to Reproduce
At this time, we do not have a public source available for reproducing this issue. However, we are actively investigating the problem on our end to ensure that there are no factors within our setup that may be causing interference.

We are open to sharing our findings and insights along with details of our setup, potentially even in a call, and we are exploring ways to provide something that could help reproduce the issue.

### Expected behavior
The expected behavior involves the successful dispatch of all event requests. When an event is triggered, it should be sent out efficiently without being trapped in the OutQueue (local storage). No event should remain in a pending state for an extended period of time.

### Screenshots
![image](https://github.com/user-attachments/assets/e15bb67b-7c1d-429c-8111-f719f1bb062a)
![screen-recording-example](https://github.com/user-attachments/assets/b4a953c0-1c0b-4f49-9112-a4170500c265)

### Environment
 - macOS 14.6.1 (23G93)
 - Chrome
 - Version 130.0.6723.59 (Official Build) (arm64)
 - @snowplow/browser-tracker: 4.0.0-beta.2

### Additional context
While we're looking into the problem, we've found that some browsers might limit "keepalive" requests. These limits could be on how many requests can be sent at the same time, or on the total size of data that can be sent in these requests. This could be part of the issue we're dealing with.
And this might not be the problem in our case as we are not seeing any failed events (pending only).

Interesting information/reads about keepalive:
- https://fetch.spec.whatwg.org/#request-keepalive-flag
- https://github.com/whatwg/fetch/issues/679
- https://github.com/whatwg/fetch/issues/662
[...]

PR #1356, opened 2024-10-17
Enable batching multiple events into a single POST request over buffer size if maxPostBytes setting is followed

matus-tomlein:
This PR enables POST requests to contain more events than configured through `bufferSize` as long as they don't surpass the `maxPostBytes`.

The reason for this is to send events queued up in the event store in a single POST request rather than individually. This may happen for instance if some events didn't manage to be saved on the previous page visit (or the collector was not reachable). With this feature, they should be sent in one batch request on the next visit.

It may also happen in events are tracked right after each other – if they manage to be saved to the event store before the flush function is called, they will be batched together.

The change might be useful also with issue #1355 where part of the problem could be too many requests done shortly after each other.

### DIFF
--- a/libraries/tracker-core/src/emitter/index.ts
+++ b/libraries/tracker-core/src/emitter/index.ts
@@ -326,7 +326,7 @@ export function newEmitter({
       LOG.warn('Event (' + bytes + 'B) too big, max is ' + maxBytes);
 
     if (usePost) {
-      const bytes = emitterEvent.getPOSTRequestBytesCount();
+      const bytes = emitterEvent.getPOSTRequestBytesCount() + 88; // 88 bytes for the payload_data envelope
       const tooBig = bytes > maxPostBytes;
       if (tooBig) {
         eventTooBigWarning(bytes, maxPostBytes);
--- a/libraries/tracker-core/src/emitter/index.ts
+++ b/libraries/tracker-core/src/emitter/index.ts
@@ -47,6 +47,7 @@ export interface EmitterConfigurationBase {
   bufferSize?: number;
   /**
    * The max size a POST request can be before the tracker will force send it
+   * Also dictates the max size of a POST request before a batch of events is split into multiple requests
    * @defaultValue 40000
    */
   maxPostBytes?: number;
@@ -312,7 +313,6 @@ export function newEmitter({
       customHeaders,
       connectionTimeout,
       keepalive,
-      bufferSize,
       maxPostBytes,
       useStm,
       credentials,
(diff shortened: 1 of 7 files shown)


libraries/tracker-core/src/emitter/index.ts
jethron, line 1: I think on line L403 here we're meant to check `if (success || retry)` instead of just `if(success)`? I think that's the issue, the fetch immediately throws when we call `customFetch` because of the quota limits, and then the retry never happens until the next event is generated; but if they also fail we just kind of deadlock...

With the above in place we still end up with a bunch of `pending` requests, but eventually it does succeed and send the full batch properly. Then I think the `pending` requests are just a weird artefact in devtools that don't actually contribute towards the limit?

A small refactor to move the AbortController up next to `timer` and then expose `AbortController.abort()` in `EmitterRequest` and calling it on L404 like `if(!success) request.abort("Retrying");` cleans up most of the "pending" requests except 1, so that's an improvement too.

jethron (approved):
Nice fix! Batching works as expected again with this change, I can trigger 100 events in quick succession and it's split into 3 batches, rather than 100 requests without this change.

The `Content-Length` is slightly larger than `maxPostBytes` though, should we take this opportunity to account for `payload_data` overhead or something? Even just a static `>= (maxPostBytes - 200)` or something would probably be adequate. (We've suggested this setting to some people who have a WAF in front of the collector that blocks requests over a certain size; ideally we'd treat it a bit more strictly, but doesn't have to be perfect; we can always just advice under-specifying it.)

My comment above should help address #1355, let me know if you want me to do a dedicated PR or you can just include it here.

libraries/tracker-core/src/emitter/index.ts
jethron: (We may also want to `cancelTimeoutTimer()` in the `catch` branch, too; the printed error is confusing in this failure scenario!)

matus-tomlein: Good points! I am a bit unsure if we should call `continueFlush` right after a request fails – I think [previously](https://github.com/snowplow/snowplow-javascript-tracker/blob/master/libraries/browser-tracker-core/src/tracker/out_queue.ts#L282) we stopped sending after a request fails and would retry on the next event. I am just worried of introducing an infinite loop of retries without a delay in between.

Do you think we should add a timeout, maybe 1-5 seconds after which to retry?

jethron: Oh, ok, if that's the previous behaviour we can probably keep it as-is.

Then just aborting and cancelling the timeout after we catch to clean up the requests should be fine with batches fixed.

PR #1358, opened 2024-10-21
Read the request response to avoid overlapping requests to collector

matus-tomlein:
Aims to address issue #1355.

This PR adds an await for the response body to be read before continuing with the emitter loop. Previously it was possible that requests overlapped since the emitter could start a new request (for follow up events) before the previous response was fully received. This could have caused problems with the fetch keepalive limitations.


jethron (approved):
LGTM!

PR #1359, opened 2024-10-23
Change the default value for keepalive to false

matus-tomlein:
This PR disables fetch keepalive by default.

The reason for this is the 64KB limit that browsers put on the keepalive requests. Although the default `maxPostBytes` is under this limit so we shouldn't run into for single requests, the limit is shared for parallel requests. We have seen this when testing in internal apps that have multiple initialized trackers that send events at once – this caused the requests to go into a pending state and fail.

A related problem was reported in #1355 – I think we were able to fix this for the case where there is a single tracker instance sending events, but not for the multiple tracker instances case. Since we can't fix it for that case, we can't keep the default setting to true.

### DIFF
--- a/libraries/tracker-core/src/emitter/emitter_request.ts
+++ b/libraries/tracker-core/src/emitter/emitter_request.ts
@@ -88,7 +88,7 @@ export function newEmitterRequest({
   eventMethod = 'post',
   customHeaders,
   connectionTimeout,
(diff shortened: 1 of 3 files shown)


jethron (approved):
LGTM!

All "solutions" to this problem continue to be cursed, I guess. :sweat_smile:

PR #1360, opened 2024-10-24
Fill empty targetUrl for link clicks with "about:nothing" to avoid failed events and output a warning

matus-tomlein:
In case a click on a link with empty `href` is tracked, set `about:nothing` as the `targetUrl` in the event to avoid the event failing validation.

Also logs a `console.warn` to help users identify and fix these issues.

### DIFF
--- a/plugins/browser-plugin-link-click-tracking/src/index.ts
+++ b/plugins/browser-plugin-link-click-tracking/src/index.ts
@@ -190,6 +190,8 @@ export function trackLinkClick(
 
 /**
  * Process a clicked element into a link_click event payload.
+ * 
+ * In case the href of the element is empty, "about:invalid" is used as the target URL.
  *
  * @param sourceElement The trackable element to be used to build the payload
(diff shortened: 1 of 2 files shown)


plugins/browser-plugin-link-click-tracking/src/index.ts
```diff
@@ -210,9 +212,13 @@ function processClick(sourceElement: TrackableElement, includeContent: boolean =
     elementTarget = anchorElement.target;
     elementContent = includeContent ? anchorElement.innerHTML : undefined;
 
+    if (!targetUrl) {
+      console.warn('Link click target URL empty', anchorElement);
```
jethron, line 216: I'm not sure on the semantic difference, but should we be using `_logger?.warn` instead of the `console` directly?

jethron (approved):
Otherwise LGTM!

jethron:
Maybe also retitle the PR to `about:invalid` rather than `about:nothing` so it shows up in search results properly.

PR #1361, opened 2024-10-25
Fix removal of context generator functions

jethron:
This fixes an issue when using `removeGlobalContexts` to remove a [context generator function](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/javascript-trackers/web-tracker/custom-tracking-using-schemas/global-context/#context-generators) when multiple generator functions have been added.

In 3.0.0 the method of identifying the `ContextPrimitive` to remove was changed from equality/identity to comparing the JSON serialization of the entities. Unfortunately, this doesn't handle the generator function case, because `JSON.stringify(function(){})` returns `undefined`, which means all added context generators appear identical, so removing one generator actually removes all of them.

This changes the comparison to be identity-based for functions, and remain JSON-based for static entities.

It also fixes matching on for `ConditionalContextProviders` by implementing a deep comparison. A purely identity-based approach is much simpler, but only a partial solution; non-identical providers with the same static entities added with different filter functions would still collide if the identities of the array don't match correctly.

### DIFF
--- a/common/changes/@snowplow/tracker-core/remove-context-generators-v3_2024-10-25-00-04.json
+++ b/common/changes/@snowplow/tracker-core/remove-context-generators-v3_2024-10-25-00-04.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/tracker-core",
+      "comment": "Fix removal of global context generators",
+      "type": "none"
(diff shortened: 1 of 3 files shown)


matus-tomlein (approved):
LGTM, thanks Jethro!

PR #1362, opened 2024-10-25
Fix removal of context generator functions (v4)

jethron:
This is the same as #1361 but for v4.


matus-tomlein:
I have rebased the `release/4.0.0` branch on top of `master`, so I think this PR is not necessary since we got the change from the v3 release in.

PR #1363, opened 2024-10-25
Release/3.24.5

matus-tomlein:
This patch release fixes a problem with removing global context generator functions in case multiple were used.

**Bug fixes**
- Fix removal of context generator functions (#1361)


jethron (approved):
LGTM

PR #1365, opened 2024-10-28
Allow 0x0 viewport in browser context to avoid schema validation errors

matus-tomlein:
This PR allows the viewport property in the [browser_context schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/browser_context/jsonschema/2-0-0) to be 0x0 instead of making it null. The property is required in the schema, so null caused validation errors.

I think it's sensible to allow the 0x0 values since we do the same for `documentSize` and `resolution` properties in the same context entity. I don't see why we should make it null in that case especially since it's causing failed events.


greg-el (approved):
LGTM

PR #1369, opened 2024-11-01
Fix initialization of the HTML5 media tracking plugin so that it initializes the parent media plugin

matus-tomlein:
This PR fixes a bug in the v4 HTML5 media tracking plugin in that it didn't initialize the trackers in the parent media plugin which then couldn't track any events. The fix is the same as in the YouTube and Vimeo plugins to make sure the `activateBrowserPlugin` function from the media plugin is propagated from the media tracking one as well.


jethron (approved):
LGTM!

PR #1370, opened 2024-11-01
Release/4.0.1

matus-tomlein:
This patch release fixes an issue in the HTML5 media tracking plugin in that it didn't correctly initialize and failed to track events.

**Bug fixes**

- Fix initialization of the HTML5 media tracking plugin so that it initializes the parent media plugin (#1369)

**Under the hood**

- Fix broken API docs generation by adding explicit type for EventStoreIteratorNextResult (#1367)


jethron (approved):
LGTM!

PR #1371, opened 2024-11-05
Add new tracker for React Native

matus-tomlein:


trackers/react-native-tracker/README.md
```diff
@@ -0,0 +1,86 @@
...
+git clone https://github.com/snowplow/snowplow-javascript-tracker.git
+rush update
+```
+
+## Package Installation
+
+With npm:
+
+```bash
+npm install @snowplow/node-tracker
```
greg-el, line 26: Should this be `@snowplow/react-native-tracker`? (A couple of others in the doc too).

trackers/react-native-tracker/src/types.ts
```diff
@@ -0,0 +1,631 @@
...
+  /**
+   * Sets subject data
+   *
+   * @param config - The new subject data
+   */
+  readonly setSubjectData: (config: SubjectConfiguration) => void;
+
+  // TODO:
+  // /**
+  //  * Gets the dentifier for the user of the session
```
greg-el, line 547:
```suggestion
  //  * Gets the identifier for the user of the session
```

trackers/react-native-tracker/src/subject.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  const addSubjectToPayload = (payload: PayloadBuilder): void => {
+    payload.add('duid', domainUserId);
+    payload.add('nuid', networkUserId);
+  };
+
+  const setScreenResolution = (screenSize: ScreenSize) =>
+    core.setScreenResolution(String(screenSize[0]), String(screenSize[1]));
+
+  const setNetworkUserId = function (userId: string | undefined) {
```
greg-el, line 16: More of a style nitpick than anything else, but `const foo = function() {}` feels like (in the words of Ryan) a real Hannah Montana moment. I'm not too fussed as long as it's consistent, the `const foo = () => {}` syntax seems to be used above which I see more.

trackers/react-native-tracker/src/subject.ts
```diff
@@ -0,0 +1,87 @@
+import { CorePluginConfiguration, PayloadBuilder, TrackerCore } from '@snowplow/tracker-core';
+import { ScreenSize, SubjectConfiguration } from './types';
+
+export function newSubject(core: TrackerCore, configuration?: SubjectConfiguration) {
+  let domainUserId: string | undefined;
+  let networkUserId: string | undefined;
+
+  const addSubjectToPayload = (payload: PayloadBuilder): void => {
```
greg-el, line 8: Another style nitpick, there's a return type here which the other functions declared in this scope don't have. Not fussed again which way round you want to do it, as long as it's consistent.

trackers/react-native-tracker/src/subject.ts
```diff
@@ -0,0 +1,87 @@
+import { CorePluginConfiguration, PayloadBuilder, TrackerCore } from '@snowplow/tracker-core';
+import { ScreenSize, SubjectConfiguration } from './types';
+
+export function newSubject(core: TrackerCore, configuration?: SubjectConfiguration) {
```
greg-el, line 4: I would however like an explicit return type of sorts here.

trackers/react-native-tracker/src/subject.ts
```diff
@@ -0,0 +1,87 @@
...
+  const subjectPlugin: CorePluginConfiguration = {
+    plugin: {
+      beforeTrack: addSubjectToPayload,
+    },
+  };
+
+  return {
+    subjectPlugin,
+    properties: {
+      setUserId: core.setUserId,
```
greg-el, line 74: There's probably a good reason, but why are we making these these `core.set*` functions available both here and as part of `setSubjectData`?

trackers/react-native-tracker/src/events.ts
```diff
@@ -0,0 +1,55 @@
...
+  buildSelfDescribingEvent,
+  buildStructEvent,
+  PageViewEvent,
+  SelfDescribingJson,
+  StructuredEvent,
+  TrackerCore,
+} from '@snowplow/tracker-core';
+import { EventContext, MessageNotificationProps, TimingProps } from './types';
+
+export function newTrackEventFunctions(core: TrackerCore) {
```
greg-el, line 12: And an explicit return type here would be lovely.

trackers/react-native-tracker/src/subject.ts
matus-tomlein: It's mostly for user convenience so that they have an explicit setter for `setUserId` instead of the less intuitive `setSubjectData({userId: ..})`. Also trying to avoid a breaking change to come as close to the previous API as possible.

trackers/react-native-tracker/src/subject.ts
matus-tomlein: I agree that we should have well defined return types for publicly exported functions, but is it important for internal functions as well? I feel like it's more of a nuisance than useful for internal code as we don't need to specify a public contract for this code. It'd also create a lot of boilerplate here that I'm not sure if it's worth it.

trackers/react-native-tracker/src/events.ts
greg-el: Same as https://github.com/snowplow/snowplow-javascript-tracker/pull/1371#discussion_r1836715416.

trackers/react-native-tracker/README.md
Mookiies, line 1: Am I correct that this gets rid of the `removeTracker` method? 

I know this doesn't have the native tie-in plugins yet, but will that be something that's safe to get rid of? Thinking of the things like `application_foreground` that might get duplicated if you can't remove a tracker.

matus-tomlein: Thanks for looking into this and the feedback!

That's correct, we don't have the `removeTracker` function here. We can look into it adding it though, it's a good feedback.

But I'm not sure I understand the relation to the `application_foreground` event. The tracker shouldn't be removed when the app goes to background/foreground. In order to calculate session metrics correctly, the same tracker instance should be maintained. So normally you would only create a new tracker on app start and then keep the tracker instance for the duration of the app lifetime.

Mookiies: As an example we have a legacy part of the app and a new part, each with their respective trackers both track the application lifecycle. When we swap between the two we `removeTracker`. If we didn't I think we'd get duplicate events sent, which is why I mention `application_foreground`.

So the case I'm thinking about is: if you setup a plugin to automatically send events for you, how to you get it to stop?

greg-el (approved):
Nice work! LGTM

PR #1375, opened 2024-11-11
Add option to disable `performanceNavigationTiming` plugin

greg-el:
This PR adds the `performanceNavigationTiming` property to the `JavaScriptTrackerConfiguration`, allowing for the performance navigation timing plugin to be disabled in the javascript tracker.


matus-tomlein (approved):
LGTM!

Just note that this makes the performance navigation plugin disabled by default whereas previously it was enabled by default. I am not against this as it more closely aligns with the other configuration, but we should note it in the release. I hope that this comes early enough in the v4 cycle that we can affort to sneak in this "breaking change".

PR #1377, opened 2024-11-11
Fix postPath configuration being ignored (close #1376)

### LINKED ISSUE #1376: Version 4.0.1: custom collector POST path (postPath) is not applied
**Describe the bug**
custom collector POST path (postPath) is not applied, although it is set in the tracker Initialization

**To Reproduce**
1. Initialize a tracker with a custom postPath
2. check the collector POST path of event requests, the default /tp2 path remains

**Expected behavior**
If a postPath is set in the tracker Initialization, the  collector POST path needs to change accordingly

**Additional context**
Brave and ad blockers are blocking the /tp2, therefore we had to roll back todays tracker version v4 deployment.

matus-tomlein:
Issue #1376

In the v4 release, a bug was introduced that the `postPath` configuration option was ignored by the tracker – despite it being configured, the tracker is still using the default tp2 path.

This PR fixes the issue and makes sure to use the configuration option.

### DIFF
--- a/common/changes/@snowplow/tracker-core/issue-fix_post_path_2024-11-11-14-42.json
+++ b/common/changes/@snowplow/tracker-core/issue-fix_post_path_2024-11-11-14-42.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/tracker-core",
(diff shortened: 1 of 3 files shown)


greg-el (approved):
LGTM!

PR #1378, opened 2024-11-11
Fix ignored label property in the startHtml5MediaTracking call

matus-tomlein:
This PR fixes an issue where the `label` configuration option passed to the `startHtml5MediaTracking` call was ignored.


greg-el (approved):
LGTM!

PR #1379, opened 2024-11-11
Release/4.0.2

### LINKED ISSUE #1376: Version 4.0.1: custom collector POST path (postPath) is not applied

matus-tomlein:
This release fixes a bug that caused the `postPath` configuration option to be ignored when making requests to the collector. It also fixes a bug in the media tracking plugin which didn't add the configured label to media events.

In the tag-based JavaScript tracker, we added a new `contexts.performanceNavigationTiming` option to enable/disable the performance navigation timing context entity. Since the option was previously missing, the context was added to all events. Now the context is disabled by default.

**Bug fixes**

- Fix ignored label property in the startHtml5MediaTracking call (#1378)
- Fix postPath configuration being ignored (close #1376)
- Add option to disable `performanceNavigationTiming` plugin (#1375)

### DIFF
--- a/common/changes/@snowplow/browser-plugin-media-tracking/issue-fix_ignored_label_in_media_tracking_2024-11-11-15-16.json
+++ b/common/changes/@snowplow/browser-plugin-media-tracking/issue-fix_ignored_label_in_media_tracking_2024-11-11-15-16.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-plugin-media-tracking",
+      "comment": "Fix ignored label property in the startHtml5MediaTracking call",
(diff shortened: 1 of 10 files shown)


greg-el (approved):
🥾  LGTM!
🐛

PR #1382, opened 2024-11-18
Serialize new session cookie synchronously to avoid overlapping sessions (close #1381)

### LINKED ISSUE #1381: Session handling issue with Browser Tracker V4
**Describe the bug**
Significant increase in "artificial" sessions:
- consecutive sessions within less than 600 milliseconds
- most session have just 1 event, the majority zero page_views 
- Sessions from legitimate users, definitely not bots

First assumption: Maybe the issue is caused the new async cookie writes.

**To Reproduce**
Not clear if it can be reproduced, but very likely:
1. Use a Browser tracker version 4 (4.0.1  or 4.0.2) in production, initialize the tracker with keepalive: true
2. Check you modeled event data and search for session with only one event see query below.

_Disclaimer: this query only works with our data model, it is only an example to adapt the query to other data models:_
```SQL
WITH PREP_Sessions_Details AS (
SELECT
  DATE(StartTstampUTC) AS SessionStartDate,
  S.StartTstampUTC,
  S.EndTstampUTC,
  TIMESTAMP_DIFF(S.EndTstampUTC, S.StartTstampUTC, SECOND) AS SessionDuration_inSeconds,
  -- Calculate the time between the sessions of a NetworkUserId
  TIMESTAMP_DIFF(
    S.StartTstampUTC, 
    LEAD(S.StartTstampUTC, 1) OVER (PARTITION BY S.NetworkUserId ORDER BY StartTstampUTC DESC), MILLISECOND
  ) AS TimeBetweenSessions_inMilliSeconds,
  S.NetworkUserId,
  LEAD(S.NetworkUserId, 1) OVER (PARTITION BY S.NetworkUserId ORDER BY StartTstampUTC DESC) AS PreviousNetworkUserId,
  S.DomainSessionId,
  S.DeviceAndBrowserDetails.DeviceClass,
  S.DeviceAndBrowserDetails.AgentName,
  S.PageViews, -- PageViews per Session
  S.Events -- Events per Session
FROM
  `<<session_table>>`
WHERE TRUE
  AND DATE(S.StartTstampUTC, "Europe/Zurich") >= "2024-11-11"
  AND S.NetworkUserId IN (
    -- find all NetworkUserIds that have only one event and no page views, channel: direct
    SELECT
      NetworkUserId
    FROM
      `<<session_table>>`
    WHERE TRUE
      AND DATE(StartTstampUTC, "Europe/Zurich") >= "2024-11-12"
      AND TrafficAttributionChannel_Array[SAFE_OFFSET(0)].TrafficAttributionChannel.Level1 = "Direct"
      AND PageViews = 0
      AND Events = 1
  )
GROUP BY
  ALL
)

SELECT
  *,
  COUNT(DISTINCT DomainSessionId) OVER(PARTITION BY NetworkUserId, SessionStartDate) AS NumberOfSessionsPerDayAndUser
FROM
  PREP_Sessions_Details  
WHERE TRUE
-- only show consecutive sessions of the same networkuserid and the time between the sessions is less than 1 second or null to catch the first session
  AND (TimeBetweenSessions_inMilliSeconds < 1000 OR TimeBetweenSessions_inMilliSeconds IS NULL)
QUALIFY
-- only show users that have at least 2 sessions per day
  NumberOfSessionsPerDayAndUser >= 2
ORDER BY
  NetworkUserId,
  StartTstampUTC
```

**Expected behavior**
Consecutive sessions that are only milliseconds apart should only occur very rarely.

**Screenshots**
Issue Query and Viz

Mostly Firefox 132, Desktop, but also Chromium Browser

**Additional context**
We deployed tracker version 4.0.1 on Monday for multiple hours, then rolled back due to [this issue](https://github.com/snowplow/snowplow-javascript-tracker/issues/1376)

The above described issue did not happen before the deployment of 4.0.1, but also happend in version 4.0.2. 

Options to isolate the issue:
- Just in case it is caused by the new async cookie writes: create a temporary tracker Initialization option to disable / enable async cookie writes
- A test which we could run in theory: set keepalive to false to make sure this is not the issue.

Unfortunately we had to revert tracker version V4 again, due to the increase in "artificial" sessions sessions.

PS: please test the next release on your own websites! Neither snowplow.io, nor discourse.snowplow.io, iglucentral.com or docs.snowplow.io is running Browser Tracker V4. console.snowplowanalytics.com is still on Version 4.0.0 (which has the event queue issue in Chromium).

matus-tomlein:
Issue #1381 

The issue reported that after upgrading to v4, there is a higher number of very short sessions in very short sequences. The hypothesis was that this is due to the async cookie access introduced in v4.

I was able to reproduce the problem with a simple setup that I added to the integration tests – a single page that embeds 15 iframe pages. The iframes are loaded at the same time and each initializes a new tracker and tracks a page view. Since the iframes share the same cookies, we should only see a single session id across all tracked page views.

With this setup, I was getting multiple different session ids across the page views. It was usually around 10 different session IDs.

The change I made was to flush the cookies (write them synchronously) each time the session changes. That means the first session or every subsequent new session. When a new event is tracked without triggering a new session, the cookies are still updated async by default.

This fixed the problem in the iframe test setup. Now I am only getting a single session ID across all page views. Have kept this test in the integration tests so that we don't regress on the behaviour.

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/issue-duplicate_sessions_2024-11-18-13-57.json
+++ b/common/changes/@snowplow/browser-tracker-core/issue-duplicate_sessions_2024-11-18-13-57.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
+      "comment": "Serialize new session cookie synchronously to avoid overlapping sessions (#1381)",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/browser-tracker-core"
+}
--- a/common/changes/@snowplow/javascript-tracker/issue-duplicate_sessions_2024-11-18-13-57.json
+++ b/common/changes/@snowplow/javascript-tracker/issue-duplicate_sessions_2024-11-18-13-57.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/javascript-tracker",
+      "comment": "Serialize new session cookie synchronously to avoid overlapping sessions (#1381)",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/javascript-tracker"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -2241,8 +2241,8 @@ importers:
         specifier: 4.1.2
         version: 4.1.2
       chromedriver:
-        specifier: ~129.0.0
-        version: 129.0.4
+        specifier: ~131.0.0
+        version: 131.0.0
       dockerode:
         specifier: ~3.3.1
         version: 3.3.5
@@ -2299,7 +2299,7 @@ importers:
         version: 4.6.4
       wdio-chromedriver-service:
         specifier: ~8.1.1
-        version: 8.1.1(@wdio/types@8.39.0)(chromedriver@129.0.4)(webdriverio@8.39.1(encoding@0.1.13)(typescript@4.6.4))
+        version: 8.1.1(@wdio/types@8.39.0)(chromedriver@131.0.0)(webdriverio@8.39.1(encoding@0.1.13)(typescript@4.6.4))
       wdio-edgedriver-service:
         specifier: ~3.0.3
         version: 3.0.3(@wdio/types@8.39.0)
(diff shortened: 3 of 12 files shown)


jethron (approved):
LGTM!

PR #1385, opened 2024-11-20
Fix multiple link click events tracked for a single tracker in case multiple trackers are initialized on the page (close #1384)

### LINKED ISSUE #1384: Multiple link click events tracked for a single tracker in case multiple trackers are initialized on the page
**Describe the bug**
In case multiple trackers are intialized on the same page, after enabling link click tracking, clicking on a link will track multiple events for each tracker.

For instance, when 3 trackers are initialized, each tracker will track 3 link click events when clicking on a link on the page.

**To Reproduce**
1. Call `newTracker` multiple times to initialize more trackers. 
2. Enable link click tracking.
3. Multiple link click events should be tracked for each tracker.

**Expected behavior**
Each tracker should track only one link click event.

matus-tomlein:
Issue #1384

Since the v4 release, in case multiple trackers were initialized on the same page, link click tracking tracked multiple events on each tracker for each link click.

For instance, in case 3 trackers were initialized and link click tracking was enabled, each link click would result in 3 events on each tracker.

This PR fixes the problem by qualifying which tracker should be used in the `trackLinkClick` function call from the click handler.

### DIFF
--- a/common/changes/@snowplow/browser-plugin-link-click-tracking/issue-1384-link_click_fix_2024-11-20-11-36.json
+++ b/common/changes/@snowplow/browser-plugin-link-click-tracking/issue-1384-link_click_fix_2024-11-20-11-36.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-plugin-link-click-tracking",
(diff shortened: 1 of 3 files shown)


jethron (approved):
Oops, my bad... :(

Good find, LGTM!

PR #1402, opened 2025-01-02
Add WebView plugin

mscwilson:
I've manually tested the plugin for iOS and React Native (iOS and Android).
Are there any other unit tests that would be good?

The dependency on the Webview tracker can be updated when the new version is released, it's currently a local dependency.

### DIFF
--- a/plugins/browser-plugin-webview/src/index.ts
+++ b/plugins/browser-plugin-webview/src/index.ts
@@ -5,10 +5,17 @@ import { base64urldecode } from './utils';
 
 const _trackers: Record<string, BrowserTracker> = {};
 
+const defaultTrackerNamespaces: string[] = [];
+
+type WebViewPluginOptions = {
+  trackerNamespaces?: string[];
+};
+
 /**
  * Forwards events to Snowplow mobile trackers running in a WebView.
+ * @param configuration - Configuration. Specify certain tracker namespaces to forward events to.
  */
-export function WebViewPlugin(): BrowserPlugin {
+export function WebViewPlugin(configuration?: WebViewPluginOptions): BrowserPlugin {
   let LOG: Logger;
 
   return {
@@ -43,7 +50,9 @@ export function WebViewPlugin(): BrowserPlugin {
         };
         let event = getSelfDescribingEventData(payload);
         let entities = getEntities(payload);
-        let trackers: Array<string> = Object.keys(_trackers);
+        let trackers: Array<string> = configuration?.trackerNamespaces
+          ? configuration?.trackerNamespaces
+          : defaultTrackerNamespaces;
 
         trackWebViewEvent({ properties: atomicProperties, event: event, context: entities }, trackers);
 
--- a/plugins/browser-plugin-webview/src/index.ts
+++ b/plugins/browser-plugin-webview/src/index.ts
@@ -35,11 +35,11 @@ export function WebViewPlugin(): BrowserPlugin {
           action: payload.se_ac as string | undefined,
           label: payload.se_la as string | undefined,
           property: payload.se_pr as string | undefined,
-          value: payload.se_va ? parseFloat(payload.se_va as string) : undefined,
-          minXOffset: payload.pp_mix ? parseInt(payload.pp_mix as string) : undefined,
-          maxXOffset: payload.pp_max ? parseInt(payload.pp_max as string) : undefined,
-          minYOffset: payload.pp_miy ? parseInt(payload.pp_miy as string) : undefined,
-          maxYOffset: payload.pp_may ? parseInt(payload.pp_may as string) : undefined,
+          value: payload.se_va !== undefined ? parseFloat(payload.se_va as string) : undefined,
+          minXOffset: payload.pp_mix !== undefined ? parseInt(payload.pp_mix as string) : undefined,
+          maxXOffset: payload.pp_max !== undefined ? parseInt(payload.pp_max as string) : undefined,
+          minYOffset: payload.pp_miy !== undefined ? parseInt(payload.pp_miy as string) : undefined,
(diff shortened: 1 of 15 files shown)


plugins/browser-plugin-webview/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+          useragent: payload.ua as string,
+          url: payload.url as string,
+          title: payload.page as string,
+          referrer: payload.refr as string,
+          category: payload.se_ca as string,
+          action: payload.se_ac as string,
+          label: payload.se_la as string,
+          property: payload.se_pr as string,
+          value: payload.se_va as number,
+          minXOffset: payload.pp_mix as number,
```
matus-tomlein, line 35: The values in the payload are string and type casting `as number` won't really change the type. I think you will need to use `parseInt` to convert strings to integers and `parseFloat` to convert strings to floats.

plugins/browser-plugin-webview/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+        LOG.debug(`Payload (event ID: ${payload.eid}) was filtered out and forwarded to WebView tracker.`);
+
+        let atomicProperties = {
+          eventName: payload.e as string,
+          trackerVersion: payload.tv as string,
+          useragent: payload.ua as string,
+          url: payload.url as string,
+          title: payload.page as string,
+          referrer: payload.refr as string,
+          category: payload.se_ca as string,
```
matus-tomlein, line 30: Some of these properties are optional so it's not completely correct to always cast to `string`. Rather it would be more correct to cast `as string | undefined`. It doesn't really change anything in practice but at least the types will be more representative in Typescript.

plugins/browser-plugin-webview/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+      }
+    },
+  };
+}
+
+function getSelfDescribingEventData(payload: Payload): SelfDescribingEvent | undefined {
+  if (payload.ue_pr) {
+    return JSON.parse(payload.ue_pr as string);
+  } else if (payload.ue_px) {
+    return JSON.parse(payload.ue_px as string);
```
matus-tomlein, line 58: You can't just parse a base64 encoded string to JSON directly, you will first need to decode it. Unfortunately the function is not exported but there is a `base64urldecode` function in the core library that you could copy ([see here](https://github.com/snowplow/snowplow-javascript-tracker/blob/743e77df8bf914682017e350469b99c22207505f/libraries/tracker-core/src/payload.ts#L189))


The same in `getEntities`. I think this is worth a unit test.

plugins/browser-plugin-webview/test/webview.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      customFetch,
+    });
+
+    tracker?.trackPageView();
+
+    let events = await eventStore.getAllPayloads();
+    expect(events).toHaveLength(1);
+  });
+
+  it('Filters out the events if a mobile interface is present', async () => {
```
matus-tomlein, line 34: We will need to add tests that check that the `trackWebViewEvent` was called with the correct parameters from the event.

plugins/browser-plugin-webview/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+
+    filter: (payload) => {
+      if (hasMobileInterface() === true) {
+        LOG.debug(`Payload (event ID: ${payload.eid}) was filtered out and forwarded to WebView tracker.`);
+
+        let atomicProperties = {
+          eventName: payload.e as string,
+          trackerVersion: payload.tv as string,
+          useragent: payload.ua as string,
```
matus-tomlein, line 26: I think by default we don't set the `ua` property in events on the JS tracker since that is parsed from the HTTP headers by collector. However, since in the WebView tracker, we won't be making the request directly from the browser, we will need to pass them manually.

So I'd suggest that we add a default value here – in case `payload.ua` is not present, we should read the useragent from `window.navigator.userAgent`.

plugins/browser-plugin-webview/src/index.ts
```diff
@@ -0,0 +1,76 @@
...
+          minXOffset: payload.pp_mix ? parseInt(payload.pp_mix as string) : undefined,
+          maxXOffset: payload.pp_max ? parseInt(payload.pp_max as string) : undefined,
+          minYOffset: payload.pp_miy ? parseInt(payload.pp_miy as string) : undefined,
+          maxYOffset: payload.pp_may ? parseInt(payload.pp_may as string) : undefined,
+        };
+        let event = getSelfDescribingEventData(payload);
+        let entities = getEntities(payload);
+        let trackers: Array<string> = Object.keys(_trackers);
+
+        trackWebViewEvent({ properties: atomicProperties, event: event, context: entities }, trackers);
```
matus-tomlein, line 48: Can we make the last `trackers` argument optional and by default omit it?

I think it's quite unlikely (and probably not a good design) that users will use the same tracker namespace for their JS tracker and their mobile trackers. So forcing the JS tracker namespace here is probably not a good idea.

I would suggest making it configurable:
1. by default, we wouldn't specify the list of trackers so that the events will be tracked with any trackers
2. in the `WebViewPlugin` function we could add a configuration object that would let users provide the list of tracker namespaces to use.

matus-tomlein (commented):
Sorry for reviewing this late, hope it will be possible for you to get some time to check the last few comments, but in case not, let me know and I'll try to help here.

It's working very nicely, but I think we need to change the default of requiring the same tracker namespace on Web as on mobile trackers.

Also after merging the 4.2 branch into master, I kind of messed up your commits 🙈 Can you cherry pick them on top of the latest master?

matus-tomlein (approved):
LGTM!

PR #1404, opened 2025-01-07
Round resolution in mobile context to avoid validation failures on Android

matus-tomlein:
Fixes a bug identified in https://github.com/snowplow-incubator/snowplow-javascript-tracker-examples/pull/60 where the mobile_context entity failed validation due to the resolution field being too long.

The change converts the width and height to integer by taking the floor of the values.


mscwilson (approved):
would be good to update the Android test too!

PR #1407, opened 2025-01-22
Allow for usage of `snowplow` callback without `this`

greg-el:
This PR allows for the `snowplow(function() {})` callback to get tracker information without requiring the `this` keyword, which is useful in environments that don't support the keyword _*(GTM)*_ and is a non-breaking change.

The callback has been updated to pass the tracker dictionary (the object that is referred to by `this`) as the last argument. This allows for it to be used as a simple replacement for `this`, like so:
```js
snowplow(function (trackers) {
    var sp = trackers.sp;
    var domainUserId = sp.getDomainUserId();
});
```

As it is the last parameter, it still allows for the passing of arguments in the `snowplow` function:
```js
snowplow(function(a, b, trackers) {
    console.log(a, b); // 1, 2
    var sp = trackers.sp;
    var domainUserId = sp.getDomainUserId();
}, 1, 2);
```

The `this` keyword is still supported in conjunction with the parameter:
```js
snowplow(function(trackers) {
  trackers.sp.getDomainUserId() === this.sp.getDomainUserId() // true
});

### DIFF
--- a/common/changes/@snowplow/javascript-tracker/issue-1085-snowplow-callback-without-this_2025-01-22-17-05.json
+++ b/common/changes/@snowplow/javascript-tracker/issue-1085-snowplow-callback-without-this_2025-01-22-17-05.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/javascript-tracker",
+      "comment": "Allow usage of Snowplow callback without 'this' keyword",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/javascript-tracker"
+}
--- a/trackers/javascript-tracker/src/in_queue.ts
+++ b/trackers/javascript-tracker/src/in_queue.ts
@@ -265,7 +265,15 @@ export function InQueueManager(functionName: string, asyncQueue: Array<unknown>)
             // Strip GlobalSnowplowNamespace from ID
             fnTrackers[tracker.id.replace(`${functionName}_`, '')] = tracker;
           }
-          input.apply(fnTrackers, parameterArray);
+
+          // Create a new array from `parameterArray` to avoid mutating the original
+          const parameterArrayCopy = Array.prototype.slice.call(parameterArray);
+
+          // Add the `fnTrackers` object as the last element of the new array to allow it to be accessed in the callback
+          // as the final argument, useful in environments that don't support `this` (GTM)
+          const args = Array.prototype.concat.apply(parameterArrayCopy, [fnTrackers]);
+
+          input.apply(fnTrackers, args);
         } catch (ex) {
           LOG.error('Tracker callback failed', ex);
         } finally {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: trackers/javascript-tracker/test/unit/in_queue.test.ts [hunk 1]
```diff
@@ -183,4 +211,58 @@ describe('InQueueManager', () => {
...
+      function (this: any, a: number, b: number, trackers: Record<string, any>) {
+        expect(this.sp1).toEqual(mockTrackers.callback_sp1);
+        expect(a).toEqual(1);
+        expect(b).toEqual(2);
+        expect(trackers.sp1).toEqual(mockTrackers.callback_sp1);
+      },
+      1,
+      2,
+    ]);
+  });
```
jethron, line 267: I think you could get by with just 1/3/5 of these; 2 & 4 seem redundant. (1 & 3 are also replicated by 5 but show different use cases that are kind valuable to have as exposition).

common/changes/@snowplow/javascript-tracker/issue-1085-snowplow-callback-without-this_2025-01-22-17-05.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/javascript-tracker",
+      "comment": "Allow usage of Snowplow callback without 'this' keyword",
+      "type": "none"
```
jethron, line 6: I'd say this is _at least_ `minor`.

jethron (commented):
LGTM if LGT @matus-tomlein

trackers/javascript-tracker/src/in_queue.ts
```diff
@@ -265,6 +265,11 @@ export function InQueueManager(functionName: string, asyncQueue: Array<unknown>)
             // Strip GlobalSnowplowNamespace from ID
             fnTrackers[tracker.id.replace(`${functionName}_`, '')] = tracker;
           }
+
+          // Append the `fnTrackers` object to the parameterArray, to allow it to be
+          // accessed as the final argument in the callback, useful in environments that don't support `this` (GTM)
+          Array.prototype.push.call(parameterArray, fnTrackers);
```
matus-tomlein, line 271: Nitpick, but might be safer to create a new concatenated array (`parameterArray + [fnTrackers]`) rather than mutating `parameterArray`? I see `parameterArray` being used below although likely not in the case this is function, but I guess better to be safe.

matus-tomlein (approved):
LGTM!

common/changes/@snowplow/javascript-tracker/issue-1085-snowplow-callback-without-this_2025-01-22-17-05.json
greg-el: Good point, I've updated the base branch to `release/4.4.0`.

trackers/javascript-tracker/src/in_queue.ts
greg-el: Fine by me! Had to do it in a bit less concise way than that though 😄

PR #1408, opened 2025-01-23
Allow for usage of `snowplow` callback without `this` (close #1085)

### LINKED ISSUE #1085: Consider alternatives to get information from JavaScript tracker
At the moment the way to get information (`getDomainUserInfo`, `getUserId` etc) from a JavaScript tracker is through callbacks. Tracker callbacks are called with the [trackers dictionary as context](https://github.com/snowplow/snowplow-javascript-tracker/blob/a608ead2e99bcb399a1763376619eaba55fdf9c5/trackers/javascript-tracker/src/in_queue.ts#L268). That way the callback can access the trackers dictionary using `this`.
However, `this` cannot be used when the JavaScript tracker is loaded in GTM:

 - in tag template code `this` is a reserved keyword
 - in custom HTML tags and in JS variables, `this` is not the trackers dictionary inside the callback

A way to get information from a JavaScript tracker that is independent of `this` and possibly without providing access to trackers dictionary. 
Even though this is currently a problem only when the JavaScript tracker is injected through this specific environment (GTM), providing an alternative way to get information from a JavaScript tracker may make sense in a wider scope.

greg-el:
This PR backports the patch in https://github.com/snowplow/snowplow-javascript-tracker/commit/6f87b5a429580e760f05fd08534ae0d647ad2353 to v3 of the tracker.

### DIFF
--- a/common/changes/@snowplow/javascript-tracker/issue-1085-snowplow-callback-without-this_2025-01-22-17-05.json
+++ b/common/changes/@snowplow/javascript-tracker/issue-1085-snowplow-callback-without-this_2025-01-22-17-05.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/javascript-tracker",
+      "comment": "Allow usage of Snowplow callback without 'this' keyword",
+      "type": "none"
+    }
+  ],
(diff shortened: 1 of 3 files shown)


matus-tomlein (approved):
LGTM!

PR #1412, opened 2025-03-06
Bump web-vitals version

### LINKED ISSUE #1411: Upgrade web-vitals to v4

The version of web-vitals used by this library (3.3.2) is quite old (May 2023) and unsupported. The latest version is 4.2.4 and we're even working on a v5.

v4 includes a number of improvements, particular for INP tracking.

It might be a good idea to upgrade to v4.2.4 now to get the latest, and minimise issues when upgrading to v5 later.

[FID has also been deprecated](https://web.dev/blog/fid) and replaced by INP (already being tracked). So might be good idea to remove that since it it's deprecated in v4 and removed in v5.

Leave as is, or wait for v5. Or upgraxde to v4, but still collect FID for now.

**Additional context**

I'm the maintainer of web-vitals so happy to answer any questions you might have!

I'm also happy to add a PR for this as looks fairly simple. However, there are some [breaking changes in v4](https://github.com/GoogleChrome/web-vitals/blob/main/docs/upgrading-to-v4.md), which may affect users of this so not sure what the best way is of communicating this (of it it needs to wait for a major upgrade of Snowplow?)

jethron:
From ~3 to ~4. Also make the metric interfaces optional in preparation for v5, which will remove the (deprecated in v4) `onFID` method.

The FID metric is already optional in the [schema](https://iglucentral.com/?q=web_vitals), so there should be no breaking change here when this metric becomes unavailable.

We assume if the user is using the library themselves they would be manually loading it and telling us not to via:

- `loadVitalsScript: false`
- `webVitalsSource` override

So I don't consider this a breaking change unless a user is relying on us to load v3 and with this update we start loading v4 (which has some [breaking changes](https://github.com/GoogleChrome/web-vitals/blob/main/docs/upgrading-to-v4.md)). Users in this situation can set `webVitalsSource` manually to continue loading v3 until they can support v4.

When v5 is released, users can opt-in to using it by specifying a v5 source for `webVitalsSource`, which should work as expected with these changes until we update the plugin's default source to be v5.

v4 [removes](https://github.com/GoogleChrome/web-vitals/commit/538cdf59be90c78a9da5c54baa471860dae61477) the IIFE-version of the library's global var type definitions (`webVitals: WebVitalsGlobal`), so we now declare it ourselves.

Closes #1411

### DIFF
--- a/common/changes/@snowplow/browser-plugin-web-vitals/feat-web-vitals-bump_2025-03-06-01-39.json
+++ b/common/changes/@snowplow/browser-plugin-web-vitals/feat-web-vitals-bump_2025-03-06-01-39.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-plugin-web-vitals",
+      "comment": "Update default external library version to v4. Add compatibility for future v5, which deprecates the FID metric.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/browser-plugin-web-vitals"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -1950,8 +1950,8 @@ importers:
         specifier: ^2.3.1
         version: 2.8.1
       web-vitals:
-        specifier: ~3.3.2
-        version: 3.3.2
+        specifier: ~4.2.4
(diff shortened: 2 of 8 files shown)


igneel64 (approved):

tunetheweb:
Thank you so much for this! Let us know if you run into any problems.

PR #1415, opened 2025-03-20
Release/4.4.0

greg-el:
- Bump web-vitals version
- Allow for usage of `snowplow` callback without `this`
- Add react-native-get-random-values as a peer dependency in the React Native tracker


matus-tomlein (approved):
LGTM

Issue #1421, opened 2025-04-01
Add empty string to `button_click` event


Jack-Keene:
Many customers end up with lots of failed events related to the button_click schema because ‘label’ property is marked as required. However, many buttons don’t contain any inner text as they use icons or svgs instead. For example, nav bars, exit buttons, etc.

In these cases, all of these button_click events fail validation.

There is a workaround for the button_click tracking to attach [‘data-sp-button-label’](https://docs.snowplow.io/docs/sources/trackers/javascript-trackers/web-tracker/tracking-events/button-click/#overriding-the-label-with-data-sp-button-label) attribute to the button element. But this would require a lot of work on the engineering team to update all buttons with relevant labels.

It seems to be the better approach would be to add a default empty label and then they at least can capture the click event even if it doesn’t have a label.

Issue #1413, opened 2025-03-07
Allow configuring a custom event store in React Native tracker


matus-tomlein:
> After having reasoned a bit on it, I have an absurd idea. What if the React Native tracker accepts a storage interface in `newTracker`? This would allow consumers to define other storage instead of defaulting to `AsyncStorage`. For example, MMKV is a popular alternative in the React Native community that could be used alternatively. 

 _Originally posted by @valeriobelli in [#1409](https://github.com/snowplow/snowplow-javascript-tracker/issues/1409#issuecomment-2671634775)_

PR #1422, opened 2025-04-03
Release 4.5.0

### LINKED ISSUE #1397: disableButtonClickTracking disables tracking of buttons for all trackers
**Describe the bug**
`disableButtonClickTracking` disables tracking of buttons for all trackers, and it doesn't allow to stop button click tracking only for the specified trackers.

**To Reproduce**
Steps to reproduce the behavior or code snippets that produce the issue:
1. Create two seperate trackers.
1. Call `enableButtonClickTracking` with specific trackers, e.g. `trackerA` and `trackerB`.
2. Call [disableButtonClickTracking](https://github.com/snowplow/snowplow-javascript-tracker/blob/743e77df8bf914682017e350469b99c22207505f/plugins/browser-plugin-button-click-tracking/src/api.ts#L83).
3. Observe that button click tracking is disabled for all trackers.

**Expected behavior**
`disableButtonClickTracking` should allow disabling button click tracking only for the specified trackers.

 - OS: [e.g. macOS] n/a
 - Browser [e.g. Chrome, Safari] n/a
 - Version [e.g. 22] @snowplow/browser-plugin-button-click-tracking 3.24.3

**Additional context**
We still use tracker v3 as there are some breaking changes in v4 for us, and we need this fixed in v3 as well.

matus-tomlein:
This release avoids the error `ResiezObserver loop completed with undelivered notifications` being generated under certain circumstances.

It also makes it possible to provide a custom event store implementation in the React Native tracker and use a different provider than `@react-native-async-storage/async-storage`.

Finally, the release adds a configurable default label (`(empty)`) for automatic button click tracking to prevent validation failures for buttons that are missing a label. It also fixes disabling button click tracking for specific trackers.

NOTE: The `@react-native-async-storage/async-storage` dependency has been moved to peer dependencies for the React Native tracker. We expect that users already had this dependency listed under the app dependencies due to it providing native bindings that needed to be installed separately. Make sure your app has this dependency when upgrading the tracker.

**Enhancements**

- Allow configuring a custom event store in React Native tracker (#1413) thanks to @valeriobelli
- Add a configurable default label in button click tracking and fix disabling for specific trackers (close #1397 and #1421)

**Bug fixes**

- Avoid undelivered notifications error (#1335) thanks to @AngusMorton


xavi-999:
Hey @matus-tomlein, thanks for the MMKV support on this release!
Does the `@react-native-async-storage/async-storage` dependency still need to be installed if the client is using a custom event store such as MMKV?

valeriobelli:
> Hey @matus-tomlein, thanks for the MMKV support on this release! Does the `@react-native-async-storage/async-storage` dependency still need to be installed if the client is using a custom event store such as MMKV?

Hey @xavi-999 👋🏽 
Yes, it will still be needed to satisfy the peer dependency requirement. AsyncStorage will be required during bundling, given it is declared in the import graph.

As far as I know, the complete removal of AsyncStorage might be scheduled for v5, but @matus-tomlein can answer better here.

xavi-999:
> Hey @xavi-999 👋🏽
> Yes, it will still be needed to satisfy the peer dependency requirement. AsyncStorage will be required during bundling, given it is declared in the import graph.
> 
> As far as I know, the complete removal of AsyncStorage might be scheduled for v5, but @matus-tomlein can answer better here.

Hey :) 

Makes sense, I will be waiting for v5 then. Great job by the way @valeriobelli. Regards

PR #1423, opened 2025-04-04
button-click-tracking fixes

jethron:
- Makes it possible to disable button click tracking for specific trackers (see #1397 but this will only fix for v4)
- Fixes an issue where enabling button click tracking for specific trackers would disable it for all other trackers
- Add a default `label` (`(empty)`) for button_click events where one can not be determined by the button's content or data attributes (see #1421)
- Make the default label configurable (static string or via callback for the element)
- Fixes an integration test that wasn't running properly and so wasn't actually testing the right thing

Technically a breaking change, but only in the sense that events that were previously bad will not be any more. A fixing change? So aiming for 4.5.0 since a minor release seems appropriate.

### DIFF
--- a/plugins/browser-plugin-button-click-tracking/src/types.ts
+++ b/plugins/browser-plugin-button-click-tracking/src/types.ts
@@ -18,6 +18,8 @@ export interface ButtonClickTrackingConfiguration {
   filter?: Filter;
   /** The dynamic context which will be evaluated for each button click event */
   context?: DynamicContext;
+  /** A default label to use if one can not be determined by the content or data attributes */
+  defaultLabel?: string | ((element: HTMLElement) => string);
 }
 
 /**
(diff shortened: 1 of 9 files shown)


plugins/browser-plugin-button-click-tracking/src/types.ts
```diff
@@ -18,6 +18,8 @@ export interface ButtonClickTrackingConfiguration {
   filter?: Filter;
   /** The dynamic context which will be evaluated for each button click event */
   context?: DynamicContext;
+  /** A default label to use if one can not be determined by the content or data attributes */
```
matus-tomlein, line 21: If I understand correctly, if this is not specified, the default will be `(empty)`, should we note this here or somewhere else?

matus-tomlein (approved):
LGTM! 👍 Happy to include this in the 4.5 release

plugins/browser-plugin-button-click-tracking/src/types.ts
matus-tomlein: Maybe this will be more visible in the tracker docs.

PR #1428, opened 2025-05-05
Add a default 0 value for start and end in the media_element entity time ranges

matus-tomlein:
This PR aims to prevent a schema validation error occurs for some mobile browsers which seem to be omitting the `end` value in timeranges within the `seekable` [property in the `media_element` entity schema](https://github.com/snowplow/iglu-central/blob/master/schemas/org.whatwg/media_element/jsonschema/1-0-0#L170).

I couldn't reproduce this unfortunately but assume it is some type of browser bug, so the best I can think of doing is to add a default 0 value to the `start` and `end` properties within the time ranges.

If you have some better ideas, please let me know!

The majority of failed events for the customer have the following value for the seekable property:

``` 
"seekable": [
  {
    "start": 0
  }
]
```

### DIFF
--- a/plugins/browser-plugin-media-tracking/src/helperFunctions.ts
+++ b/plugins/browser-plugin-media-tracking/src/helperFunctions.ts
@@ -53,7 +53,7 @@ type TimeRange = { start: number; end: number };
 export function timeRangesToObjectArray(t: TimeRanges): TimeRange[] {
   const out: TimeRange[] = [];
   for (let i = 0; i < t.length; i++) {
-    out.push({ start: t.start(i), end: t.end(i) });
+    out.push({ start: t.start(i) || 0, end: t.end(i) || 0 });
   }
   return out;
 }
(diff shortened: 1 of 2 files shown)


plugins/browser-plugin-media-tracking/src/helperFunctions.ts
```diff
@@ -53,7 +53,7 @@ type TimeRange = { start: number; end: number };
 export function timeRangesToObjectArray(t: TimeRanges): TimeRange[] {
   const out: TimeRange[] = [];
   for (let i = 0; i < t.length; i++) {
-    out.push({ start: t.start(i), end: t.end(i) });
+    out.push({ start: t.start(i) || 0, end: t.end(i) || 0 });
```
jethron, line 56: Should we also handle the [`Infinity` case](https://html.spec.whatwg.org/multipage/media.html#seeking:dom-media-duration) for `end`, which would make it `null` and also invalid?

> whose `end` is the same as the time of the first frame plus the `duration` attribute's value (which would equal the time of the last frame, and might be positive `Infinity`)

jethron (approved):
LGTM!

PR #1430, opened 2025-05-16
Fix Element Tracking element_content entity properties

jethron:
In preparation for publishing the Element Tracking schemas to Iglu Central, changes were made to the `element_content` schema to be more consistent with the other related schemas.

The type definition in the plugin implementation was updated to match, but the function that builds the context with that entity also includes the entities from the user's custom entity generators, so accepts arbitrary `SelfDescribingJson` to build a context of type `(ElementContentEntity | SelfDescribingJson)[]`. Since the implementation for that entity was not updated, it no longer satisfied `ElementContentEntity` but _did_ still satisfy `SelfDescribingJson`, so passed the type check and compiled successfully, despite no longer matching the schema.

This builds the entity in a type-safer way, declaring it explicitly as `ElementContentEntity`, and fixes the implementation to match the schema.

### DIFF
--- a/common/changes/@snowplow/browser-plugin-element-tracking/element-tracker-schemafix_2025-05-16-01-10.json
+++ b/common/changes/@snowplow/browser-plugin-element-tracking/element-tracker-schemafix_2025-05-16-01-10.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-plugin-element-tracking",
+      "comment": "Fix element_content schema implementation.",
+      "type": "none"
+    }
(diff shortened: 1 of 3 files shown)


matus-tomlein (approved):
LGTM!

matus-tomlein:
I was going to make a release but got confused because this commit already seems to be in the `release/4.6.2` [branch](https://github.com/snowplow/snowplow-javascript-tracker/commit/f99816f8fa25221a18222d26e490c09f40f00b14)?

jethron:
Sorry @matus-tomlein, fixed!

PR #1433, opened 2025-05-23
Fix session resumption when changing storage strategy with disableAnonymousTracking

jethron:
If you create a tracker in anonymousTracking mode with session tracking enabled, and then change the storage strategy when disabling anonymousTracking mode, the tracker looks for the `ses` cookie using the new strategy rather than the old one, which fails. This makes it think a new session has begun. In order to tie the anonymous behaviour to the non-anonymous behaviour, you have to stitch via pageview ID or network user ID rather than the expected session ID.

If this is the initial pageview, this can cause interesting cases like events from a single pageview with two different `domain_sessionid` values, both with `domain_sessionidx` = 1.

With this change, we detect when a storage strategy is being changed, and if we are already in a session with the old strategy, update the `ses` cookie with the new strategy before loading the ID. This allows the tracker to notice a session is already in progress with the new strategy, and resume it with the existing session ID/index.

If there is no change to the storage strategy, or there is currently no session in progress, the behavior remains unchanged.

E.g.

```javascript

snowplow('newTracker', 'sp', 'http://localhost:9090', {
    anonymousTracking: {
      withSessionTracking: true,
      withServerAnonymisation: true
    },
    stateStorageStrategy: 'localStorage',
});

snowplow('trackPageView'); // session ID 1

snowplow('disableAnonymousTracking', {stateStorageStrategy: 'cookieAndLocalStorage'}); // got consent, switch to cookies

snowplow('trackStructEvent', { // session ID 2 - oops!
  category: 'Mixes',
  action: 'Play',
  label: 'MrC/fabric-0503-mix',
  property: '',
  value: 0.0
});
```

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/issue-bcpf-1813-stratchange-resume_2025-05-23-06-07.json
+++ b/common/changes/@snowplow/browser-tracker-core/issue-bcpf-1813-stratchange-resume_2025-05-23-06-07.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
+      "comment": "Resume sessions when disabling anonymous tracking that was already tracking session state",
+      "type": "none"
+    }
+  ],
+  "packageName": "@snowplow/browser-tracker-core"
+}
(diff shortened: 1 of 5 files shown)


matus-tomlein (approved):
LGTM!

PR #1434, opened 2025-05-27
Release 4.6.3

matus-tomlein:
**Bug fixes**

- Fix session not continuing when disableAnonymousTracking changes storage strategy (#1433)

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/issue-bcpf-1813-stratchange-resume_2025-05-23-06-07.json
+++ b/common/changes/@snowplow/browser-tracker-core/issue-bcpf-1813-stratchange-resume_2025-05-23-06-07.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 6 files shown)


jethron:
Hadn't notice the `build` CI fails so much before. Looks like the earliest version of Safari SauceLabs supports nowadays is 14. Perhaps we should remove [the 12.1 request](https://github.com/snowplow/snowplow-javascript-tracker/blob/f0ed85f179f2b31f306aebb174dcb4115ec527ea/trackers/javascript-tracker/test/wdio.sauce.conf.ts#L67-L74)?

[WBM says](https://web.archive.org/web/20250106045608/https://api.us-west-1.saucelabs.com/v1/signup/api/platforms) it was still around in January, but it's [gone now](https://api.us-west-1.saucelabs.com/v1/signup/api/platforms).

matus-tomlein:
Great call, thanks @jethron! Still need to fix the PR check, but will leave that for next PR.

PR #1435, opened 2025-05-28
Fixing GA4 Add To Cart for GTM Ecommerce Template

jborlase-snowplow:
The current implementation of the trackGA4AddToCart is incompatible with the Snowplow Ecommerce tag template. The tag template combines the opts and ecommerce objects. There is no method to pass multiple parameters through the GTM template. The result is that the add to cart and remove to cart events are always invalid since they require a total_value property that comes from the finalCartValue.

This fix allows the JavaScript tracker to check the ecommerce object for the finalCartValue.


greg-el (approved):
Nice one! LGTM!

PR #1441, opened 2025-06-25
Release/4.6.5

matus-tomlein:
This patch release truncates the language property in the platform context entity on the React Native tracker to avoid validation errors with the entity schema.

**Bug fixes**

- Truncate language in platform context entity to max 8 characters


jethron (approved):
LGTM

PR #1445, opened 2025-08-15
Unify browser dimension checks

jethron:
Should address issues where NaN values are included in generated values for document/viewport/screen size dimensions, which cause events to be invalid when they fail to match the regex pattern that validates these fields.

This has mostly been observed in the viewport size, (`vp` parameter) which doesn't handle this case like document size does. This change extends the NaN checks from document size to all three dimensions to drop their parameter if a valid value can't be found.

Issue: AISP-527

### DIFF
--- a/libraries/browser-tracker-core/test/browser_props.test.ts
+++ b/libraries/browser-tracker-core/test/browser_props.test.ts
@@ -1,14 +1,18 @@
-import { floorDimensionFields, getBrowserProperties } from '../src/helpers/browser_props';
+import { makeDimension, getBrowserProperties } from '../src/helpers/browser_props';
 
 describe('Browser props', () => {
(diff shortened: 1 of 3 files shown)


libraries/browser-tracker-core/test/browser_props.test.ts
```diff
@@ -1,14 +1,17 @@
-import { floorDimensionFields, getBrowserProperties } from '../src/helpers/browser_props';
+import { makeDimension, getBrowserProperties } from '../src/helpers/browser_props';
 
 describe('Browser props', () => {
-  it('floorDimensionFields correctly floors dimension type values', () => {
+  it('makeDimension correctly floors dimension type values', () => {
```
igneel64, line 4: Since the case we fixed was about none I would just add one more test with NaN values

igneel64 (approved):

PR #1460, opened 2026-03-17
Add browser-plugin-bot-detection wrapping FingerprintJS BotD

stanch:
Introduces a new browser tracker plugin that detects bots client-side using @fingerprintjs/botd and attaches the result as a client_side_bot_detection context entity to all tracked events.

### DIFF
--- a/common/changes/@snowplow/browser-plugin-bot-detection/bot-detection-plugin_2026-03-17-17-11.json
+++ b/common/changes/@snowplow/browser-plugin-bot-detection/bot-detection-plugin_2026-03-17-17-11.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "Add browser-plugin-bot-detection plugin wrapping FingerprintJS BotD for client-side bot detection",
+      "type": "none",
+      "packageName": "@snowplow/browser-plugin-bot-detection"
+    }
+  ],
+  "packageName": "@snowplow/browser-plugin-bot-detection",
+  "email": "nick.stanch@snowplowanalytics.com"
+}
--- a/common/changes/@snowplow/browser-plugin-bot-detection/bot-detection-plugin_2026-03-17-17-11.json
+++ b/common/changes/@snowplow/browser-plugin-bot-detection/bot-detection-plugin_2026-03-17-17-11.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "comment": "Add browser-plugin-bot-detection plugin wrapping FingerprintJS BotD for client-side bot detection",
-      "type": "none",
+      "type": "minor",
       "packageName": "@snowplow/browser-plugin-bot-detection"
     }
   ],
--- a/plugins/browser-plugin-bot-detection/package.json
+++ b/plugins/browser-plugin-bot-detection/package.json
@@ -1,0 +1,53 @@
+{
+  "name": "@snowplow/browser-plugin-bot-detection",
+  "version": "4.6.9",
+  "description": "Detects bots client-side using FingerprintJS BotD and attaches the result as a context entity.",
+  "homepage": "https://docs.snowplow.io/",
+  "bugs": "https://github.com/snowplow/snowplow-javascript-tracker/issues",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/snowplow/snowplow-javascript-tracker.git"
+  },
+  "license": "BSD-3-Clause",
+  "sideEffects": false,
+  "main": "./dist/index.umd.js",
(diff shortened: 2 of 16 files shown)


stanch:
FWIW, botd has no dependencies, so I think it’s not what upset Wiz...

plugins/browser-plugin-bot-detection/src/types.ts
```diff
@@ -0,0 +1,25 @@
+export type BotKind =
+  | 'awesomium'
```
matus-tomlein, line 2: `BotKind` is manually duplicated from `@fingerprintjs/botd`'s internal type list. When botd adds new bot signatures in a future release, they'll silently fall outside this union and won't be reflected in the context data.

Consider importing the type directly from botd instead:

```typescript
export type { BotKind } from '@fingerprintjs/botd';
```

This keeps the type in sync automatically without any maintenance overhead.

common/changes/@snowplow/browser-plugin-bot-detection/bot-detection-plugin_2026-03-17-17-11.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "Add browser-plugin-bot-detection plugin wrapping FingerprintJS BotD for client-side bot detection",
+      "type": "none",
```
matus-tomlein: The change type is `"none"`, which means this new package won't get a version bump or changelog entry on release. For a brand-new published package this should be `"minor"` to signal it as a new feature:

```json
"type": "minor"
```

plugins/browser-plugin-bot-detection/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "version": "4.6.9",
+  "description": "Detects bots client-side using FingerprintJS BotD and attaches the result as a context entity.",
+  "homepage": "https://docs.snowplow.io/",
+  "bugs": "https://github.com/snowplow/snowplow-javascript-tracker/issues",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/snowplow/snowplow-javascript-tracker.git"
+  },
+  "license": "BSD-3-Clause",
+  "sideEffects": false,
```
matus-tomlein, line 13: Missing `"author"` field — all other plugins include one. Please add:

```json
"author": "Snowplow Analytics Ltd (https://snowplow.io/)"
```

plugins/browser-plugin-bot-detection/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "main": "./dist/index.umd.js",
+  "module": "./dist/index.module.js",
+  "types": "./dist/index.module.d.ts",
+  "files": [
+    "dist"
```
matus-tomlein, line 18: This plugin is missing `README.md` and `LICENSE` files that every other plugin in the repo includes. Please add them before merging.

plugins/browser-plugin-bot-detection/test/bot-detection.test.ts
```diff
@@ -0,0 +1,149 @@
...
+
+jest.mock('@fingerprintjs/botd', () => ({
+  load: () => {
+    if (mockLoadReject) {
+      return Promise.reject(mockLoadReject);
+    }
+    return Promise.resolve({
+      detect: () => {
+        if (mockDetectReject) {
+          throw mockDetectReject;
```
matus-tomlein: The mock `detect()` throws synchronously here, but the real botd `detect()` returns a `Promise<BotDetectorResult>` — errors arrive as rejections, not throws.

In the current implementation:
```typescript
.then((detector) => detector.detect())
```
…a synchronous throw from inside `.then()` is caught by `.catch()`, so the test passes. But the mock doesn't accurately represent the real API contract.

For a more faithful test, return a rejected promise instead:
```typescript
detect: () => {
  if (mockDetectReject) {
    return Promise.reject(mockDetectReject);
  }
  return Promise.resolve(mockDetectResult);
}
```

matus-tomlein:
Just some small nitpicks but I think this looks good!

plugins/browser-plugin-bot-detection/src/types.ts
stanch: This one replicates the schema, so that if we accidentally bump the dependency (it’s pinned) and it brings new types, this code alerts us to that fact. (Related discussion in https://github.com/snowplow/iglu-central/pull/1470)

matus-tomlein (approved):
LGTM!

Can you please point this at the `release/4.7.0` branch so that we can prepare a release?

PR #1467, opened 2026-04-06
Add opt-in extended activity tracking with activity_metrics entity

stanch:
Attach quantitative activity metrics (mouse distance, scroll distance, key presses, clicks, touches) to page_ping events when extendedActivityTracking is enabled. Reuses existing DOM event handlers with zero overhead when the feature is off.


api-docs/docs/browser-tracker/browser-tracker.api.md
```diff
@@ -15,10 +15,21 @@ export type ActivityCallbackData = {
...
+    mouseDistance: number;
+    scrollDistance: number;
+    keyPresses: number;
+    clicks: number;
+    touches: number;
 };
 
 // @public
 export interface ActivityTrackingConfiguration {
+    extendedActivityTracking?: boolean;
```
matus-tomlein: Don't have a strong opinion but if the entity is called `activity_metrics`, can't we just use that term here as well rather than calling it extended activity tracking?

stanch: something like `enableActivityMetrics`, or just `activityMetrics`? it kinda bothered me also, so I am happy to change

matus-tomlein: `activityMetrics` sounds better to me.

libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -264,6 +271,13 @@ export function Tracker(
       maxXOffset: number,
       minYOffset: number,
       maxYOffset: number,
+      // Extended activity tracking
+      extendedActivityTrackingEnabled = false,
+      extMetrics: ActivityMetrics = { mouseDistance: 0, scrollDistance: 0, keyPresses: 0, clicks: 0, touches: 0 },
+      extLastMouseX: number | undefined = undefined,
+      extLastMouseY: number | undefined = undefined,
+      extLastScrollX: number | undefined = undefined,
+      extLastScrollY: number | undefined = undefined,
```
matus-tomlein: Nitpick, but this might be easier to manage if these 6 state vars could be grouped together?

```ts
let extActivity: {
  enabled: boolean;
  metrics: ActivityMetrics;
  lastMouseX?: number; lastMouseY?: number;
  lastScrollX?: number; lastScrollY?: number;
} | null = null;
```

matus-tomlein:
This looks good, but could you please add something closer to an integration test in `tracker.test.ts` that would also check the entity being tracked?

Here is something that AI suggested, but whatever gets us closer to an integration test (probably extending the e2e browser tests we have would be too much complicated):

```ts
it('attaches activity_metrics entity to page ping payload when using callback with extendedActivityTracking', async () => {
  const ACTIVITY_METRICS_SCHEMA = 'iglu:com.snowplowanalytics.snowplow/activity_metrics/jsonschema/1-0-0';
  const state = new SharedState();
  const t =
    addTracker('sp8', 'sp8', '', '', state, {
      stateStorageStrategy: 'cookie',
      encodeBase64: false,
      customFetch,
      eventStore,
    }) ?? fail('Failed to create tracker');

  t.enableActivityTrackingCallback({
    minimumVisitLength: 0,
    heartbeatDelay: 10,
    extendedActivityTracking: true,
    callback: () => {},  // no-op — we're testing the payload, not the callback data
  });
  t.trackPageView();

  // advance 1s, fire two clicks, then complete the heartbeat window
  jest.advanceTimersByTime(1000);
  document.dispatchEvent(new MouseEvent('click'));
  document.dispatchEvent(new MouseEvent('click'));
  jest.advanceTimersByTime(9000);

  const events = await eventStore.getAllPayloads();
  const pings = getPPEvents(events);
  expect(F.size(pings)).toBe(1);

  const entities = JSON.parse(pings[0].co as string).data as Array<{ schema: string; data: any }>;
  const metricsEntity = entities.find((e) => e.schema === ACTIVITY_METRICS_SCHEMA);
  expect(metricsEntity).toBeDefined();
  expect(metricsEntity!.data.clicks).toBe(2);
});
```

matus-tomlein (approved):
LGTM!

PR #1470, opened 2026-04-20
Release 4.8.0

matus-tomlein:
This release adds extended activity tracking with new activity metrics tracked using the `activity_metrics` entity.

It also enahences the optimizely-x plugin to capture the `campaignId` for personalization campaigns.

**Enhancements**
- Add opt-in extended activity tracking with activity_metrics entity (#1467)
- fix(optimizely-x): capture campaignId for Personalization campaigns (#1468) thanks to @JamesShebester


libraries/browser-tracker-core/src/tracker/index.ts
```diff
@@ -514,17 +531,66 @@ export function Tracker(
      * Process all "activity" events.
      * For performance, this function must have low overhead.
      */
-    function activityHandler() {
+    function activityHandler(event?: Event) {
       const now = new Date();
       lastActivityTime = now.getTime();
+
+      if (isActivityMetricsEnabled() && event) {
```
igneel64, line 538: `isActivityMetricsEnabled()` is called during every type of event triggering that is valid for the handler. It could be replaced by a variable and only updated on `enableActivityTracking*` APIs.

igneel64 (approved):

PR #1474, opened 2026-05-19
Strip empty strings from optional media player entity fields

Jack-Keene:
The media_player entity (iglu:com.snowplowanalytics.snowplow/media_player/jsonschema/2-0-0) can produce inconsistent values for optional string fields (label, playerType, quality). The `removeEmptyProperties` function in `browser-plugin-media/src/core.ts` strips null and undefined but passes through empty strings "". This means callers supplying "" where undefined is intended produce events where the field is present but empty, rather than absent.

Downstream, the dbt media player package treats "" and NULL as distinct values. When the same media content generates events with both representations, it produces different surrogate keys for media_identifier, causing merge conflicts and duplicate rows in aggregated models.

### Fix
Added a stripEmptyStrings option to removeEmptyProperties, enabled only for buildMediaPlayerEntity. This keeps the change scoped to the media player entity — other entity builders (ad, ad break, session, event) are unaffected.
When stripEmptyStrings is true, empty string values are filtered out alongside null/undefined, so optional string fields that are "" are omitted from the payload entirely (equivalent to NULL in the warehouse).

### DIFF
--- a/plugins/browser-plugin-media/src/core.ts
+++ b/plugins/browser-plugin-media/src/core.ts
@@ -18,7 +18,7 @@ export function buildMediaPlayerEvent(event: MediaEvent): SelfDescribingJson {
 export function buildMediaPlayerEntity(mediaPlayer: MediaPlayer): SelfDescribingJson {
   return {
     schema: MEDIA_PLAYER_SCHEMA,
-    data: removeEmptyProperties(mediaPlayer, { stripEmptyStrings: true }),
+    data: removeEmptyProperties(mediaPlayer),
   };
 }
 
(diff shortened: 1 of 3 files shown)


plugins/browser-plugin-media/src/core.ts
```diff
@@ -47,12 +47,16 @@ export function buildMediaAdBreakEntity(adBreak: MediaAdBreak): SelfDescribingJs
  * Returns a copy of a JSON with undefined and null properties removed
  *
  * @param event - Object to clean
+ * @param options - Optional settings; if stripEmptyStrings is true, empty string values are also removed
  * @returns A cleaned copy of eventJson
  */
-function removeEmptyProperties(event: Record<string, unknown>): Record<string, unknown> {
+function removeEmptyProperties(
+  event: Record<string, unknown>,
+  options?: { stripEmptyStrings?: boolean }
```
matus-tomlein: Why do we need to make this configurable if it's always set to true in the tracker internals?

matus-tomlein (approved):
LGTM but please update the PR title and description to give more context on exactly what changed and why.

plugins/browser-plugin-media/src/core.ts
Jack-Keene: This probably isn't needed, my thinking was some required properties on the schema that may have a legitimate blank string value, but I think most are set internally.

matus-tomlein (approved):

PR #1480, opened 2026-06-18
update uuid, drop closure-compiler

### LINKED ISSUE #1471: Upgrade uuid to v14, and v10 is deprecated
uuid v10 is deprecated and should be upgraded to v14. thanks
https://npmx.dev/package/uuid/v/10.0.0

vlinder:
I gave the task of updating `uuid` to a non-deprecated version to AI. Apparently closure compiler has become too old and since it is no longer maintained it will never support newer syntax. Opted to remove it from the pipeline.

Fixes #1471.

AI output below.

## Changes

- **uuid → `^14.0.0`** in the 7 packages that use it (`tracker-core`, `browser-tracker-core`, `react-native-tracker`, and the `media`, `media-tracking`, `youtube-tracking`, `screen-tracking` plugins). The code uses only the `v4` named import, which is unchanged.
- **Removed the redundant `@types/uuid`** devDependency from those packages (uuid v14 bundles its own types).
- **Dropped `@ampproject/rollup-plugin-closure-compiler`** from all **33** rollup configs and their `package.json` devDependencies. 31 already chained `compiler(), terser()` → now just `terser()`; `javascript-tracker` likewise; `browser-tracker` was the one config relying on closure as its *only* minifier, so I swapped `compiler()` → `terser()` (it already had `rollup-plugin-terser` available).
- Regenerated the lockfile (`uuid@14.0.0` resolves) and verified a full `rush build` succeeds for every package (only the pre-existing browserslist warning remains).

## Size trade-off

**Missed savings (the cost of dropping closure-compiler)**:

| Bundle | gzip before | gzip now | cost |
|---|---|---|---|
| `tracker-core` min | 9,330 B | 9,507 B | +177 B (+1.9%) |
| `browser-tracker-core` min | 16,880 B | 17,285 B | +405 B (+2.4%) |
| `sp.js` (full CDN) | 25,264 B | 26,064 B | +800 B (+3.2%) |
| `sp.lite.js` (lite CDN) | 17,786 B | 18,267 B | +481 B (+2.7%) |

So we give up roughly **2–3% gzipped** per bundle — at most ~0.8 KB on the largest CDN artifact. Terser alone retains ~97–98% of the minification benefit.

**What we win:**
- **Unblocks uuid v11+/v14** (and any future modern-syntax dependency) — closure's bundled `acorn-walk@7.1.1` couldn't parse `??`/`?.()` and was the hard blocker.
- **Removes an unmaintained build dependency** [Deprecated since 2024](https://en.wikipedia.org/wiki/Google_Closure_Tools#:~:text=As%20of%20August%201%2C%202024,needs%20of%20modern%20JavaScript%20development%22.)
- **Faster builds** — in my runs the no-closure rebuilds were consistently quicker (e.g. `browser-tracker-core` ~9.7s vs ~13.6s).
- **Simpler, single-minifier pipeline** (terser) that's actively maintained and understands modern JS.

Note: uuid v14 requires Node 20+ (repo's `rush.json` is already `>=20.19.0`, so fine). I left the harmless `closure-compiler` entry in `common/config/rush/browser-approved-packages.json` and didn't touch `react-native-tracker`'s build (it doesn't use closure).

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -19,12 +19,9 @@ importers:
         specifier: ^2.3.1
         version: 2.8.1
       uuid:
-        specifier: ^10.0.0
-        version: 10.0.0
+        specifier: ^14.0.0
+        version: 14.0.0
     devDependencies:
-      '@ampproject/rollup-plugin-closure-compiler':
-        specifier: ~0.27.0
-        version: 0.27.0(rollup@2.70.2)
       '@rollup/plugin-commonjs':
         specifier: ~21.0.2
         version: 21.0.3(rollup@2.70.2)
@@ -40,9 +37,6 @@ importers:
       '@types/jsdom':
         specifier: ~16.2.14
         version: 16.2.15
-      '@types/uuid':
-        specifier: ^10.0.0
-        version: 10.0.0
       '@typescript-eslint/eslint-plugin':
         specifier: ~5.15.0
         version: 5.15.0(@typescript-eslint/parser@5.15.0(eslint@8.11.0)(typescript@4.6.4))(eslint@8.11.0)(typescript@4.6.4)
@@ -95,12 +89,9 @@ importers:
         specifier: ^2.3.1
         version: 2.8.1
       uuid:
-        specifier: ^10.0.0
-        version: 10.0.0
+        specifier: ^14.0.0
+        version: 14.0.0
     devDependencies:
-      '@ampproject/rollup-plugin-closure-compiler':
-        specifier: ~0.27.0
-        version: 0.27.0(rollup@2.70.2)
       '@rollup/plugin-commonjs':
         specifier: ~21.0.2
         version: 21.0.3(rollup@2.70.2)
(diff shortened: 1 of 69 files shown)


snowplowcla:

vlinder:
> Thanks for your pull request. Is this your first contribution to a Snowplow open source project? Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

Signed

snowplowcla:
Confirmed! @vlinder has signed the Contributor License Agreement. Thanks so much.

joaogarin:
Can this be merged? Right no we have to add resolutions for the uuid vulnerability would be nice to not have to 🙏 Thanks!

matus-tomlein:
Thank you for the contribution, @vlinder!

The only concern here is that this would be a breaking change due to the uuid v14 not supporting CommonJS modules and requiring Node 20+. The current version of our NodeJS tracker (part of this monorepo) supports Node v18 (understand it has reached EOL), so we would need to publish a new major version of the tracker. We try to do these major releases sparingly, so would prefer to avoid it.

Although uuid@10 has been deprecated, uuid@11 is still being maintained this year, so we would prefer to update to that version for now so that we keep current Node version compatibility.

vlinder:
Sounds good!
Didn't see the issue first since closure compiler had issues with both of them. But that seems like the pragmatic approach for now.

PR #1481, opened 2026-06-24
fix: update uuid to v11, drop closure-compiler (keep Node 18 support)

### LINKED ISSUE #1471: Upgrade uuid to v14, and v10 is deprecated

matus-tomlein:
## Summary

Alternative to #1480 that keeps **Node 18 / CommonJS consumer support** by pinning `uuid` to **v11** instead of v14.

`uuid` v12+ is **ESM-only** (CommonJS support was dropped in v12). Since `@snowplow/tracker-core` / `@snowplow/node-tracker` ship CommonJS builds that `require('uuid')`, jumping to v14 would break downstream CJS consumers on Node < 20.19. **uuid v11 still ships both CJS and ESM builds**, so it's the last line that preserves that support — and it is not yet deprecated (the deprecation notice itself recommends v11 for CommonJS codebases).

The code uses only the `v4` named import, which is unchanged across these versions.

Fixes #1471.

## Changes

- **`uuid` `^10.0.0` → `^11.0.0`** in the 7 packages that use it (`tracker-core`, `browser-tracker-core`, `react-native-tracker`, and the `media`, `media-tracking`, `youtube-tracking`, `screen-tracking` plugins).
- **Removed the redundant `@types/uuid`** devDependency (uuid v11 bundles its own types).
- **Dropped `@ampproject/rollup-plugin-closure-compiler`** from all 33 rollup configs and their `package.json` devDependencies. This is **required**, not optional: uuid v11 emits modern syntax (`??` / `?.`) that closure-compiler's bundled `acorn-walk@7.1.1` cannot parse — the build fails at the closure step otherwise. 31 configs already chained `compiler(), terser()` → now just `terser()`; `browser-tracker` (closure was its only minifier) was swapped to `terser()` (already an available devDependency).
- **Updated the `uuid.v4` jest spies** in 3 `browser-tracker-core` test files to satisfy v11's overloaded type signature (`v4` now also has a `Uint8Array`-returning overload).
- Regenerated the lockfile and added Rush change files.

## Validation

- ✅ `rush build` succeeds for every package (only the pre-existing browserslist warning remains).
- ✅ Tests pass for all affected packages: `tracker-core` (130), `browser-tracker-core` (162), `react-native-tracker` (68), `media` (67), `media-tracking` (38), `screen-tracking` (9), `youtube-tracking` (15).
- ✅ `rush change --verify` passes.

## Size trade-off

Dropping closure-compiler costs ~2–3% gzipped per bundle (≤0.8 KB on `sp.js`); terser retains ~97–98% of the minification benefit. Same trade-off as #1480.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/common/changes/@snowplow/browser-plugin-ad-tracking/fix-update-uuid-v11_2026-06-24-08-39.json
+++ b/common/changes/@snowplow/browser-plugin-ad-tracking/fix-update-uuid-v11_2026-06-24-08-39.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "Update uuid to v11 and remove closure-compiler from the build pipeline",
+      "type": "none",
+      "packageName": "@snowplow/browser-plugin-ad-tracking"
+    }
+  ],
+  "packageName": "@snowplow/browser-plugin-ad-tracking",
+  "email": "matus.tomlein@gmail.com"
+}
--- a/common/changes/@snowplow/browser-plugin-ad-tracking/fix-update-uuid-v11_2026-06-24-08-39.json
+++ b/common/changes/@snowplow/browser-plugin-ad-tracking/fix-update-uuid-v11_2026-06-24-08-39.json
@@ -6,6 +6,5 @@
       "packageName": "@snowplow/browser-plugin-ad-tracking"
     }
   ],
-  "packageName": "@snowplow/browser-plugin-ad-tracking",
-  "email": "matus.tomlein@gmail.com"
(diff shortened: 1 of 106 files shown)


vlinder:
It could make sense to "force" a later v11, `^11.1.1`, that has the vulnerabilities patched.

vlinder (approved):

igneel64 (approved):

PR #1483, opened 2026-07-02
fix(react-native-tracker): allow react-native-get-random-values v2.0.0

matus-tomlein:
## Summary

Widens the `react-native-get-random-values` peer dependency range in `@snowplow/react-native-tracker` from `^1.11.0` to `^1.11.0 || ^2.0.0`, so consumers on React Native 0.81+ can install `v2.0.0` (which requires RN >= 0.81), while existing users on older RN versions stay on `v1.x`.

Same pattern as the earlier `@react-native-async-storage/async-storage` bump.

## Why widen instead of bumping to `^2.0.0`

- **`v2.0.0` has no JS API changes.** Per its release notes, the only breaking change is the raised React Native version floor (New Architecture, RN >= 0.81).
- The tracker's only usage is the side-effect polyfill import (`import 'react-native-get-random-values'` in `src/index.ts`), which is unaffected by the major bump.
- Widening (rather than replacing) lets users on RN 0.81+ install `v2`, while **RN < 0.81 users keep working on `v1.x`**. A hard `^2.0.0` would force RN >= 0.81 on all consumers — an unnecessary regression.
- The `devDependencies` entry is widened the same way; pnpm correctly resolves it back to `1.11.0` for the existing dev toolchain (RN 0.74.5), so no dev-side conflict and no risky dev RN upgrade.

## Testing

- `rush update` regenerated the lockfile cleanly (specifier `^1.11.0 || ^2.0.0`, resolved to `1.11.0` for dev), no new peer warnings.
- `rush build --to @snowplow/react-native-tracker` succeeds.
- All 68 tests across 13 suites pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/common/changes/@snowplow/react-native-tracker/claude-loving-feynman-b5ef83_2026-07-02-08-48.json
+++ b/common/changes/@snowplow/react-native-tracker/claude-loving-feynman-b5ef83_2026-07-02-08-48.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "comment": "Widen react-native-get-random-values peer dependency range to allow v2.0.0 for React Native 0.81+ compatibility",
+      "type": "none",
+      "packageName": "@snowplow/react-native-tracker"
+    }
+  ],
+  "packageName": "@snowplow/react-native-tracker"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -2592,7 +2592,7 @@ importers:
         specifier: ^0.42.1
         version: 0.42.1
       react-native-get-random-values:
-        specifier: ^1.11.0
+        specifier: ^1.11.0 || ^2.0.0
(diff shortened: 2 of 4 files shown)


common/changes/@snowplow/react-native-tracker/claude-loving-feynman-b5ef83_2026-07-02-08-48.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "comment": "Widen react-native-get-random-values peer dependency range to allow v2.0.0 for React Native 0.81+ compatibility (CSTMR-2117)",
```
igneel64: Not sure we want to add  (CSTMR-2117) on the comment of changes. Plus it would be nice if we adjusted the branch name before merging. We can do it for the next one but let's keep this as a practise.

igneel64 (approved):

PR #1485, opened 2026-07-12
Support non-alphabetic characters in URL scheme detection (close #1238)

### LINKED ISSUE #1238: Reconsider regex for URL scheme
**Describe the bug**

Currently [this regex is used](https://github.com/snowplow/snowplow-javascript-tracker/blob/de97c3bef9ef4ceccd001ea051eddbf2a05e2c70/libraries/browser-tracker-core/src/tracker/index.ts#L426) to validate the protocol scheme. This means that a protocol like `safari-web-extention` will become `null`.

This can result in unexpected behaviour for `setCustomUrl`, if it is called with a url whose protocol does not conform to this regex. In that case, the url argument [gets appended](https://github.com/snowplow/snowplow-javascript-tracker/blob/de97c3bef9ef4ceccd001ea051eddbf2a05e2c70/libraries/browser-tracker-core/src/tracker/index.ts#L457) to base url.

**To Reproduce**

Example for JavaScript tracker, setup the tracker and call:

```
snowplow('setCustomUrl', 'safari-web-extention://whatever.com');
```

**Expected behavior**

That `setCustomUrl` sets the url to its argument independent of protocol scheme.

**Additional context**

- Currently the `page_urlscheme` is also a Snowplow atomic property, whose current max length is 16, which results in events getting invalidated if their page url has a protocol longer than that.
- See also: [RFC 2396 Scheme Component](https://datatracker.ietf.org/doc/html/rfc2396#section-3.1)

sarmah-rup:
## Bug

`setCustomUrl` mangles any URL whose scheme contains a non-alphabetic character. For example:

```js
tracker.setCustomUrl('safari-web-extension://abcdefg/index.html');
```

is recorded as `https://<base>/.../safari-web-extension://abcdefg/index.html` instead of the URL that was passed in.

## Root cause

`getProtocolScheme` detects the scheme with `^([a-z]+):`. RFC 3986 allows a scheme to contain digits, `+`, `-` and `.` after the first letter (and to be case-insensitive), so `safari-web-extension`, `chrome-extension`, `ms-appx`, etc. don't match. `resolveRelativeReference` then treats the value as a relative reference and appends it to the base URL.

## Solution

Broaden the pattern to the RFC 3986 scheme grammar:

```js
new RegExp('^([a-z][a-z0-9+\\-.]*):', 'i')
```

Closes #1238

## Testing

Added a unit test in `browser-tracker-core`'s `page_view.test.ts` (using the existing `afterTrack` plugin to capture the tracked `url`): `setCustomUrl('safari-web-extension://abcdefg/index.html')` now records that exact URL. Before the change the test fails (URL appended to base). Full `browser-tracker-core` suite passes (163/163).

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/fix-url-scheme-hyphen_2026-07-20-22-44.json
+++ b/common/changes/@snowplow/browser-tracker-core/fix-url-scheme-hyphen_2026-07-20-22-44.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@snowplow/browser-tracker-core",
+      "comment": "Support non-alphabetic characters in URL scheme detection and keep the detected scheme within the atomic schema length limit (close #1238)",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@snowplow/browser-tracker-core"
+}
--- a/libraries/browser-tracker-core/src/tracker/index.ts
+++ b/libraries/browser-tracker-core/src/tracker/index.ts
@@ -456,7 +456,10 @@ export function Tracker(
      * Extract scheme/protocol from URL
      */
     function getProtocolScheme(url: string) {
-      const e = new RegExp('^([a-z]+):'),
+      // RFC 3986: scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." ), case-insensitive.
+      // A stricter [a-z]+ pattern misclassifies schemes such as safari-web-extension as
+      // relative references, which then get appended to the base URL.
+      const e = new RegExp('^([a-z][a-z0-9+\\-.]*):', 'i'),
         matches = e.exec(url);
 
       return matches ? matches[1] : null;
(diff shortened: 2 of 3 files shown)


snowplowcla:

sarmah-rup:
I signed it!

snowplowcla:
Confirmed! @sarmah-rup has signed the Contributor License Agreement. Thanks so much.

matus-tomlein:
Thank you for the contribution!

The change makes sense, however even if we allow the `safari-web-extension://` protocol, it will then cause event validation failures, as the schema restricts the length of the protocol to 16 characters: https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/atomic/jsonschema/1-0-0#L136-L138

I can see the length restriction is unrelated to the non-alphabetic characters support, but kind of one is keeping the other one in check. It would be better to add a dedicated validation for length to be less than 16 chars.

Raising the limit in the event schema would be ideal, but that is a breaking change that affects every deployment of Snowplow, so not something that can be done lightly.

sarmah-rup:
Makes sense, thanks for catching that. I pushed a length guard so any scheme over 16 chars falls back to a relative reference, which keeps events inside the atomic schema limit. The test now uses `chrome-extension://` (a valid 16-char scheme) plus a case that keeps `safari-web-extension://` relative.

matus-tomlein (approved):
LGTM, thanks again!

matus-tomlein:
Sorry, just a final small request, could you please run `rush change` to generate the changefiles?

PR #1487, opened 2026-07-16
Add fallback for element_index

Jack-Keene:
- When the intersection observer callback fires, the plugin re-queries the DOM to determine the element's position among siblings. If the element has already been removed, findIndex returns -1, producing element_index: 0 which fails validation against the element schema (minimum: 1).
- Fall back to the element's last-known position stored in state.lastPosition when the DOM lookup
  fails, and keep that value fresh by updating it on every successful lookup.


igneel64:
@Jack-Keene Changing target to release/4.9.0

igneel64 (approved):

PR #1488, opened 2026-07-16
[AISP-1002] fix(browser-tracker-core): use ResizeObserver entry sizes to avoid fo…

jpiekos:
## Summary

Google PageSpeed flagged the JS tracker as causing forced synchronous layouts originating from `detectDocumentSize()` in `browser-tracker-core`. Accessing `offsetWidth`, `offsetHeight`, `scrollWidth`, `scrollHeight`, `clientWidth`, and `clientHeight` after a DOM change forces the browser to immediately recompute layout.

The `ResizeObserver` callback already deferred cache updates via `requestAnimationFrame`, but the update path still called `readBrowserProperties()` which re-queried all the reflow-triggering layout properties on each resize event.

Fixes AISP-1002.

## Changes

- Store element dimensions captured from `ResizeObserverEntry` (`borderBoxSize` with `contentRect` fallback) in module-level variables updated by the ResizeObserver callback.
- Update `detectDocumentSize()` to use those stored entry values when available, falling back to direct DOM reads only on the initial call (before the ResizeObserver has fired) or in browsers without ResizeObserver support.
- Add `getSizeFromEntry()` helper to extract border-box dimensions from an entry with a `contentRect` fallback for older ResizeObserver implementations.

## Validation

- All 6 tests in `browser_props.test.ts` pass, including two new tests covering the `borderBoxSize` path and the `contentRect` fallback.

### DIFF
--- a/common/changes/@snowplow/browser-tracker-core/fix-browser-props-resize-observer-reflow_2026-07-16-20-23.json
+++ b/common/changes/@snowplow/browser-tracker-core/fix-browser-props-resize-observer-reflow_2026-07-16-20-23.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "Avoid forced reflows when reading document size from ResizeObserver callbacks",
+      "type": "none",
+      "packageName": "@snowplow/browser-tracker-core"
+    }
+  ],
+  "packageName": "@snowplow/browser-tracker-core",
+  "email": "198982749+Copilot@users.noreply.github.com"
+}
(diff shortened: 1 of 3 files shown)


matus-tomlein:
@copilot are we confident that these changes will fix the number of layout shifts? It seems the change has only been tested using unit tests, does that give enough confidence? should an integration test be added and can that be done in a straightforward way?

matus-tomlein:
Closing in favour of #1490

